PR #1, opened 2019-09-13
Example for extracting data from multiples pages

VMRuiz:


example/example/spiders/books_06.py
```diff
@@ -0,0 +1,51 @@
...
+        product_urls = page.product_urls()
+        page_collection = PageCollection(len(product_urls))
+        for url in product_urls:
+            yield response.follow(
+                url, self.parse_book, cb_kwargs={"page_collection": page_collection}
+            )
+
+    def parse_book(self, response, page: BookPage, page_collection: PageCollection):
+        page_collection.add_page(page)
+        if page_collection.is_complete():
```
BurnzZ: I like this approach! I think we should also have a contingency for something like the last few URLs hasn't reached this callback because of some error, max retry has been reached, etc since the item wouldn't be yielded.

PR #2, opened 2020-02-06
Recursive injection with an example

ivanprado:
Having recursive dependency injection can help to create smaller PageObject components that extracts only parts of a page, and then define compounded PageObjects with several of them. See example `books_06`. 

The example of `books_06` presents boilerplate. A lot of `yield from self.` and `response` repeated and repeated. 

@kmike It would be nice if you can have a look and give your opinion about all that. 

I was not able to check if cyclic dependency detection is working
because I failed to create an example. The following fails with the error `NameError: name 'B' is not defined` in method `get_type_hints`:

```py
@attr.s(auto_attribs=True)
class A(WebPage):
    b: 'B'

@attr.s(auto_attribs=True)
class B(WebPage):
    a: A
```

### DIFF
--- a/example/example/spiders/books_06.py
+++ b/example/example/spiders/books_06.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+"""
+Scrapy spider with reusable components used for two types of pages:
+* Listings page, with the components: book list and pagination
+* Detail book page, with the components: the item itself, the breadcrumbs and
+  the recently viewed list of books
+
+All links to detail book pages are found are followed.
+All links to search result pages are followed as well.
+"""
+import scrapy
+import attr
+
+from scrapy_po import WebPage
+
+
+class ListingsPiece(WebPage):
+    def urls(self):
+        return self.css('.image_container a::attr(href)').getall()
+
+
+class PaginationPiece(WebPage):
+    def urls(self):
+        return self.css('.pager a::attr(href)').getall()
+
+
+class BreadcrumbsPiece(WebPage):
+    def urls(self):
+        return self.css('.breadcrumb a::attr(href)').getall()
+
+
+@attr.s(auto_attribs=True)
+class ListingsPage(WebPage):
+    book_list: ListingsPiece
+    pagination: PaginationPiece
+
+
+@attr.s(auto_attribs=True)
+class BookPage(WebPage):
+    recently_viewed: ListingsPiece
+    breadcrumbs: BreadcrumbsPiece
+
+    def to_item(self):
+        return {
+            'url': self.url,
+            'name': self.css("title::text").get(),
+        }
+
+
+class BooksSpider(scrapy.Spider):
+    name = 'books_06'
+    start_urls = ['http://books.toscrape.com/']
+
+    def parse(self, response, page: ListingsPage):
+        """ Callback for Listings pages """
+        yield from self.follow_listing(response, page.book_list)
+        yield from self.follow_pagination(response, page.pagination)
+
+    def parse_book(self, response, page: BookPage):
+        yield from self.follow_listing(response, page.recently_viewed)
+        yield from self.follow_breadcrumbs(response, page.breadcrumbs)
+        yield page.to_item()
+
+    def follow_listing(self, response, item_list: ListingsPiece):
+        for url in item_list.urls():
+            yield response.follow(url, self.parse_book)
+
+    def follow_pagination(self, response, pagination: PaginationPiece):
+        for url in pagination.urls():
+            yield response.follow(url, self.parse, priority=+10)
+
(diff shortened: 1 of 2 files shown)


kmike:
I wonder if a problem with get_type_hints can be solved using https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations, though I kind-of expected it to work as-is.

example/example/spiders/books_06.py
```diff
@@ -0,0 +1,74 @@
...
+All links to detail book pages are found are followed.
+All links to search result pages are followed as well.
+"""
+import scrapy
+import attr
+
+from scrapy_po import WebPage
+
+
+class ListingsPiece(WebPage):
```
kmike: Any ideas about a better API? Inheriting "page part" component from a WebPage looks a bit strange, on a first sight

scrapy_po/middleware.py
```diff
@@ -33,32 +37,38 @@ def process_response(self, request: Request, response, spider):
...
         raise returnValue(response)
 
-    @classmethod
-    def is_page_object(cls, argument_type) -> bool:
-        return issubclass(argument_type, PageObject)
-
 
 @inlineCallbacks
-def create_page_object(page_cls, response):
+def create_page_object(page_cls, response, dependency_graph=[]):
```
kmike: mutable default  detected :)

example/example/spiders/books_06.py
ivanprado: Yes is strange. Also, all the "pieces" are incorporating the selector API without sharing it, which I don't like very much. Well, I have to think about it. This is very preliminary to me, I just noticed that I needed the recursive injection and started to experiment with it.

scrapy_po/middleware.py
ivanprado: TIL  Still a newbie :smile:

example/example/spiders/books_06.py
kmike: > Also, all the "pieces" are incorporating the selector API without sharing it, which I don't like very much.

Sorry, could you please explain it more?

kmike: If the goal of inheriting is to be able to say "please do dependency injection for this class", it could make sense to have a base class for this (e.g. Component, or something like this - this is becoming Java pretty fast), and have PageObject inherit from Component.

kmike: Aha, maybe I know what you mean. In the code I was showing you on a call (which is not commited), selector-specific logic is moved to a mixin:

```py
class PageObject(abc.ABC):
    """
    All web page objects should inherit from this class.
    It is a marker for a middleware to pick anargument and populate it.

    Instead of inheriting you can also use ``PageObject.register(MyWebPage)``.
    """
    pass


class ItemPage(PageObject):
    """ Page Object which requires to_item method to be implemented. """
    @abc.abstractmethod
    def to_item(self):
        pass


class ResponseShortcutsMixin:
    """
    A mixin with common shortcut methods for working with self.response.
    It requires "response" attribute to be present.
    """
    _cached_selector = None

    @property
    def selector(self) -> parsel.Selector:
        from parsel.selector import Selector
        if self._cached_selector is None:
            self._cached_selector = Selector(self.html)
        return self._cached_selector

    @property
    def url(self):
        return self.response.url

    @property
    def html(self):
        return self.response.html

    def xpath(self, query, **kwargs):
        return self.selector.xpath(query, **kwargs)

    def css(self, query):
        return self.selector.css(query)


@attr.s(auto_attribs=True)
class WebPage(PageObject, ResponseShortcutsMixin):
    """
    Default WebPage class. It uses basic response data (html, url),
    and provides xpath / css shortcuts.

    This class is most useful as a base class for page objects, when
    extraction logic is similar to what's usually happening in scrapy
    spider callbacks.
    """
    response: ResponseData
```

Let me asend a PR with these changes.

kmike: https://github.com/scrapinghub/scrapy-po/pull/3

ivanprado: Thanks @kmike ! I left some comments in the PR itself that I think covers yoursquestions.

ivanprado:
Thank @kmike for all your feedback. A quick sight to the  `get_type_hints` problems reveals that the globals that `get_type_hints` use at the constructor method are limited to the class itself so module globals are not included (and they should). Probably andi should be changed to also include the module in globals. Could you give me rights in andi to propose PR requests?

ivanprado:
It seems that the problem with the type is related to an issue in attrs: https://github.com/python-attrs/attrs/issues/593. I'll create a `xfail` test in andi for it.

ivanprado:
This has been implemented in https://github.com/scrapinghub/scrapy-po/pull/2 in a different way, probably better. And the problem with `get_type_hints` was addressed in andi. So closing.

PR #3, opened 2020-02-06
Extract selector-specific logic to a mixin.

kmike:
This is backwards incompatible: ItemPage doesn't provide selector
shortcuts anymore, and so it doesn't require a response.

### DIFF
--- a/scrapy_po/webpage.py
+++ b/scrapy_po/webpage.py
@@ -17,17 +17,18 @@ class PageObject(abc.ABC):
     pass
 
 
-@attr.s(auto_attribs=True)
-class WebPage(PageObject):
-    """
-    Default WebPage class. It uses basic response data (html, url),
-    and provides xpath / css shortcuts.
+class ItemPage(PageObject):
+    """ Page Object which requires to_item method to be implemented. """
+    @abc.abstractmethod
+    def to_item(self):
+        pass
 
-    This class is most useful as a base class for page objects, when
-    extraction logic is similar to what's usually happening in scrapy
-    spider callbacks.
+
+class ResponseShortcutsMixin:
     """
-    response: ResponseData
+    A mixin with common shortcut methods for working with self.response.
+    It requires "response" attribute to be present.
+    """
     _cached_selector = None
 
     @property
@@ -52,10 +53,17 @@ class WebPage(PageObject):
         return self.selector.css(query)
 
 
-class ItemPage(WebPage):
-    @abc.abstractmethod
-    def to_item(self):
-        pass
+@attr.s(auto_attribs=True)
+class WebPage(PageObject, ResponseShortcutsMixin):
+    """
+    Default WebPage class. It uses basic response data (html, url),
+    and provides xpath / css shortcuts.
+
+    This class is most useful as a base class for page objects, when
+    extraction logic is similar to what's usually happening in scrapy
+    spider callbacks.
+    """
+    response: ResponseData
 
 
 def callback_for(page_cls):


ivanprado:
Thanks, @kmike for this proposals. Impressions I have. 
* Should we considering the introduction of the `ItemMixin`?. This will allow having `WebPage` based PO that also returns an item. For example:

```py
class BookPage(WebPage, ItemMixin):
   ....
```

ItemPage would be redefined as:
```py
class ItemPage(PageObject, ItemMixin):
    pass
```

By the other side, I instead think `ItemPage` should be removed completely. It is only required as a "convention" for the method `callback_for`. It seems not possible to implement it in a way that `mypy` type checking can detect wrong code statically. So I think it is just boilerplate then and is making things more complex when you need both the `WebPage` behaviour and the `ItemPage` one. In other words, I would keep the `to_item` as a convention without any explicit declaration in the form of a class. Same that `parse` in Scrapy. What do you think?

* The introduction of `ResponseShortcutsMixin` allows having "scrapy flavour" in compounded PO where `ResponseData` is required in addition to other required POs, which is nice. By the other hand, given that currently `WebPage` definition is so simple (just a single `ResponseData`) the same behaviour can be achieved by just extending `WebPage`, but maybe in the future WebPage includes something else, so I think that `ResponseShortcutsMixin` is something good to have. 

* By the other side, it does not solve the problem of redundant `Selector` in compound objects. For example:

```py
class Breadcrumbs(WebPage):
    urls = lambda: self.css(...).getall()

@attr.s(auto_attribs=True)
class Book(WebPage):
    breadcrumbs: Breacrumbs
    title = lambda: self.css(...).get()
```

In this case, there are two equal instances of `ResponseData` (one in `Breadcrumbs` and one in `Book`) and the html is parsed twice, which seems not convenient. 

A possible solution would be to:

1. Ensure that instances injected by the middleware are singletons within the scope of this particular injection. In other words, a single instance of `ResponseData` is created and it is injected where required.
2. move the cached selector to `ResponseData` and modify the `ResponseShortcutsMixin` to access it. 

What do you think of all of that? I can do a PR with all these changes to see how it would look like. It would be nice I can do the PR directly over this repository instead of my own fork, could I have rights for that?

kmike:
> I would keep the to_item as a convention without any explicit declaration in the form of a class. Same that parse in Scrapy. What do you think?

ItemPage is an abstract method which requires a subclass to implement to_item method; if a class doesn't do it, instantiation would fail (i.e. it would fail earlier, before the object is created). That was my intention with ItemPage - provide a validated interface for developers to implement. One example, as you noticed, is callback_for method. That's not the only use case though, another example: if we put extraction code in the backend, it may be required that a page object has to_item method.

In case of Scrapy `parse` is a regular method, you can have spiders which don't use it (just use another method as a callback); it is totally up to users. In scrapy-po one can also have objects which don't have to_item method (e.g. they're components and are not intended to return items); in this case there is no need to mark them as ItemPage.

Maybe ItemPage can be called ItemPageMixin; it can be used as a mixin.  I named it ItemPage to have a marker - "this page is an ItemPage", can be used in a particular context. 

> By the other hand, given that currently WebPage definition is so simple (just a single ResponseData) the same behaviour can be achieved by just extending WebPage

This refactoring was triggered by AutoExtractProductPage with fullHtml=True. This page shouldn't require ResponseData, as it requires AutoextractProductResponse. But it can still allow using Selectors, by getting html and url from the autoextract response output. Actually, maybe it was not AutoExtractProductPage :) But response which is e.g. SplashResponse. Anyways, the idea is that you don't need to require ResonseData to use selector.

kmike:
> Ensure that instances injected by the middleware are singletons within the scope of this particular injection. In other words, a single instance of ResponseData is created and it is injected where required.

I think that'd be good to ensure for recursive injection in general (unless there are problems with that).

Regarding selector cache, data point: Scrapy used to have a weakref cache with parser lxml trees (it got removed after some time).

kmike:
> It would be nice I can do the PR directly over this repository instead of my own fork, could I have rights for that?

You should have it since yesterday :)

ivanprado:
Fair enough :+1: Thanks @kmike . I'm going to merge that and work on the recursive injection on top of it.

PR #4, opened 2020-02-07
[WIP] Recursive injection

ivanprado:


scrapy_po/middleware.py
```diff
@@ -33,32 +37,39 @@ def process_response(self, request: Request, response, spider):
...
     arguments = andi.inspect(page_cls.__init__)
-    kwargs_to_provide = andi.to_provide(arguments, can_provide=provider_funcs)
+    kwargs_to_provide = andi.to_provide(arguments, can_provide=can_provide)
     not_supported = arguments.keys() - kwargs_to_provide.keys()
     if not_supported:
         raise TypeError("Can't instantiate arguments: %s" % not_supported)
 
     # instantiate all arguments
     kwargs = {}
+    dependency_graph = dependency_graph or []
```
kmike: I wonder if obtaining of a full dependency graph can be moved to andi, or separated in some way. We may need this outside Scrapy as well.

ivanprado: Good idea. I have implemented a tentative version for andi here: https://github.com/scrapinghub/andi/pull/7 and a new implementation to replace current PR here: https://github.com/scrapinghub/scrapy-po/pull/5 . I will close this PR.

PR #5, opened 2020-02-18
Recursive dependency injection in the Middleware

ivanprado:
Middleware rewrote using the new `andi.plan` function (see https://github.com/scrapinghub/andi/pull/7). 

It now supports:
* recursive dependency injection, with cyclic dependencies detection
* shared instances in all the dependency tree. (each instance is a singleton in the request scope)

Also `None` is now supported by its type instead of by the instance itself. Then the provider for None is not required anymore. 

An example using this recursive dependency injection can be found in books_06.py

`asyc` is now supported, but now Scrapy 2.0.1+  is required. 

**Note that this changes are dependant on https://github.com/scrapinghub/andi/pull/7 so don't merge them until andi gets released properly**


example/example/spiders/books_06.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class BooksSpider(scrapy.Spider):
+    name = 'books_06'
+    start_urls = ['http://books.toscrape.com/']
+
+    def parse(self, response, page: ListingsPage):
+        """ Callback for Listings pages """
+        yield from self.follow_listing(response, page.book_list)
+        yield from self.follow_pagination(response, page.pagination)
```
kmike: Extracting this code in a spider to separate methods makes sense. I wonder if we can have Scrapy-specific wrappers around Paginator&friends, which provide a method to get scrapy.Requests instead of just urls. Code could look like this then:

```py
def parse(self, response, page: ListingsPage):
    yield from page.book_list.requests(self.parse_book)
    yield from page.pagination.requests(self.parse, priority=+10)
```

Wrapping can be either implicit (wrap everythig which callbacks receive) or, better, explicit; a few options:

```py
class ScrapyListingPage(ListingsPage, RequestsFromUrlsMixin):
    pass

ScrapyListingPage = scrapy_po.add_requests_method(ListingsPage)
```

or even just this (requires upcoming scrapy 2.0 - see https://github.com/scrapy/scrapy/pull/4057):
```py
yield from response.follow_all(page.pagination.urls(), self.parse_book)
```

It is better to handle this separately though, not in this PR, unless we switch to scrapy 2.0.

ivanprado: I see. I understand. Makes sense. I'm going to create a PR with the explicit variant you proposed.

> unless we switch to scrapy 2.0.

I have not enough confidence/knowledge to say if we should or not. Should we?

ivanprado: Filled a PR here: https://github.com/scrapinghub/scrapy-po/pull/9

kmike: I think we should target Scrapy 2.0 in examples, but provide compatibility with Scrapy 1.8 if possible (it doesn't have to have all the same usability features - i.e. no need to backport response.follow_all). It is going to be released very soon, maybe even this week.

Scrapy 2.0 is good, because it should allow to use https://github.com/scrapinghub/scrapinghub-autoextract asyncio interface for Scrapy integration, which provides a proper retry feature / handling of rate limits.

For examples it shouldn't matter that they're not compatible with Scrapy < 2.0.

kmike: See https://github.com/scrapy/scrapy/blob/034e2c31c7d55333c3de208f80dcee1bf45ef9b9/scrapy/http/response/text.py#L122

example/example/spiders/books_06.py
```diff
@@ -0,0 +1,75 @@
...
+
+@attr.s(auto_attribs=True)
+class ListingsPage(WebPage):
+    book_list: ListingsPiece
+    pagination: PaginationPiece
+
+
+@attr.s(auto_attribs=True)
+class BookPage(WebPage, ItemPage):
+    recently_viewed: ListingsPiece
```
kmike: This is the part I'm not sure about, as a general pattern. There can be several listings on a single page - e.g. 

* recently viewed books 
* suggested/recommended books
* reviews for this book

It works on this particular website, but as an example it may not scale. Because of that I'd remove ListingPiece from the example, and add a method to BookPage instead for now.

example/example/spiders/books_06.py
kmike: Sorry, I should have thought about it more. I think switching to Scrapy 2.0 in examples is better than adding custom shortcuts to the codebase.

scrapy_po/middleware.py
```diff
@@ -21,45 +23,44 @@ class InjectPageObjectsMiddleware:
...
+        instances = {}
+        for cls, params in plan.items():
+            if cls in providers:
+                instances[cls] = yield maybeDeferred(providers[cls])
+            elif cls == andi.FunctionArguments:
+                pass
+            else:
+                kwargs = {param: instances[pcls]
+                          for param, pcls in params.items()}
+                instances[cls] = cls(**kwargs)
```
kmike: I think it'd be nice to move this code to a separate function, similar to andi.build or create_page_object

kmike: It can be seen as "async andi.build", and it may be useful to have it available for use outside a middleware.

scrapy_po/middleware.py
```diff
@@ -21,45 +23,44 @@ class InjectPageObjectsMiddleware:
     """
     @inlineCallbacks
     def process_response(self, request: Request, response, spider):
-        # find out which arguments need to be created
+        # find out the dependencies
         callback = get_callback(request, spider)
-        kwargs_to_provide = andi.to_provide(callback,
-                                            can_provide=self.is_page_object)
+        providers = build_providers(response)
```
kmike: provider_instances? PyCharm shows me that this name shadows "providers" imported from page_input_providers, which is not good

scrapy_po/middleware.py
```diff
@@ -21,45 +23,44 @@ class InjectPageObjectsMiddleware:
...
+        can_provide = can_provide_fn(providers)
```
kmike: I think it'd be a bit more clear what's going on if you pass providers.keys() instead of just providers

scrapy_po/middleware.py
```diff
@@ -21,45 +23,44 @@ class InjectPageObjectsMiddleware:
...
+        plan = andi.plan(callback, can_provide,
+                         providers.__contains__)
```
kmike: could you please use keyword arguments for arguments other than callback? I wonder if we should make them keyword-only in andi..

kmike:
@ivanprado overall the approach makes a lot of sense (here and in andi); I left a ton of comments, but they're mostly stylistic.

scrapy_po/middleware.py
ivanprado: Good catch. Changed here: https://github.com/scrapinghub/scrapy-po/pull/5/commits/9983a04d037b612d51c887dd3e7db9a3a0f5c32a

scrapy_po/middleware.py
ivanprado: Done here: https://github.com/scrapinghub/scrapy-po/pull/5/commits/166f48fe3b47fc7963629867bff9010b382c1bef

scrapy_po/middleware.py
ivanprado: Andi was changed. This was addressed here: https://github.com/scrapinghub/scrapy-po/pull/5/commits/166f48fe3b47fc7963629867bff9010b382c1bef

scrapy_po/middleware.py
ivanprado: Done in https://github.com/scrapinghub/scrapy-po/pull/5/commits/61b1b53d3015cb40d17d212bd8e67eaf031cc0a7 and https://github.com/scrapinghub/scrapy-po/pull/5/commits/7d343c9dbede0cbf495b9d022f2816f37ca2fb33

example/example/spiders/books_06.py
ivanprado: Ok, I changed that way by now: https://github.com/scrapinghub/scrapy-po/pull/5/commits/4ed02bbc5b13f2e032a982f6b8b10e8171e26120

But once we change the spiders for Scrapy 2.0 we can rewrite it with the `follow_all`. Probably better to address this change of examples to Scrapy 2.0 in a different PR to simplify the approval of this one.

example/example/spiders/books_06.py
ivanprado: As you commented in a different PR, better to just migrate the examples to Scrapy 2.0. But let's do it better in a separate PR.

kmike:
Thanks @ivanprado, looks good!

PR #6, opened 2020-02-19
[WIP] Customizations based on the response

ivanprado:
Injected PageObjects can now be modified based on the response. This
allows to create specialized PO for each domain or based in the content
(i.e. use a particular pagination PO if a particular type of pagination
is detected in the content).

The developed mechanism is very low level at the moment. Some more helper helpers to simply configuration would be handy. For example to configure PO per domain, so that they can be configured with dicts like:
```py
{'domain.x.y': {
    BookPage: DomainXYBookPage, 
    BookPagination: DomainXYBookPagination
    }, 
 'domain.z.w': ...
}
```

Note that the base branch is not master, but `andi_recursive_injection`

@kmike what do you think?

### DIFF
--- a/example/example/customizations.py
+++ b/example/example/customizations.py
@@ -1,0 +1,15 @@
+from typing import Type
+
+from urllib.request import Request
+
+from example.spiders.books_06 import AsimovFoundationBookPage, BookPage
+from scrapy_po.customizations import Customizations
+
+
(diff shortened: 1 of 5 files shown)


ivanprado:
Closing in favour of https://github.com/scrapinghub/scrapy-poet/pull/43

PR #7, opened 2020-02-21
PageComponent introduced

ivanprado:
Question: is the name too long (`PageComponent`)? 

Alternatives::
* `Component`
* `PagePart`
* `PagePiece`

Note this is based on top of https://github.com/scrapinghub/scrapy-po/pull/5 that is in the review process.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -40,7 +40,7 @@ Usage
 First, enable middleware in your settings.py::
 
     DOWNLOADER_MIDDLEWARES = {
-       'scrapy_po.InjectPageObjectsMiddleware': 543,
+       'scrapy_po.InjectionMiddleware': 543,
     }
 
 After that you can write spiders which use page object pattern to separate
--- a/example/example/autoextract.py
+++ b/example/example/autoextract.py
@@ -10,7 +10,7 @@ from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.internet.threads import deferToThread
 
 from scrapy_po.page_input_providers import PageObjectInputProvider, provides
-from scrapy_po.webpage import PageObject
+from scrapy_po.webpage import Injectable, ItemPage
 
 
 @attr.s(auto_attribs=True)
@@ -38,7 +38,7 @@ def get_autoextract_product(url):
 
 
 @attr.s(auto_attribs=True)
-class ProductPage(PageObject):
+class ProductPage(ItemPage):
     """ Generic product page """
     autoextract_resp: AutoextractProductResponse
 
--- a/example/example/spiders/books_06.py
+++ b/example/example/spiders/books_06.py
@@ -15,31 +15,31 @@ import attr
 from scrapy_po import WebPage, ItemWebPage, Injectable
 
 
-class ListingsComponent(WebPage):
+class ListingsExtractor(WebPage):
     def urls(self):
         return self.css('.image_container a::attr(href)').getall()
 
 
-class PaginationComponent(WebPage):
+class PaginationExtractor(WebPage):
     def urls(self):
         return self.css('.pager a::attr(href)').getall()
 
 
-class BreadcrumbsComponent(WebPage):
+class BreadcrumbsExtractor(WebPage):
     def urls(self):
         return self.css('.breadcrumb a::attr(href)').getall()
 
 
 @attr.s(auto_attribs=True)
 class ListingsPage(Injectable):
-    book_list: ListingsComponent
-    pagination: PaginationComponent
+    book_list: ListingsExtractor
+    pagination: PaginationExtractor
 
 
 @attr.s(auto_attribs=True)
 class BookPage(ItemWebPage):
-    recently_viewed: ListingsComponent
-    breadcrumbs: BreadcrumbsComponent
+    recently_viewed: ListingsExtractor
+    breadcrumbs: BreadcrumbsExtractor
 
     def to_item(self):
         return {
@@ -62,14 +62,14 @@ class BooksSpider(scrapy.Spider):
         yield from self.follow_breadcrumbs(response, page.breadcrumbs)
         yield page.to_item()
 
-    def follow_listing(self, response, item_list: ListingsComponent):
+    def follow_listing(self, response, item_list: ListingsExtractor):
         for url in item_list.urls():
             yield response.follow(url, self.parse_book)
 
-    def follow_pagination(self, response, pagination: PaginationComponent):
+    def follow_pagination(self, response, pagination: PaginationExtractor):
         for url in pagination.urls():
             yield response.follow(url, self.parse, priority=+10)
 
-    def follow_breadcrumbs(self, response, breadcrumbs: BreadcrumbsComponent):
+    def follow_breadcrumbs(self, response, breadcrumbs: BreadcrumbsExtractor):
         for url in breadcrumbs.urls():
             yield response.follow(url, self.parse)
--- a/example/example/spiders/books_06.py
+++ b/example/example/spiders/books_06.py
@@ -12,32 +12,32 @@ All links to search result pages are followed as well.
 import scrapy
 import attr
 
-from scrapy_po import WebPage, ItemPage, PageComponent
+from scrapy_po import WebPage, ItemWebPage, Injectable
 
 
-class ListingsComponent(PageComponent):
+class ListingsComponent(WebPage):
     def urls(self):
         return self.css('.image_container a::attr(href)').getall()
 
 
-class PaginationComponent(PageComponent):
+class PaginationComponent(WebPage):
     def urls(self):
         return self.css('.pager a::attr(href)').getall()
 
 
-class BreadcrumbsComponent(PageComponent):
+class BreadcrumbsComponent(WebPage):
     def urls(self):
         return self.css('.breadcrumb a::attr(href)').getall()
 
 
 @attr.s(auto_attribs=True)
-class ListingsPage(WebPage):
+class ListingsPage(Injectable):
     book_list: ListingsComponent
     pagination: PaginationComponent
 
 
 @attr.s(auto_attribs=True)
-class BookPage(WebPage, ItemPage):
+class BookPage(ItemWebPage):
     recently_viewed: ListingsComponent
     breadcrumbs: BreadcrumbsComponent
 
--- a/example/example/spiders/books_06.py
+++ b/example/example/spiders/books_06.py
@@ -13,33 +13,34 @@ import scrapy
 import attr
 
 from scrapy_po import WebPage, ItemPage
+from scrapy_po.webpage import PageComponent
 
 
-class ListingsPiece(WebPage):
+class ListingsComponent(PageComponent):
     def urls(self):
         return self.css('.image_container a::attr(href)').getall()
 
 
-class PaginationPiece(WebPage):
+class PaginationComponent(PageComponent):
     def urls(self):
         return self.css('.pager a::attr(href)').getall()
 
 
-class BreadcrumbsPiece(WebPage):
+class BreadcrumbsComponent(PageComponent):
     def urls(self):
         return self.css('.breadcrumb a::attr(href)').getall()
 
 
 @attr.s(auto_attribs=True)
 class ListingsPage(WebPage):
-    book_list: ListingsPiece
-    pagination: PaginationPiece
+    book_list: ListingsComponent
+    pagination: PaginationComponent
 
 
 @attr.s(auto_attribs=True)
 class BookPage(WebPage, ItemPage):
-    recently_viewed: ListingsPiece
-    breadcrumbs: BreadcrumbsPiece
+    recently_viewed: ListingsComponent
+    breadcrumbs: BreadcrumbsComponent
 
     def to_item(self):
         return {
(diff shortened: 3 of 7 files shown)


scrapy_po/webpage.py
```diff
@@ -66,6 +66,20 @@ class WebPage(PageObject, ResponseShortcutsMixin):
...
+    """
+    Use it for implementing extractors for particular components of the
+    page (i.e. breadcrumbs) that can be reused across different page objects.
+
+    For example, a book page object can rely on breadcrumbs page component
+    to extract the book category and also a book listings page object could
+    rely in this component to extract the category a listings page
+    belongs to.
+    """
+    response: ResponseData
```
kmike: I think it makes sense to have some base class which doesn't have ResponseShortcutsMixin or ResponseData to inherit from. With the current implementation it won't make sense to inherit AutoExtract-based page components from this PageComponent class.

ivanprado: I see. Should we introduce two instead?:
```py
class PageComponent(PageObject):
   pass

@attr.s(auto_attribs=True)
class PageWebComponent(PageComponent, ResponseShortcutsMixin):
    response: ResponseData
```

Another option is to just provide a PageComponent extending from PageObject and leave  the user the responsibility to add the response and extend the mixin when they need web data:

```py
@attr.s(auto_attribs=True)
class MyBreadcrumbsComponent(PageComponent, ResponseShortcutsMixin):
    response: ResponseData
```

It doesn't look so easy. So I would go for the first option. @kmike What do you think?

kmike: I'd try to find a name for a base class which says "handle me in the dependency injection!", then inherit PageObject and PageComponent from this class, or just drop them both (as they do nothing), if the name is good enough.

```py
class ItemPage(Component, ABC):
    @abstractmethod
    def to_item(self) -> Dict:
        return {}


@attr.s(auto_attribs=True)
class WebPage(ItemPage, ResponseShortcutsMixin):
    response: ResponseData 


class BasePaginator(Component, ABC):
    @abstractmethod
    def page_urls(self) -> List[str]:
        return []
    


```

ivanprado: Well, what about `Injectable`. It seems very aligned with that it really means. Something injectable can be injected using dependency injection. 

And I would then remove `PageObject` and `PageComponent` at all. So I like your proposal but renaming `Component` by `Injectable`: 
```py
class ItemPage(Injectable, ABC):
    @abstractmethod
    def to_item(self) -> Dict:
        return {}


@attr.s(auto_attribs=True)
class WebPage(ItemPage, ResponseShortcutsMixin):
    response: ResponseData 


class BasePaginator(Injectable, ABC):
    @abstractmethod
    def page_urls(self) -> List[str]:
        return []
``` 
I see it used in othe libraries with the same meaning: https://angular.io/api/core/Injectable 

We could also rename `InjectPageObjectsMiddleware` to `InjectionMiddleware` so that everything gets more aligned. And then Page Objects is just a way of organizing the extraction code by declaring classes that are injectables and that get injected. 

How that is sound?

A possible alternative would be to mark classes as "injectable" using a decorator (`@Injectable`), but it would require to always use this decorator on the concrete classes. It seems better, in this case, to mark the pages by inheritance so that the "injectable" property is inherited, so I would discard it.

kmike: Names.. On one hand, Injectable looks good: these are things which we can inject. But we have 2 types of DI going on:

1) Injecting these PageObjects and components in callbacks
2) Injecting ResponseData, etc. to PageObjects and components

Do we need to separate them, what do you think? I.e. should ResponseData be Injectable? It seems to go against the current design, but I haven't though that much about it.

Marking classes as injectable using decorator is not needed, because you can already mark them using Injectable.register(MyClass), if you don't want to use inheritance for some reason (maybe the class is out of your control, you don't want to change it? not sure what's the real use case). This is a feature of Python's abc module. Injectable.register can also be used as a class decorator - exactly what you proposed here.  See https://docs.python.org/3/library/abc.html#abc.ABCMeta.register

ivanprado: I see two types of classes from the DI point of view:
1) Injectables. They are automatically built by the DI mechanism. Example: PageObjects
2) Providables. They are "external" classes that require a provider to be built, so they are not directly built by the DI, but instead rely on a  provider. Example: `ResponseData`, `AutoextractProductResponse`. 

Currently, it is not required to mark "providable" classes as "injectable", because an individual registry is kept. When the middleware is looking at which types can be provided, it is considering both types ("injectables" and "providables"). See https://github.com/scrapinghub/scrapy-po/blob/page_component/scrapy_po/middleware.py#L59-L66. (I know you already know, I'm writing it to sorting my mind)

Providable classes always take precedence over Injectable ones: if you found a provider for a class, you use it, even if the class is "Injectable". 

There is then no technical benefit in forcing the annotation of "providable" classes as "injectable" so the question is if "injectable" is a good name given that there are things (the "providables") that will be also be injected even without being "injectables". The two options I see to improve coherence are:
1) Force that all providable classes to be also Injectables (i.e. include a check that fails if they are not). 
2) Put a different name, like `BuildByInjectionButNotFromProvider` (obviously a horrendous name, but I can't think in something good meaning that). 

I don't like any of both options above. From my point of view, the ambiguity of `Injectable` is manageable, so I would stay with this proposal.  

Good point regarding that using `ABCMeta.register` can be used as a decorator. So we have then both options.

kmike: I also dislike forcing ResponseData and alike objects being Injectable.

Ok, I'm fine with proceeding with Injectable. Still not a fan, but it is definitely better than what we have now, and we can change the name later if a better proposal arrives, it is early days.

ivanprado: Another point against forcing ResponseData as being injectable: This objects that are "providable" can be optional (used in Unions), but that not's the case for the rest of injectable objects.

> Ok, I'm fine with proceeding with Injectable. Still not a fan

Yes, we can change it in the future. I'll change this PR to use `Injectable` by now.

ivanprado:
I have been iterating this. I'm still not happy with the naming, etc, but I have not found the solution. Some questions I have doubts about:
* "Page Objects" is a name that has a meaning under the original design where the goal was to encapsulate the extraction code for a particular page. But this concept is not so clear if individual components can be composed (see `ListinsPage` class https://github.com/scrapinghub/scrapy-po/blob/page_component/example/example/spiders/books_06.py#L33-L36). Maybe we shouldn't allow components at all (all should be monolithic Page objects), but by the other side components seem a good idea: a breadcrumbs component itself has a meaning and can be used both in detail pages or results pages. If we allow for components, then maybe the name of "Page Objects" is not well reflecting what this is about. Maybe "Extraction Objects". By the other side may "Page Objects" be referring to the fact that all extraction is happening within the context of a page, something like "Page Extraction Objects". 
* Components inherits from `WebPage`. Strange
* A new class `ItemWebPage` was required to define a `WebPage` that also implements the `to_item`. A little bit convoluted.

I'm going to leave some comments within the code explaining the different parts. Feel free to comment there or here.

README.rst
```diff
@@ -40,7 +40,7 @@ Usage
 First, enable middleware in your settings.py::
 
     DOWNLOADER_MIDDLEWARES = {
-       'scrapy_po.InjectPageObjectsMiddleware': 543,
+       'scrapy_po.InjectionMiddleware': 543,
```
ivanprado: I changed the name because I think the Middleware is doing more than just injecting PageObjects... it is injecting many things. But probably is also influenced because I have not clear the "Page Object" concept now that it can also deal with components.

example/example/autoextract.py
```diff
@@ -38,7 +38,7 @@ def get_autoextract_product(url):
 
 
 @attr.s(auto_attribs=True)
-class ProductPage(PageObject):
+class ProductPage(ItemPage):
```
ivanprado: Extending ItemPage instead because it is implementing ``to_item``

example/example/spiders/books_06.py
```diff
@@ -12,34 +12,34 @@
 import scrapy
 import attr
 
-from scrapy_po import WebPage, ItemPage
+from scrapy_po import WebPage, ItemWebPage, Injectable
 
 
-class ListingsPiece(WebPage):
+class ListingsExtractor(WebPage):
```
ivanprado: We would need some generic posfix to use in these classes that only deals with extraction from a part of a page. I used here `Extractor`.

example/example/spiders/books_06.py
```diff
@@ -12,34 +12,34 @@
...
+class BreadcrumbsExtractor(WebPage):
     def urls(self):
         return self.css('.breadcrumb a::attr(href)').getall()
 
 
 @attr.s(auto_attribs=True)
-class ListingsPage(WebPage):
-    book_list: ListingsPiece
-    pagination: PaginationPiece
+class ListingsPage(Injectable):
```
ivanprado: Note this class is a "Page Object" (i.e. the suffix in the name is "Page" because it extracts all that can be extracted from a listings page), but that it is not implementing the `to_item` (it was not required in the code), so it is extending `Injectable`. Probably not very easy to understand. 

By the other side it looks nice that it can be just the composition of two other components. This is an alternative to the monolithc version proposed in https://gist.github.com/kmike/2fa2b1b8c77bfb98f72b2636e66c5d36#id2 with the name of `BookListPage`.

example/example/spiders/books_06.py
```diff
@@ -12,34 +12,34 @@
...
+    book_list: ListingsExtractor
+    pagination: PaginationExtractor
 
 
 @attr.s(auto_attribs=True)
-class BookPage(WebPage, ItemPage):
-    recently_viewed: ListingsPiece
-    breadcrumbs: BreadcrumbsPiece
+class BookPage(ItemWebPage):
```
ivanprado: `ItemWebPage` was introduced because `BookPage` must be a `WebPage` but also it should be an `ItemPage`. An alternative is to declare it here:
```py
class BookPage(WebPage, ItemPage):
    ....
```

Another alternative would have been to declare `ItemPage` as a `WebPage`:
```py
@attr.s(auto_attribs=True)
class ItemPage(WebPage):
    @abc.abstractmethod
    def to_item(self):
        pass
```
But then it wouldn't be allowed `ItemPages` that don't require `ResponseData`, and this is required for autoextract.

scrapy_po/webpage.py
```diff
@@ -7,17 +7,18 @@
...
+    callbacks that requires them. The ``InjectionMiddleware`` take care of it.
 
-    Instead of inheriting you can also use ``PageObject.register(MyWebPage)``.
+    Instead of inheriting you can also use ``Injectable.register(MyWebPage)``.
+    ``Injectable.register`` can also be used as a decorator.
     """
     pass
 
 
-class ItemPage(PageObject):
```
ivanprado: PageObject disappears as an entity in the code. If we use the term in the comments, etc, we should make clear what "Page Object" means.

README.rst
kmike: I like the new name 👍

example/example/autoextract.py
```diff
@@ -10,7 +10,7 @@
 from twisted.internet.threads import deferToThread
 
 from scrapy_po.page_input_providers import PageObjectInputProvider, provides
-from scrapy_po.webpage import PageObject
+from scrapy_po.webpage import Injectable, ItemPage
```
kmike: Injectable import is not needed

kmike:
> "Page Objects" is a name that has a meaning under the original design where the goal was to encapsulate the extraction code for a particular page. But this concept is not so clear if individual components can be composed (see ListinsPage class https://github.com/scrapinghub/scrapy-po/blob/page_component/example/example/spiders/books_06.py#L33-L36). Maybe we shouldn't allow components at all (all should be monolithic Page objects), but by the other side components seem a good idea: a breadcrumbs component itself has a meaning and can be used both in detail pages or results pages. If we allow for components, then maybe the name of "Page Objects" is not well reflecting what this is about. Maybe "Extraction Objects". By the other side may "Page Objects" be referring to the fact that all extraction is happening within the context of a page, something like "Page Extraction Objects".

Page Object is also a name of a pattern - see https://martinfowler.com/bliki/PageObject.html. This is complicated :) From this link:

> Despite the term "page" object, these objects shouldn't usually be built for each page, but rather for the significant elements on a page [There's an argument here that the name "page object" is misleading because it makes you think you should have just one page object per page. Something like "panel object" would be better - but the term "page object" is what's become accepted.] 

So, according to this, that's fine to call an object which provides an interface to a part of a page PageObject. 

> Components inherits from WebPage. Strange

Yeah.. It is actually even more strange than inheriting from PageObject. Maybe WebPage name is not good.

> A new class ItemWebPage was required to define a WebPage that also implements the to_item. A little bit convoluted.

Looks fine to me.

example/example/spiders/books_06.py
kmike: `Extractor` sounds fine.

ivanprado:
Thank you for the Martin Fowler link, now I'm understanding better where the concept comes from. I googled it before but didn't find something relevant. 

Now that there is some sort of agreement in the `Injection` name by now, I'm going to merge this PR into #5 to continue there with the changes to adapt to new `andi` API. Keeping both PRs would make things more complex.

PR #9, opened 2020-02-25
requests method to follow multiple urls

ivanprado:
**This is based on unit_test branch**

Inspired in: https://github.com/scrapinghub/scrapy-po/pull/5/files/b5f6f6374a48fdb40ae7532b0256c5f46e0449d5#r382783105

See changes in `book_06.py` and `book_06_2.py`  to see it applied to some spiders. 

Clearly, the spider looks better, very simplified. I like `book_06_2.py` more. 

Note that for resolving the url with `urljoin` the attribute `url` is used. So the class using the new mixin should:
* have a `url()` method. (i.e. be a `WebPage`)
* implement the `urls()` method. 

I wonder if there is a different place to obtain the url to call `urljoin`, because for example a page object from autoextract wouldn't be able to use this new mixin. 

Also, wonder if we shouldn't create some kind of abstract class, or in the same mixing, that declares `urls` method. It would make it more explicit. 

What do you think? Once we have clear the way to go I'll clean up the code and write some tests.


scrapy_po/requests_mixin.py
```diff
@@ -0,0 +1,26 @@
...
+
+class RequestsFromUrlsMixin():
+    """ Add a method `requests` that wraps urls returned by method `urls`
+    into ``Request``s and yield it so that can be used in a Scrapy callback
+    in the form of ``yield from obj.requests(...args...). Note that this
+    mixin expects the parent to have the ``urls`` and ``url`` methods defined """
+
+    def requests(self, *args, **kwargs):
+        for url in self.urls():
+            yield Request(urljoin(self.url, url),
```
kmike: response.urljoin is better, because it handles base url properly

kmike: Sorry, I should have thought about it more. I think switching to Scrapy 2.0 in examples is better than adding custom shortcuts to the codebase.

ivanprado:
Closing as is better just to use the Scrapy 2.0 `follow_all`

PR #10, opened 2020-04-09
Skipping downloads when they're not necessary to execute parsers

victor-torres:
This is done by inspecting parsers and injectable dependencies.
We search for type annotations and the use of DummyRequests.

### DIFF
--- a/example/example/autoextract.py
+++ b/example/example/autoextract.py
@@ -10,7 +10,8 @@ from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.internet.threads import deferToThread
 
 from scrapy_po.page_input_providers import PageObjectInputProvider, provides
-from scrapy_po.webpage import Injectable, ItemPage
+from scrapy_po.webpage import ItemPage
+from scrapy_po.utils import DummyResponse
 
 
 @attr.s(auto_attribs=True)
@@ -21,6 +22,10 @@ class AutoextractProductResponse:
 
 @provides(AutoextractProductResponse)
 class AutoextractProductProvider(PageObjectInputProvider):
+
+    def __init__(self, response: DummyResponse):
+        self.response = response
+
     @inlineCallbacks
     def __call__(self):
         data = (yield get_autoextract_product(self.response.url))
--- a/example/example/autoextract.py
+++ b/example/example/autoextract.py
@@ -35,7 +35,6 @@ class AutoextractProductProvider(PageObjectInputProvider):
 
 @inlineCallbacks
 def get_autoextract_product(url):
-    raise returnValue({'product': {'url': url, 'name': url}})
     # fixme: use async
     # fixme: rate limits?
     from autoextract.sync import request_batch
--- a/example/example/autoextract.py
+++ b/example/example/autoextract.py
@@ -30,6 +30,7 @@ class AutoextractProductProvider(PageObjectInputProvider):
 
 @inlineCallbacks
 def get_autoextract_product(url):
+    raise returnValue({'product': {'url': url, 'name': url}})
     # fixme: use async
     # fixme: rate limits?
     from autoextract.sync import request_batch
--- a/example/example/spiders/books_05_1.py
+++ b/example/example/spiders/books_05_1.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+"""
+Scrapy spider which uses Page Objects both for crawling and extraction.
+You can mix various page types freely.
+"""
+from typing import Optional
+
+import scrapy
+
+from scrapy_po import WebPage
+from scrapy_po.utils import DummyResponse
+from example.autoextract import ProductPage
+
+
+class BookListPage(WebPage):
+    def product_urls(self):
+        return self.css('.image_container a::attr(href)').getall()
+
+
+class BookPage(ProductPage):
+    def to_item(self):
+        # post-processing example: return only 2 fields
+        book = super().to_item()
+        return {
+            'url': book['url'],
+            'name': book['name'],
+        }
+
+
+class BooksSpider(scrapy.Spider):
+    name = 'books_05_1'
+    start_urls = ['http://books.toscrape.com/']
+
+    def parse(self, response, page: BookListPage):
+        for url in page.product_urls():
+            yield response.follow(url, self.parse_book)
+
+    # Bypassing download using DummyResponse since we'll be using AutoExtract.
+    def parse_book(self, response: DummyResponse, page: BookPage):
+        # you can also post-process data in a spider
+        book = page.to_item()
+        book['title'] = book.pop('name')
+        yield book
--- a/example/example/spiders/books_05_1.py
+++ b/example/example/spiders/books_05_1.py
@@ -2,8 +2,15 @@
 """
 Scrapy spider which uses Page Objects both for crawling and extraction.
 You can mix various page types freely.
+
+This example shows how to skip request downloads using DummyResponse type
+annotations. You should use this type annotation in your callback method
+and make sure that all of your providers don't require the use of a regular
+Response injection.
+
+Check the ``example.autoextract.AutoextractProductProvider.__init__`` method
+and the ``BooksSpider.parse_book`` callback for implementation details.
 """
-from typing import Optional
 
 import scrapy
 
--- a/scrapy_po/middleware.py
+++ b/scrapy_po/middleware.py
@@ -1,17 +1,12 @@
 # -*- coding: utf-8 -*-
-import inspect
-from typing import Dict, Callable, Type, Tuple
 
 from scrapy.utils.defer import maybeDeferred_coro
 
 import andi
 from twisted.internet.defer import inlineCallbacks, returnValue
 from scrapy import Request
-from scrapy.http import Response, TextResponse
 
-from .webpage import Injectable
-from .utils import get_callback, DummyResponse
-from .page_input_providers import providers
+from scrapy_po import utils
 
 
 class InjectionMiddleware:
@@ -23,31 +18,6 @@ class InjectionMiddleware:
 
     XXX: should it really be a downloader middleware?
     """
-    def is_response_going_to_be_used(self, request, spider):
-        """Check whether the request's response is going to be used."""
-        callback = get_callback(request, spider)
-        plan, _ = build_plan(callback, {})
-        for obj, _ in plan:
-            spec = inspect.getfullargspec(obj)
-            if 'response' not in spec.args:
-                # There's not argument named response, let's continue.
-                continue
-
-            if 'response' not in spec.annotations:
-                # There's no type annotation for the response argument,
-                # so we cannot infer that it's not going to be used.
-                return True
-
-            if not issubclass(spec.annotations['response'], DummyResponse):
-                # The response is not a DummyResponse, so it's probably used.
-                return True
-
-        # The parser method is using a DummyResponse as a type annotation.
-        # This suggests that the response might not get used.
-        # Also, we were not able to find any evidence that the response is
-        # going to be used by any injected Page Object.
-        return False
-
     def process_request(self, request: Request, spider):
         """Check if the request is needed and if the download can be skipped.
 
@@ -55,23 +25,23 @@ class InjectionMiddleware:
         by its designated parser or an injected Page Object.
 
         If we evaluate that the request could be ignored, we return a
-        DummyResponse object linked to the original Request instance.
+        utils.DummyResponse object linked to the original Request instance.
 
         With this behavior we're able to optimize spider executions avoid
         having to download URLs twice or when they're not needed, for example,
         when a Page Object relies only on a third-party API like AutoExtract.
         """
-        if self.is_response_going_to_be_used(request, spider):
+        if utils.is_response_going_to_be_used(request, spider):
             return
 
         spider.logger.debug(f'Skipping download of {request}')
-        return DummyResponse(url=request.url, request=request)
+        return utils.DummyResponse(url=request.url, request=request)
 
     @inlineCallbacks
     def process_response(self, request: Request, response, spider):
         # find out the dependencies
-        callback = get_callback(request, spider)
-        plan, provider_instances = build_plan(callback, response)
+        callback = utils.get_callback(request, spider)
+        plan, provider_instances = utils.build_plan(callback, response)
 
         # Build all instances declared as dependencies
         instances = yield from build_instances(
@@ -87,18 +57,6 @@ class InjectionMiddleware:
         raise returnValue(response)
 
 
-def build_plan(callback, response
-               ) -> Tuple[andi.Plan, Dict[Type, Callable]]:
-    """ Build a plan for the injection in the callback """
-    provider_instances = build_providers(response)
-    plan = andi.plan(
-        callback,
-        is_injectable=is_injectable,
-        externally_provided=provider_instances.keys()
-    )
-    return plan, provider_instances
-
-
 @inlineCallbacks
 def build_instances(plan: andi.Plan, providers):
     """ Build the instances dict from a plan """
@@ -109,17 +67,3 @@ def build_instances(plan: andi.Plan, providers):
         else:
             instances[cls] = cls(**kwargs_spec.kwargs(instances))
     raise returnValue(instances)
-
-
-def build_providers(response) -> Dict[Type, Callable]:
-    # find out what resources are available
-    return {cls: provider(response)
-            for cls, provider in providers.items()}
-
-
-def is_injectable(argument_type: Callable) -> bool:
-    """
-    A type is injectable if inherits from ``Injectable``.
-    """
-    return (isinstance(argument_type, type) and
-            issubclass(argument_type, Injectable))
--- a/scrapy_po/middleware.py
+++ b/scrapy_po/middleware.py
@@ -11,7 +11,7 @@ from scrapy.http import Response, TextResponse
 
 from .webpage import Injectable
 from .utils import get_callback, DummyResponse
-from.page_inputs import ResponseData
+from .page_inputs import ResponseData
 from .page_input_providers import providers
 
 
--- a/scrapy_po/middleware.py
+++ b/scrapy_po/middleware.py
@@ -11,7 +11,6 @@ from scrapy.http import Response, TextResponse
 
 from .webpage import Injectable
 from .utils import get_callback, DummyResponse
-from .page_inputs import ResponseData
 from .page_input_providers import providers
 
 
--- a/scrapy_po/middleware.py
+++ b/scrapy_po/middleware.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import inspect
 from typing import Dict, Callable, Type, Tuple
 
 from scrapy.utils.defer import maybeDeferred_coro
@@ -6,9 +7,11 @@ from scrapy.utils.defer import maybeDeferred_coro
 import andi
 from twisted.internet.defer import inlineCallbacks, returnValue
 from scrapy import Request
+from scrapy.http import Response, TextResponse
 
 from .webpage import Injectable
-from .utils import get_callback
+from .utils import get_callback, DummyResponse
+from.page_inputs import ResponseData
 from .page_input_providers import providers
 
 
@@ -20,9 +23,51 @@ class InjectionMiddleware:
     for this argument.
 
     XXX: should it really be a downloader middleware?
-    XXX: can this middleware allow to skip downloading
-    a page if it is not needed?
     """
+    def is_response_going_to_be_used(self, request, spider):
+        """Check whether the request's response is going to be used."""
+        callback = get_callback(request, spider)
+        plan, _ = build_plan(callback, {})
+        for obj, _ in plan:
+            spec = inspect.getfullargspec(obj)
+            if 'response' not in spec.args:
+                # There's not argument named response, let's continue.
+                continue
+
+            if 'response' not in spec.annotations:
+                # There's no type annotation for the response argument,
+                # so we cannot infer that it's not going to be used.
+                return True
+
+            if not issubclass(spec.annotations['response'], DummyResponse):
+                # The response is not a DummyResponse, so it's probably used.
+                return True
+
+        # The parser method is using a DummyResponse as a type annotation.
+        # This suggests that the response might not get used.
+        # Also, we were not able to find any evidence that the response is
+        # going to be used by any injected Page Object.
+        return False
+
+    def process_request(self, request: Request, spider):
+        """Check if the request is needed and if the download can be skipped.
+
+        Here we try to infer if the request's response is going to be used
+        by its designated parser or an injected Page Object.
+
+        If we evaluate that the request could be ignored, we return a
+        DummyResponse object linked to the original Request instance.
+
+        With this behavior we're able to optimize spider executions avoid
+        having to download URLs twice or when they're not needed, for example,
+        when a Page Object relies only on a third-party API like AutoExtract.
+        """
+        if self.is_response_going_to_be_used(request, spider):
+            return
+
+        spider.logger.debug(f'Skipping download of {request}')
+        return DummyResponse(url=request.url, request=request)
+
     @inlineCallbacks
     def process_response(self, request: Request, response, spider):
         # find out the dependencies
--- a/scrapy_po/middleware.py
+++ b/scrapy_po/middleware.py
@@ -1,8 +1,5 @@
 # -*- coding: utf-8 -*-
 
-from scrapy.utils.defer import maybeDeferred_coro
-
-import andi
 from twisted.internet.defer import inlineCallbacks, returnValue
 from scrapy import Request
 
@@ -44,7 +41,7 @@ class InjectionMiddleware:
         plan, provider_instances = utils.build_plan(callback, response)
 
         # Build all instances declared as dependencies
-        instances = yield from build_instances(
+        instances = yield from utils.build_instances(
             plan.dependencies, provider_instances)
 
         # Fill the callback arguments with the created instances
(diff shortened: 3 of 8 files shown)


example/example/autoextract.py
```diff
@@ -30,6 +30,7 @@ def __call__(self):
 
 @inlineCallbacks
 def get_autoextract_product(url):
+    raise returnValue({'product': {'url': url, 'name': url}})
```
kmike, line 33: Could you please remove this change? I see how it was useful for testing, but let me send you an API key instead.

example/example/spiders/books_05_1.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+"""
+Scrapy spider which uses Page Objects both for crawling and extraction.
+You can mix various page types freely.
```
kmike, line 4: Could you please mention that this example shows how to skip scrapy Response?

scrapy_po/middleware.py
```diff
@@ -1,14 +1,17 @@
...
 
 import andi
 from twisted.internet.defer import inlineCallbacks, returnValue
 from scrapy import Request
+from scrapy.http import Response, TextResponse
 
 from .webpage import Injectable
-from .utils import get_callback
+from .utils import get_callback, DummyResponse
+from.page_inputs import ResponseData
```
kmike, line 14:
```suggestion
from .page_inputs import ResponseData
```

scrapy_po/middleware.py
```diff
@@ -20,9 +23,51 @@ class InjectionMiddleware:
...
+                # The response is not a DummyResponse, so it's probably used.
+                return True
+
+        # The parser method is using a DummyResponse as a type annotation.
+        # This suggests that the response might not get used.
+        # Also, we were not able to find any evidence that the response is
+        # going to be used by any injected Page Object.
+        return False
+
+    def process_request(self, request: Request, spider):
```
kmike, line 52: 👍

scrapy_po/middleware.py
```diff
@@ -20,9 +23,51 @@ class InjectionMiddleware:
     for this argument.
 
     XXX: should it really be a downloader middleware?
-    XXX: can this middleware allow to skip downloading
-    a page if it is not needed?
     """
+    def is_response_going_to_be_used(self, request, spider):
+        """Check whether the request's response is going to be used."""
+        callback = get_callback(request, spider)
+        plan, _ = build_plan(callback, {})
```
kmike, line 30: 👍 for making it work end-to-end, but it seems the current implementation is not complete.

There are 2 things to consider:

1) have user annotated response as DummyResponse in a callback?
2) does a dependency of any of the arguments need to access scrapy Response (e.g. page object asks for ReesponseData)?

Then I think actions should be these:

1. User haven't annotated response as DummyResponse. Always download response, pass it to the callback.
2. There is response: DummyResponse; page objects don't need response. Don't download; pass DummyResponse instance to the callback.
3. There is response: DummyResponse; page objects need a Response. Download a response, but **pass DummyResponse instance to the callback**, as user asked for that.

A next question is how you detect

1) if response is annotated as DummyResponse, and
2) if a page object needs a response.

Currently this is done by checking if there is an argument named `response`. I think this is not correct, both for callback and for page objects, but for different reasons.

Callback: Scrapy passes response as a positional argument, not as a keyword argument. It means the name of the argument doesn't matter: in case of `def parse(self, resp, page: ProductPage, response: DummyResponse):` response should be downloaded. 

Page objects / injectables: argument names and types doen't matter; what matters is an implementation of providers for the types which page object uses. Examples:
* There can be e.g. page object which receives `response` argument of type SplashResponse. 
* There can be a page object which receives `resp` argument of type ResponseData, and the provider for ResponseData needs an access to response. 
* There can be a page object which receives `response` argument of type ResponseData, but a provider which is currently registered for ResponseData doesn't use Scrapy at all (e.g. it gets a response from a local cache).

So, if I'm not mistaken, It means that the logic for callback and for page objects should be different:

1. For callback - check type of the first argument.
2. For page object - find providers for all dependencies, and check if any of the providers need a Response.

The tricky part here is "if any of the providers need a Response"; currently there is no way of doing it. I think we can use dependency injection again here, and inspect their `__init__` method; if Response or its subclass is found, then mark the provider as requiring a Response. This (or some other way of doing it - e.g. having an attribute `requires_scrapy_response = True` should be documented.

scrapy_po/middleware.py
```diff
@@ -20,9 +23,51 @@ class InjectionMiddleware:
     for this argument.
 
     XXX: should it really be a downloader middleware?
-    XXX: can this middleware allow to skip downloading
-    a page if it is not needed?
     """
+    def is_response_going_to_be_used(self, request, spider):
```
kmike, line 27: 👍 for extracting it, and not having the logic in process_request method. 

Could you please also write some tests for it? 

It seems this method doesn't need `self`, so you can move/refactor it to separate function(s), and test in isolation - e.g. detection of cases like `def parse(self, resp, page: ProductPage, response: DummyResponse):`. It looks feasible to write tests involving page objects as well.

example/example/autoextract.py
victor-torres: Sure, was using it just for tests. Was expecting an initial review before removing it.

example/example/spiders/books_05_1.py
victor-torres: Done.

scrapy_po/middleware.py
victor-torres: I've just refactored the source code. Could you please take another review?

scrapy_po/middleware.py
victor-torres: I've introduced new test cases. Feedback is appreciated 😄

victor-torres:
@kmike I'll wait for your feedback before implementing more test cases (for example, testing middleware and parts pointed by coverage bot).

scrapy_po/utils.py
```diff
@@ -1,7 +1,122 @@
...
+
+    def __init__(self, url, request=None):
+        super(DummyResponse, self).__init__(url=url, request=request)
+
+
+def is_callback_using_response(callback: Callable):
+    """Check whether the request's callback method is going to use response."""
+    spec = inspect.getfullargspec(callback)
+    try:
+        arg_name = spec.args[1]  # first index is self, second is response
```
kmike, line 31: Just to be sure, could you please check (add a test?) how it works with e.g. partial callbacks:

```py

from functools import partial

    def start_requests(self):
        yield Request(partial(self.parse, use_x=True))

    def parse(self, response, use_x):
        # ...
```

Users should be using cb_kwargs, but partial or lambdas may be used in a real code.

scrapy_po/utils.py
```diff
@@ -1,7 +1,122 @@
...
+
+    if issubclass(spec.annotations[arg_name], DummyResponse):
+        # Type annotation is DummyResponse, so we're probably NOT using it.
+        return False
+
+    # Type annotation is not DummyResponse, so we're probably using it.
+    return True
+
+
+def are_dependencies_using_response(plan: andi.Plan):
```
kmike, line 49: a minor thing, but what do you think about making `providers` an argument for this function, to get further from singletons, and potentially simplify testing?

scrapy_po/utils.py
```diff
@@ -1,7 +1,122 @@
...
+    """Check whether any injectable provider makes use of a valid Response."""
+    for obj, _ in plan:
+        provider = providers.get(obj)
+        if not provider:
+            # Provider not found.
+            continue
+
+        spec = inspect.getfullargspec(provider)
```
kmike, line 57: I haven't checked this in detail, but it can be better to use andi.inspect here, because this way you can work around some attr.s-specific issues, and potentially support Union

scrapy_po/utils.py
```diff
@@ -1,7 +1,122 @@
...
+            return True
+
+    # Could not find any Response type annotation in the used providers.
+    return False
+
+
+def is_response_going_to_be_used(request, spider):
+    """Check whether the request's response is going to be used."""
+    callback = get_callback(request, spider)
+    plan, _ = build_plan(callback, {})
```
kmike, line 78: could you please explain, why is ythe second argument `{}`?

scrapy_po/utils.py
```diff
@@ -1,7 +1,122 @@
...
+
+    if is_callback_using_response(callback):
```
kmike, line 80: it seems this check can be done before creating a plan, so that you don't need to call build_plan if it passes

example/example/autoextract.py
```diff
@@ -21,6 +22,10 @@ class AutoextractProductResponse:
 
 @provides(AutoextractProductResponse)
 class AutoextractProductProvider(PageObjectInputProvider):
+
+    def __init__(self, response: DummyResponse):
```
kmike, line 26: What do you think about making it explicit that this thing needs an URL, not a DummyResponse? Not sure how best to do it.

example/example/autoextract.py
```diff
@@ -10,7 +10,8 @@
 from twisted.internet.threads import deferToThread
 
 from scrapy_po.page_input_providers import PageObjectInputProvider, provides
-from scrapy_po.webpage import Injectable, ItemPage
+from scrapy_po.webpage import ItemPage
+from scrapy_po.utils import DummyResponse
```
kmike, line 14: Technically, DummyResponse could belong to page_inputs.py module, but I see why you put it to utils. No specific suggestions here :)

tests/test_utils.py
```diff
@@ -1,6 +1,60 @@
...
+class DummyProductResponse:
+
+    data: Dict[str, Any]
+
+
+@provides(DummyProductResponse)
+class DummyProductProvider(PageObjectInputProvider):
+
+    def __init__(self, response: DummyResponse):
+        self.response = response
```
kmike, line 28: It seems this example justifies a refactoring of PageObjectInputProvider: base class here shouldn't require a response, you shouldn't be overriding argument type just to disable base class behavior. Here probably you can get away with an argument-less `__init__`. I was thinking about the same for the autoextract example, but there response was actually needed, to get an url.

kmike:
@victor-torres the approach and proposed code refactoring makes sense to me. Could you please add some integration-style tests as well, to check that response is actually not being downloaded in cases where it shouldn't?

scrapy_po/utils.py
victor-torres: Refactored.

scrapy_po/utils.py
victor-torres: Output is quite different :/

```(Pdb) inspect.getfullargspec(provider)
FullArgSpec(args=['self', 'response'], varargs=None, varkw=None, defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={'response': <class 'scrapy.http.response.Response'>})
(Pdb) andi.inspect(provider)
{'response': []}
```

scrapy_po/utils.py
victor-torres: Well, it seems like Andi does not support partials:

```
Traceback (most recent call last):
  File "/Users/victortorres/.virtualenvs/scrapy-po/lib/python3.7/site-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "/Users/victortorres/.virtualenvs/scrapy-po/lib/python3.7/site-packages/twisted/python/failure.py", line 512, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/Users/victortorres/.virtualenvs/scrapy-po/lib/python3.7/site-packages/scrapy/core/downloader/middleware.py", line 51, in process_response
    response = yield deferred_from_coro(method(request=request, response=response, spider=spider))
  File "/Users/victortorres/.virtualenvs/scrapy-po/lib/python3.7/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/Users/victortorres/git/shub/scrapy-po/scrapy_po/middleware.py", line 30, in process_response
    plan, provider_instances = build_plan(callback, response)
  File "/Users/victortorres/git/shub/scrapy-po/scrapy_po/middleware.py", line 53, in build_plan
    externally_provided=provider_instances.keys()
  File "/Users/victortorres/.virtualenvs/scrapy-po/lib/python3.7/site-packages/andi/andi.py", line 264, in plan
    dependency_stack=None)
  File "/Users/victortorres/.virtualenvs/scrapy-po/lib/python3.7/site-packages/andi/andi.py", line 294, in _plan
    arguments = inspect(class_or_func)
  File "/Users/victortorres/.virtualenvs/scrapy-po/lib/python3.7/site-packages/andi/andi.py", line 24, in inspect
    annotations = get_type_hints(func, globalns)
  File "/Users/victortorres/miniconda3/lib/python3.7/typing.py", line 993, in get_type_hints
    'or function.'.format(obj))
TypeError: functools.partial(<bound method PartialCallbackSpider.parse of <PartialCallbackSpider 'test_spider' at 0x10f487dd8>>, use_x=True) is not a module, class, method, or function.
```

scrapy_po/utils.py
victor-torres: So that we don't try to inject unavailable dependencies.

Look what happens if I remove the empty dict argument:

```
../tests/test_utils.py:114: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

request = <GET http://example.com>, spider = <MySpider 'foo' at 0x107f17908>

    def is_response_going_to_be_used(request, spider):
        """Check whether the request's response is going to be used."""
        callback = get_callback(request, spider)
        if is_callback_using_response(callback):
            return True
    
>       plan, _ = build_plan(callback)
E       TypeError: build_plan() missing 1 required positional argument: 'response'

../scrapy_po/utils.py:75: TypeError
```

We could try to pass a fake Response, but I don't think that scales if we have different dependency requirements in the future. So, as we're just trying to check provider classes, I'm passing an empty dict here.

scrapy_po/utils.py
victor-torres: Done.

example/example/autoextract.py
victor-torres: I'm not sure. In any case, DummyResponse instance does have a valid URL and references the original request so that users/developers could consult meta and other attributes.

tests/test_utils.py
victor-torres: Done. 

Currently, the source code only injects one dependency, the Response. There was already a comment on the source code telling us to "find out what resources are available". I guess we'll do it later.

scrapy_po/utils.py
kmike: could you please raise an issue in andi repo?

victor-torres:
Hi @kmike 

Thank you for you last review. 

I've just updated the pull request with some comments answering some of your questions and I have also introduced new test cases and refactored some parts of the code. 

Looking forward to your thoughts.

scrapy_po/utils.py
kmike: I think I'mbeing confused by the types build_plan function receives. If response is supposed to be a Response instance, wouldn't it be cleannerto pass None, not an empty dict?

scrapy_po/utils.py
kmike: Unlike andi.plan, you need to pass `__init__` method explicitly:

```
In [1]: import andi

In [2]: from scrapy.http  import Response

In [3]: class Foo:
   ...:     def __init__(self, resp: Response):
   ...:         pass
   ...:

In [4]: andi.inspect(Foo)
Out[4]: {'resp': []}

In [5]: andi.inspect(Foo.__init__)
Out[5]: {'resp': [scrapy.http.response.Response]}
```

scrapy_po/utils.py
victor-torres: Done https://github.com/scrapinghub/andi/issues/16

scrapy_po/utils.py
kmike: I actually wonder if this logic (https://github.com/scrapinghub/andi/blob/e26bcfd9f946aead542cacac1fa4462af7d01443/andi/andi.py#L289) should be moved to andi.inspect, so that there is less surprises. What do you think @ivanprado ?

scrapy_po/utils.py
victor-torres: No, no. I was thinking about `andi.plan` instead of `build_plan`. The later it "our" method. I'm refactoring to use `None` instead of `{}`. Thanks for insisting :)

victor-torres: Done.

scrapy_po/utils.py
kmike: Thanks @victor-torres!

What about lambdas?

scrapy_po/utils.py
kmike: Thanks for the quick fix :) 

By the way, you can treat andi as "ours" as well; it exists because of scrapy-po work, they're being developed together.

scrapy_po/utils.py
victor-torres: @kmike I've just refactored this method to make use of `andi.inspect` instead of `inspect.getfullargspec` as you had requested.

scrapy_po/utils.py
```diff
@@ -1,7 +1,131 @@
...
+
+class DummyResponse(Response):
+
+    def __init__(self, url, request=None):
+        super(DummyResponse, self).__init__(url=url, request=request)
+
+
+def is_callback_using_response(callback: Callable):
+    """Check whether the request's callback method is going to use response."""
+    spec = inspect.getfullargspec(callback)
```
victor-torres, line 29: I cannot use `andi.inspect` here since I depend on the order of the first argument and other data that only this built-in method is capable of providing. @kmike

scrapy_po/utils.py
victor-torres: It seems like it's not possible to extract type annotations from lambda functions. 

I have tried to annotate them this way (found an example on StackOverflow) but it doesn't seem to work.
```
class LambdaSkipDownloadSpider(scrapy.Spider):

    url = None

    def start_requests(self):
        cb: Callable[[DummyResponse], Dict] = lambda response: {'response': response}
        yield Request(url=self.url, callback=cb)
```

Inspect results:

```
(Pdb) spec
FullArgSpec(args=['response'], varargs=None, varkw=None, defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={})
(Pdb) callback
<function LambdaSkipDownloadSpider.start_requests.<locals>.<lambda> at 0x108da3bf8>
```

Any suggestion?

kmike: Aha, a good call. Let's document somewhere that lambdas don't work, and won't work - not sure where.

scrapy_po/utils.py
kmike: Yes, this makes sense; no need to use andi here, because the logic is very different.

scrapy_po/utils.py
```diff
@@ -1,7 +1,131 @@
...
+        return False
+
+    # Type annotation is not DummyResponse, so we're probably using it.
+    return True
+
+
+def is_provider_using_response(provider):
+    """Check whether injectable provider makes use of a valid Response."""
+    for argument, possible_types in andi.inspect(provider.__init__).items():
+        for cls in possible_types:
```
ivanprado, line 52: I think we should fail with a useful message here in case `cls` is not expected (i.e. it is not any type of Response).

ivanprado:
@victor-torres thanks for the PR, nice to meet you!. I see that this PR introduces some kind of manual `injection` into the providers (setting the request or the dummy request). I wonder if we could extend the concept and make the providers also injectable. I thought a little bit about it but I didn't found the answer, so leaving there the question open.

scrapy_po/utils.py
victor-torres: I've put something in the README.rst file.

scrapy_po/utils.py
victor-torres: Are we allowing other types of injectables in the future? 

Perhaps it's out of the scope of this method having to check this. What do you think? Should we raise a TypeError here?

victor-torres:
> @victor-torres thanks for the PR, nice to meet you!. I see that this PR introduces some kind of manual `injection` into the providers (setting the request or the dummy request). I wonder if we could extend the concept and make the providers also injectable. I thought a little bit about it but I didn't found the answer, so leaving there the question open.

Hey, @ivanprado. Thank you for your comments. 

Would you mind explaining a little bit more about your thoughts? Maybe giving an example?

scrapy_po/utils.py
ivanprado: I'm thinking in the case where somebody declares a provider with a wrong `__init__`. It would be more meaningful to fail here with a TypeError saying something like: `Type foo not supported for arguments for providers __init__. Only Response is supported.` or `Too many arguments for provider __init__, only empty or one supported`. 

Otherwise, it will fail on invocation but the message will be much less informative. What do you think?

ivanprado:
@victor-torres this is just an idea, so I don't want to mess this PR with it. Is out of the scope of it. 

But let's imagine we have some kind of `Conf` object that allows configuring some particular behaviour. You could declare a provider as usual:

```py
@provides(AutoextractProductResponse)
class AutoextractProductProvider(PageObjectInputProvider):

    def __init__(self, response: DummyResponse):
        self.response = response
    ...
```
  
Or alternatively, if you require the configuration to configure some behaviour, you could declare:

```py
@provides(AutoextractProductResponse)
class AutoextractProductProvider(PageObjectInputProvider):

    def __init__(self, response: DummyResponse, conf: Conf):
        self.response = response
        self.conf = conf
    ...
```

You would need also a provider for `Conf`:

```py
@provides(Conf)
class ConfProvider(PageObjectInputProvider):
     # It could read a file or could access a database depending the environment
```

This would require some injection mechanism using `andi.plan` similar to what is applied for the callback, but applied to the provider itself. 

A difficulty is that some providers might depend on other providers. 

Anyway, I'm not sure we really need something like that. It is only an idea. Forget about it.

victor-torres:
Oh, I see @ivanprado. That also crosses my mind when developing the `build_providers` method. Agree that it might be out of current scope, but it's a very interesting feature :D

scrapy_po/utils.py
victor-torres: Well, it makes sense. But I still think it's out of the scope of this method.

Think of that Config injectable you've suggested in another comment. Does that make sense to check for sub-classes of this Config here too? I don't think so. 

But let me know if you have a different opinion. Also, your two cents here would be helpful, @kmike.

ivanprado: Ok, forget about it. I think I'm messing things here :stuck_out_tongue_closed_eyes:  and this is very minor.

kmike:
Thanks @victor-torres! The PR looks good to me as-is, let's merge it.
It seems we all have many ideas for improvements, which is good :)

ivanprado:
I noticed that when `DummyRequest` is used, some functionality of Scrapy does not work:

* `CONCURRENT_REQUESTS`
* `CONCURRENT_REQUESTS_PER_DOMAIN`
* `CONCURRENT_REQUESTS_PER_IP`
* `RANDOMIZE_DOWNLOAD_DELAY`
* `DownloaderAwarePriorityQueue`
* `AutoThrottle`

The reason is that the request skip the downloader, so none of these limits can be applied to DummyRequests. 

Whether this is a useful behaviour or not depends on the case.

ivanprado:
I have analyzed my last comment deeper and it seems that `CONCURRENT_REQUESTS` are actually enforced even in the presence of `DummyRequest`. What is not working is everything that needs to use the downloader slots. So it doesn't work for:
* CONCURRENT_REQUESTS_PER_DOMAIN
* CONCURRENT_REQUESTS_PER_IP
* RANDOMIZE_DOWNLOAD_DELAY
* DownloaderAwarePriorityQueue
* AutoThrottle

victor-torres:
@ivanprado, I think this is the expected behavior. 

DummyRequests are meant to skip downloads, so it makes sense not checking for concurrent requests, delays, or autothrottle settings since we won't be making any download at all. If your parser needs a regular Request, it will be downloaded and I believe all of those settings will be respected.

I suspect the problem here is when you are using third-party libraries to acquire content for a page object. Auto Extract is a good example. 

Before DummyRequests implementation, there was a delay between Auto Extract API requests caused by regular Requests going through the Scrapy's scheduler. The only issue here is that in this case, the crawler was downloading a Response that was going to be discarded. That's a waste of network resources.

I believe that if a third-party library needs to implement some kind of rate limit or delay between its requests, it must do it on its side of the source code or try to leverage Scrapy's mechanisms to do so — for example, making use of Requests to fetch its data.

Let me know if you have other considerations or if you would like to suggest another way to outcome this limitation.

ivanprado:
@victor-torres I think your comment makes a lot of sense. 

My intention was just to state what is the current behaviour. In the presence of `DummyRequest` when a thirty party library is used to fetch content some Scrapy limits still apply (the `CONCURRENT_REQUEST`) but not others. 

Whether if that is a problem or not and what would be the solution is another question. I think your ideas are in the right direction anyway.

victor-torres:
Nice to see that we are on the same page, @ivanprado. 

Just documenting here that after talking today we discussed exposing Scrapy settings to the Injectables. That might at least provide some information to the so-called third-party libraries so that they can enforce the spider's concurrency settings.

Cheers!

PR #11, opened 2020-04-15
Make sure `callback_for` method is ready for production

victor-torres:


tests/test_callback_for.py
```diff
@@ -0,0 +1,64 @@
...
+    assert isinstance(request_dict, dict)
+    assert request_dict['url'] == 'http://example.com/'
+    assert request_dict['callback'] is None
+
+    # Sample request referencing callback on spider
+    request = scrapy.Request('http://example.com/', callback=spider.parse_item)
+    request_dict = request_to_dict(request, spider)
+    assert isinstance(request_dict, dict)
+    assert request_dict['url'] == 'http://example.com/'
+    assert request_dict['callback'] == 'parse_item'
```
victor-torres, line 48: Hey @kmike, this assertion is failing because Scrapy is failing to find the proper parser name. `request_dict['callback']` is returning `'parse'` here because that's the name of the method we're returning from the `callback_for` function.

Currently, Scrapy is executing this code to find the callback name:

```python
def _find_method(obj, func):
    if obj:
        try:
            func_self = func.__self__
        except AttributeError:  # func has no __self__
            pass
        else:
            if func_self is obj:
                name = func.__func__.__name__
                if _is_private_method(name):
                    return _mangle_private_name(obj, func, name)
                return name
    raise ValueError("Function %s is not a method of: %s" % (func, obj))
```

I'd like to refactor this method on Scrapy's repository. That will make it possible for us to use `callback_for` as a spider attribute. 

It's worth noting that the generated callback won't work as an inline callback. It has to be attributed to the spider.

Let me know if you have other thoughts.

victor-torres: Here's the pull request https://github.com/scrapy/scrapy/pull/4500

kmike: The Scrapy change makes sense, thanks @victor-torres.

Inline usage would be nice though. I wonder if we should change the approach - create a mixing for a spider, which provides callback_for method, and overrides `__getattr__`(?). 

Inline usage would be then `scrapy.Request(url, self.callback_for(ProductPage))`.

callback_for can generate some method name (e.g. `__page_object_callback_<ProductPage import path>`. `__getattr__` would ensure that when attribute with a name starting with `__page_object_callback_` is requested, a callback for this page object is created; page object class is probably loaded by scrapy.utils.misc.load_object. Generated callbacks are cached.

This is a hack, but it seems it can work in the meantime. Once we stabilize the whole thing, and organize it better, we can add direct pageobject callbacks support to Scrapy.

kmike: If the hack is too hackish, we can skip it for now, and just document that inline callback_for doesn't work properly with disk queues, consuming the memory.

victor-torres: Thank you for your suggestion. It's definitely a way of archiving this behavior. 

One thing that we might consider here is that we would be overwriting a Scrapy's behavior as it currently does not support inline callbacks with Disk Queues independent from scrapy-po.

Implementing this hack would bring inconsistency to our ecosystem. I believe we should discuss something like this on the Scrapy repo itself. Does it make sense?

tests/test_callback_for.py
```diff
@@ -0,0 +1,64 @@
...
+    """Simple test case to ensure it works as expected."""
+    cb = callback_for(FakePage)
+    assert callable(cb)
+
+    fake_page = FakePage()
+    response = DummyResponse('http://example.com/')
+    assert list(cb(response=response, page=fake_page)) == ['it works!']
+
+
+def test_callback_serialization():
```
kmike, line 32: A nitpick: could you please split this test into separate tests? Probably comments provide good splitting points.

tests/test_callback_for.py
kmike: Fair enough. I don't have a generic proposal on how to fix it for Scrapy though. Any ideas? If any, please open a Scrapy ticket for that :)

scrapy_po/utils.py
```diff
@@ -26,19 +26,23 @@ def __init__(self, url, request=None):
 
 def is_callback_using_response(callback: Callable):
     """Check whether the request's callback method is going to use response."""
-    spec = inspect.getfullargspec(callback)
-    try:
-        arg_name = spec.args[1]  # first index is self, second is response
-    except IndexError:
+    if getattr(callback, '__scrapy_po_callback', False) is True:
```
victor-torres, line 29: This is a new check.

scrapy_po/utils.py
```diff
@@ -26,19 +26,23 @@ def __init__(self, url, request=None):
...
+        # The callback_for function was used to create this callback.
+        return False
+
+    signature = inspect.signature(callback)
```
victor-torres, line 33: I've refactored this code to make use of `inspect.signature` instead of `inspect.getfullargspec` because the first automatically omits the 'self' arguments.

scrapy_po/utils.py
```diff
@@ -129,3 +133,19 @@ def build_instances(plan: andi.Plan, providers):
...
+def callback_for(page_cls):
+    """Helper for defining callbacks for pages with to_item methods."""
+
+    # When the callback is used as an instance method of the spider, it expects
+    # to receive 'self' as its first argument. When used as a simple inline
+    # function, it expects to receive a response as its first argument.
+    #
+    # To avoid a TypeError, we need to receive a list of unnamed arguments and
+    # a dict of named arguments after our injectable.
+    def parse(*args, page: page_cls, **kwargs):
```
victor-torres, line 147: I've had to revert the `*args` and to include a `**kwargs` to make sure this parse method is able to receive optional 'self' instances as arguments. It's well explained in the comment above.

scrapy_po/utils.py
```diff
@@ -129,3 +133,19 @@ def build_instances(plan: andi.Plan, providers):
...
+        yield page.to_item()
+
+    setattr(parse, '__scrapy_po_callback', True)
```
victor-torres, line 150: Since we cannot annotate the first argument with DummyRequest type (the first argument is now optional/dynamic), I'm using this private attribute to signal that this callback is a scrapy-po's callback and therefore it shouldn't be using response (see updated middleware source code).

scrapy_po/utils.py
ejulio, line 29: Remove `is True`? Or can it have some true (non empty list, ..) value?

tests/test_callback_for.py
```diff
@@ -0,0 +1,119 @@
...
+class FakeItemPage(ItemPage):
+
+    def to_item(self):
+        return 'it works!'
+
+
+class FakeItemWebPage(ItemWebPage):
+
+    def to_item(self):
+        return 'it works!'
```
ejulio, line 22: Probably having different texts here can avoid unwanted changes.
Or missing an error because of some wrong type mapping

scrapy_po/utils.py
ejulio: Couldn't it be a callable class and then check it's type?

scrapy_po/utils.py
```diff
@@ -129,3 +133,26 @@ def build_instances(plan: andi.Plan, providers):
...
+            f'{page_cls.__name__} should implement to_item method.')
+
+    # When the callback is used as an instance method of the spider, it expects
+    # to receive 'self' as its first argument. When used as a simple inline
+    # function, it expects to receive a response as its first argument.
+    #
+    # To avoid a TypeError, we need to receive a list of unnamed arguments and
+    # a dict of named arguments after our injectable.
+    def parse(*args, page: page_cls, **kwargs):  # type: ignore
+        yield page.to_item()  # type: ignore
```
ejulio, line 155: `return` instead of `yield`?

tests/test_callback_for.py
```diff
@@ -0,0 +1,119 @@
...
+    fake_page = FakeItemPage()
+    response = DummyResponse('http://example.com/')
+    assert list(cb(response=response, page=fake_page)) == ['it works!']
+
+
+def test_callback_for_instance_method():
+    spider = MySpider()
+    response = DummyResponse('http://example.com/')
+    fake_page = FakeItemPage()
+    assert spider.parse_item(response, page=fake_page)
```
ejulio, line 46: asserting what?

tests/test_callback_for.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_callback_for_inline():
+    callback = callback_for(FakeItemPage)
+    response = DummyResponse('http://example.com/')
+    fake_page = FakeItemPage()
+    assert callback(response, page=fake_page)
```
ejulio, line 53: Same as above

tests/test_callback_for.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_default_callback():
+    """Sample request not specifying callback."""
+    spider = MySpider()
+    request = scrapy.Request('http://example.com/')
+    request_dict = request_to_dict(request, spider)
+    assert isinstance(request_dict, dict)
```
ejulio, line 61: Is this required?
Asserting scrapy code..

victor-torres: It's a public method. Also, that's a nice way to make sure a callback is serializable among its request.

tests/test_callback_for.py
victor-torres: Done

scrapy_po/utils.py
victor-torres: Keeping old behavior... But I've had some experiences when the spider ignored items "returned" while processed items "yielded". Possibly related to async behavior (inline callbacks)?

tests/test_callback_for.py
victor-torres: Done

tests/test_callback_for.py
victor-torres: Enhanced assertions.

scrapy_po/utils.py
victor-torres: It could work like in the following example:

```python
class ItemCallback(Callable):

    pass


def callback_for(page_cls: Type[ItemPage]) -> Callable:
    """Helper for creating callbacks for ItemPage sub-classes."""
    if not issubclass(page_cls, ItemPage):
        raise TypeError(
            f'{page_cls.__name__} should be a sub-class of ItemPage.')

    if getattr(page_cls.to_item, '__isabstractmethod__', False):
        raise NotImplementedError(
            f'{page_cls.__name__} should implement to_item method.')

    class Callback(ItemCallback):

        # When the callback is used as an instance method of the spider,
        # it expects to receive 'self' as its first argument.
        # When used as a simple inline function,
        # it expects to receive a response as its first argument.
        #
        # To avoid a TypeError, we need to receive a list of
        # unnamed arguments and a dict of named arguments after our injectable.
        def __call__(*args, page: page_cls, **kwargs):  # type: ignore
            yield page.to_item()  # type: ignore

    return Callback()
```

And the check would be:

```python
def is_callback_using_response(callback: Callable):
    """Check whether the request's callback method is going to use response."""
    if issubclass(callback, ItemCallback):
        # The callback_for function was used to create this callback.
        return False

    ...
```

The issue is that Scrapy expects a function as a request's callback ([see here](https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request) and a Callable is an object. 

We would need to refactor [this line](https://github.com/scrapy/scrapy/pull/4500/files#diff-89dda711d8912841f3d91bd1816a6ca6R81) removing the `inspect.ismethod` predicate and update the logic to support class-based callables. Not to mention updating Scrapy's docs and making sure everything works as expected. 

I'm not sure which implications this might have.

Any thoughts? (cc @kmike)

victor-torres: I think the current private attribute is a simple and straightforward check that works just fine.

kmike: I was thinking it should be possible to implement parse as a class with `__call__` method and a descriptor (`__get__`), which ensures `self` is not passed (it would need https://github.com/scrapinghub/andi/pull/17 as well).

But given the issues you highlighted with Scrapy, a private attribute sounds like a good approach.

scrapy_po/utils.py
kmike: Could you please move the string to a constant (_CALLBACK_FOR_MARKER or something like that)?

scrapy_po/utils.py
```diff
@@ -129,3 +133,26 @@ def build_instances(plan: andi.Plan, providers):
         else:
             instances[cls] = cls(**kwargs_spec.kwargs(instances))
     raise returnValue(instances)
+
+
+def callback_for(page_cls: Type[ItemPage]) -> Callable:
+    """Helper for creating callbacks for ItemPage sub-classes."""
+    if not issubclass(page_cls, ItemPage):
+        raise TypeError(
+            f'{page_cls.__name__} should be a sub-class of ItemPage.')
```
kmike, line 142: I wonder if this check is needed - it would allow to fail slightly earlier, but previously there was no requirement to inherit from ItemPage, it was more like a helper, to ensure you don't forget to_item.

scrapy_po/utils.py
```diff
@@ -26,19 +26,23 @@ def __init__(self, url, request=None):
...
-    spec = inspect.getfullargspec(callback)
-    try:
-        arg_name = spec.args[1]  # first index is self, second is response
-    except IndexError:
+    if getattr(callback, '__scrapy_po_callback', False) is True:
+        # The callback_for function was used to create this callback.
+        return False
+
+    signature = inspect.signature(callback)
+    first_parameter_key = next(iter(signature.parameters))
```
kmike, line 34: I'm not sure I understand the code well, but could signature.parameters be empty?

kmike: My understanding: no, because Scrapy requires 'response' to be the first argument.

PR #13, opened 2020-04-23
Moving core logic to another library

victor-torres:


victor-torres:
I'll open another pull request from another branch.

PR #14, opened 2020-04-24
Moving core logic to another library

victor-torres:
Superseds #13


dangra:
where is this other core_po library?  what if its name is _pageobjects_ instead of the short `po` ?

victor-torres:
> where is this other core_po library? what if its name is _pageobjects_ instead of the short `po` ?

We've discussed it internally. The derived package will be called `web-poet`.

PR #15, opened 2020-04-24
Update Scrapy version/dependency

victor-torres:
a new version of scrapy was released with a code that we need for finding spider callbacks


kmike:
👍

PR #16, opened 2020-04-27
Rename package from scrapy-po to scrapy-poet

victor-torres:
@kmike, you'll probably need to generate another deploy token since we'll be renaming package on PyPI as well.


kmike:
@victor-torres could you please check examples in README, it looks like some of them are not changed? It could make sense to search "scrapy_po" in the whole repo, just to be sure.

victor-torres:
> @victor-torres could you please check examples in README, it looks like some of them are not changed? It could make sense to search "scrapy_po" in the whole repo, just to be sure.

updated @kmike

PR #21, opened 2020-04-29
Create sphinx documentation

victor-torres:


victor-torres:
Documentation is available here, but it won't render automodule directives. Still figuring out how to solve this issue. https://scrapy-poet.readthedocs.io/en/docs/

.gitignore
```diff
@@ -16,4 +16,7 @@
 .tox/
 .pytest_cache/
 .coverage
-htmlcov/
\ No newline at end of file
+_build/
```
ivanprado, line 19: Given that `_build`, `_static` and `_templates` are all of them related with documentation, maybe it can be grouped together within the same folder (e.g. `_doc`) to avoid polluting too much the root folder.

ivanprado:
(Sorry, I pop here because I thought I was in the list of reviewers, but now I'm unsure I was)

README.rst
```diff
@@ -61,11 +61,13 @@ extraction code from a spider:
...
     class BooksSpider(scrapy.Spider):
+
         name = 'books'
         start_urls = ['http://books.toscrape.com/']
 
         def parse(self, response):
-            for url in response.css('.image_container a::attr(href)').getall():
+            query = '.image_container a::attr(href)'
+            for url in response.css(query).getall():
                 yield response.follow(url, self.parse_book)
```
kmike, line 71: It can be shortened further to

`yield from response.follow_all(css='.image_container a', callback=self.parse_book)`

or 

```py
book_links = response.css(".image_container a")
yield from response.follow_all(book_links, self.parse_book)
```

.gitignore
victor-torres: I guess apart from `_build`, the other directories are created automatically by sphinx.

victor-torres:
> (Sorry, I pop in here because I thought I was in the list of reviewers, but now I'm unsure I was)

Don't worry, @ivanprado. Your reviews are always welcome. 



> Thank you @victor-torres! I'm learning with this PR how this sphinx thing work 👏 . I left a comment and will have a deeper look at the documentation content later.
> 
> I tried to access the API reference here https://scrapy-poet.readthedocs.io/en/docs/ but I didn't manage to found it. Where is it present?

That's the problem I mentioned during our meeting today. It works when docs are built locally. @rennerocha will help us to make it work. It seems like he has already done  it on https://github.com/scrapinghub/spidermon.

rennerocha:
@victor-torres ,

I took a look more closely and as you said it is working locally.

The problem probably is within ReadTheDocs settings. You need to set a flag in `Advanced Settings` to install your project inside a virtualenv. Doing that, when `sphinx` is executed (which will be inside this virtualenv) it will be able to find `scrapy-poet` module and generate the docs from the docstrings.

.gitignore
ivanprado: Changing the paths here
https://github.com/scrapinghub/scrapy-poet/pull/21/files#diff-85987f48f1258d9ee486e3191495582dR50 
and here:
https://github.com/scrapinghub/scrapy-poet/pull/21/files#diff-85987f48f1258d9ee486e3191495582dR93
wouldn't work?

scrapy_poet/middleware.py
```diff
@@ -35,8 +37,20 @@ def process_request(self, request: Request, spider):
...
+        """This method instantiates all ``Injectable`` sub-classes declared as
+        request callback arguments and any other parameter with a provider for
+        its type. Otherwise, this middleware doesn't populate
+        ``request.cb_kwargs`` for this argument.
+
+        .. warning::
+
+            Currently, we're able to inject any type that inherits from
+            Injectable, but the only external dependency that we're able to
+            build and inject is the ``scrapy.Response``.
```
ivanprado, line 51: `ResponseData` can also be injected even if it is not an Injectable. Also, types with a defined provider would be injected.

scrapy_poet/page_input_providers.py
```diff
@@ -1,41 +1,66 @@
-# -*- coding: utf-8 -*-
+"""The Injection Middleware needs a standard way to build dependencies for
```
ivanprado, line 1: Maybe it can be mentioned here that one use case to implement a provider is to obtain data from external services or APIs like Splash, AutoExtract, etc.

victor-torres:
> @victor-torres ,
> 
> I took a look more closely and as you said it is working locally.
> 
> The problem probably is within ReadTheDocs settings. You need to set a flag in `Advanced Settings` to install your project inside a virtualenv. Doing that, when `sphinx` is executed (which will be inside this virtualenv) it will be able to find `scrapy-poet` module and generate the docs from the docstrings.
> 
> install

This checkbox was already marked. I'll get back to this issue this morning and will keep you updated.

.gitignore
victor-torres: It would work, but I'm not sure it's a regular practice. Scrapy, for example, keeps the original value for this config: https://github.com/scrapy/scrapy/blob/master/docs/conf.py

victor-torres: Although it commits `_static` and `_templates` directories 🤔

victor-torres:
@rennerocha I was able to fix the [readthedocs.io](https://readthedocs.io) build:

- missing dependency (scrapy)
- setuptools' bug (https://github.com/pypa/setuptools/issues/498)

Link is here (make sure to empty your browser cache before reloading): https://scrapy-poet.readthedocs.io/en/docs/

README.rst
victor-torres: Done

scrapy_poet/middleware.py
victor-torres: Fixed. I've just updated this docstring to clarify what should be said. 

```rst
We should be able to inject any type into classes that inherit from
``web_poet.pages.Injectable``, but currently, we're only able to
build and inject ``scrapy.Response`` instances.
```

scrapy_poet/page_input_providers.py
victor-torres: Done.

setup.py
```diff
@@ -12,7 +12,18 @@
...
+    install_requires=[
+        # We need to install pyasn1 before trying to install service_identity,
+        # one of Scrapy's dependencies. This is a bug related to setuptools.
+        #
+        # https://github.com/pypa/setuptools/issues/498
+        'pyasn1',
+        'andi>=0.3',
+        'attrs',
+        'parsel',
+        'scrapy>=2.1.0',
```
kmike, line 24: Hm, I'm not sure about this change. On one hand, it seems some framework-related libraries do that (e.g. some django-related packages do require django in setup.py). On the other hand, not everyone is doing this, and adding a framework as an install_requirement for a plugin can lead to problems - e.g. a framework can be occasionally upgraded when user tries to upgrade a package. 

pip used to upgrade all dependencies always, when upgrade of a package is requested; this seems to be changed (now it upgrades only dependencies which don't have a compatible version installed), but it still may lead to problems. Why would user want to have Scrapy upgraded (or installed) when `pip install scrapy-poet` or `pip install -U scrapy-poet` is executed? For me failing in case of incompatible Scrapy version looks like a better behavior than silently upgrading Scrapy.

An issue with docs building and CI can be solved in different ways, e.g. having docs/requirements.txt file which includes Scrapy.

victor-torres: Thank you for your comments. That makes sense. I've reverted the setup.py file and included a requirements.txt to the docs directory. Doc is  still building fine after that :)

kmike:
Hey @victor-torres! Congratulations on making everything built & configured! 
I think we can merge it as-is, unless you want to work on the docs themselves in this PR. What do you prefer?

I think docs here can be cleaned up in the following way (this is without writing any additional docs, just rearranging the existing content):

* Simplify README.rst: add a link to docs instead of most content. README could contain basic information about what this package is doing, it license and links to various resources (docs, issue tracker); other information can be in docs, to avoid duplication.
* Move auto-generated docs to a separate "API reference" file
* Keep basic information in the index docs page: description of what is this package for, install instructions, contributing information; maybe - a basic usage example (for illustration purposes, not as a tutorial / real documentation). Add toctree to this page, with links to other sections (API reference at the moment). The goal of this web page is to give a basic overview of what's this all about, and help to navigate the rest of the documentation.
* remove "TODO" items and other incomplete stuff moved from README

victor-torres:
Thanks for your review, @kmike. I'm opening different pull requests to keep changes separated.

.gitignore
ivanprado: I think I missed that `_build`, `_static` and `_templates` are going to be created inside the `docs` folder, right?  If this is the case then my proposal makes no sense. 

Besides, I don't know if these folders should committed. I see that scrappy does not include `_static` and `_templates`  in `.gitignore` but it does for `docs/build` instead (the equivalent to `_build` in your PR).

ivanprado: By the way, I see this is already merged! Forget about that :+1:

PR #22, opened 2020-04-30
Updating docstrings

victor-torres:


README.rst
```diff
@@ -1,4 +1,3 @@
-===========
```
kmike, line 1: This line gave a different header level; after this change scrapy-poet is on a same level as e.g. Usage.

If that's a problem for docs, I think a little of duplication is fine (i.e. pull in changelog, but not readme).

scrapy_poet/utils.py
```diff
@@ -22,8 +22,50 @@ def get_callback(request, spider):
...
+                self.response = response
+
+            def __call__(self):
+                return ResponseData(
+                    url=self.response.url,
+                    html=self.get_data()
+                )
+
+            def get_data(self):
+                # make an api call
```
kmike, line 61: to make an API call you would probably want an async def function

README.rst
victor-torres: reverted

scrapy_poet/utils.py
victor-torres: I've updated this example to use async/await.

kmike:
Thanks @victor-torres!

PR #24, opened 2020-05-04
Organizing docs

victor-torres:
Implements suggestions from @kmike here: https://github.com/scrapinghub/scrapy-poet/pull/21#issuecomment-623614062


victor-torres:
@kmike I'm planning to create another pull request moving examples from that directory into this doc as soon as this one gets merged, alright?

docs/index.rst
```diff
@@ -3,126 +3,12 @@
...
-=====
-
-.. autoclass:: scrapy_poet.utils.DummyResponse
-   :members:
-   :no-special-members:
-
-.. automethod:: scrapy_poet.utils.callback_for
-
-.. include:: ../CHANGES.rst
+.. include:: overview.rst
```
kmike, line 6: Why is this include needed?

docs/index.rst
```diff
@@ -3,126 +3,12 @@
...
 
 .. toctree::
    :maxdepth: 2
    :hidden:
 
-   index
-
-.. _tox: https://tox.readthedocs.io
-.. _`web-poet`: https://github.com/scrapinghub/web-poet
+   overview
```
kmike, line 12: I'd remove a separate overview page, because the content is duplicated in the docs: it is included in index, and then also appears in toctree, as a separate page. What do you think about keeping this content in index.rst, and removing overview.rst?

victor-torres:
Hey @kmike, I've just updated the docs. A live version is available here: https://scrapy-poet.readthedocs.io/en/organize-docs/. Could you take another look? Thanks.

docs/intro/overview.rst
```diff
@@ -0,0 +1,18 @@
+===============
+Getting started
+===============
+
+The goal of this project is to make reusable Page Objects that separates
+extraction logic from crawling. They could be easily tested and distributed
+across different projects. Also, they could make use of different backends,
+for example, acquiring data from Splash and Auto Extract API.
```
kmike, line 8: Could you please add links for both Splash and AutoExtract?

docs/intro/tutorial.rst
```diff
@@ -0,0 +1,213 @@
+.. _`intro-tutorial`:
+
+========
+Tutorial
+========
+
+In this tutorial, we’ll assume that scrapy-poet is already installed on your
+system. If that’s not the case, see :ref:`intro-install`.
+
+We are going to scrape `quotes.toscrape.com <http://quotes.toscrape.com/>`_,
```
kmike, line 10: book.toscrape is used in the examples

docs/intro/tutorial.rst
```diff
@@ -0,0 +1,213 @@
...
+Let's create our first Page Object abstraction by moving extraction logic
+outside our spider file.
+
+.. code-block:: python
+
+    from web_poet.pages import ItemWebPage
+
+
+    class BookPage(ItemWebPage):
+        """Extracts data from a book page"""
```
kmike, line 74:
```suggestion
        """Individual book page on books.toscrape.com website, e.g.
        http://books.toscrape.com/catalogue/a-light-in-the-attic_1000/index.html
        """
```

docs/intro/tutorial.rst
```diff
@@ -0,0 +1,213 @@
...
+
+        def to_item(self):
+            """Converts page into an item"""
```
kmike, line 77:
```suggestion
            """Convert page into an item"""
```
a nitpick: as per [pep-257](https://www.python.org/dev/peps/pep-0257/), 

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

docs/intro/tutorial.rst
```diff
@@ -0,0 +1,213 @@
...
+            links = response.css('.image_container a')
+            yield from response.follow_all(links, self.parse_book)
+
+        def parse_book(self, response, book_page: BookPage):
+            """Extract data from book pages"""
+            yield book_page.to_item()
+
+You might have noticed that our parser method is quite simples and it's just
+returning the result of the ``to_item`` method call. We could make use of the
+``callback_for`` helper to reduce source code here.
```
kmike, line 142:
```suggestion
``callback_for`` helper to reduce boilerplate.
```

docs/intro/tutorial.rst
```diff
@@ -0,0 +1,213 @@
...
+        name = 'books'
+        start_urls = ['http://books.toscrape.com/']
+        parse_book = callback_for(BookPage)
+
+        def parse(self, response):
+            """Discovers book links and follows them"""
+            links = response.css('.image_container a')
+            yield from response.follow_all(links, self.parse_book)
+
+This helper could be used as an inline callback, but it would not work with
```
kmike, line 161: I wonder if this can be reworded; "inline callback" can reming people of scrapy-inline-callbacks or twisted's `@inlineCallbacks`, but this is not what is meant here.

docs/intro/overview.rst
```diff
@@ -0,0 +1,18 @@
+===============
+Getting started
+===============
```
kmike, line 3: I think this content can be put directly in index.rst, before the links. "Please, see also our :ref:`intro-install` and our :ref:`intro-tutorial` for a quick start." can just refer to the links below. It seems index.rst won't become much larger, but it'd give some information about what's the package about, without making user to visit any additional links.

docs/advanced.rst
```diff
@@ -0,0 +1,208 @@
...
+
+DummyRequests are meant to skip downloads, so it makes sense not checking for
+concurrent requests, delays, or auto throttle settings since we won't be making
+any download at all.
+
+By default, if your parser and its page inputs need a regular Request, it will
+be downloaded and all of those Scrapy settings related to it are going to be
+respected, for example:
+
+- ``CONCURRENT_REQUESTS``
```
kmike, line 191: If I'm not mistaken, this option is still respected with scrapy-poet (/cc @ivanprado).

kmike:
❤️ the new docs, great job @victor-torres!

docs/intro/tutorial.rst
kmike: Maybe just show an example: "you can also write `Request(url, callback_for(BookPage))` directly, without creating an attribute, but currently this won't work with Scrapy disk queues", something like that?

docs/advanced.rst
ivanprado: Yes, it is still respected even when dummy request is used.

docs/intro/overview.rst
victor-torres: Done.

docs/intro/tutorial.rst
victor-torres: Thanks, it's fixed now.

docs/intro/tutorial.rst
victor-torres: I've just rewritten this text and formatted it as a note.

docs/intro/overview.rst
victor-torres: Done.

- moved overview content to index
- removed overview page
- removed links from the index page since we already have toctrees

docs/advanced.rst
victor-torres: Yes, it's respected. Docs says:

- by default, those settings are respected
- be aware when using third-party libraries for requests because they're probably ignoring these settings

Just double-checking, is any action needed here, @kmike? Any suggestions on how to clarify this passage?

kmike: If you make an HTTP request in a provider using some third-party async library (aiohttp, treq, etc.), CONCURRENT_REQUESTS option will be respected, but not the other settings. 

To have other settings respected, in addition to CONCURRENT_REQUESTS, you'd need to use crawler.engine.download or something like that; alternatively, implement limits in a library.

We'll also make it possible to use DownloaderAwarePriorityQueue in such cases, by refactoring Scrapy (this is a separate task).

victor-torres: I've just updated the docs. Would you mind having another look? If you're cool with the latest modifications, we could go ahead and merge this PR.

kmike:
Thanks @victor-torres! I might have a couple of suggestions, but it'd be easier for me to just submit a follow-up PR with proposed changes.

PR #26, opened 2020-05-11
Documentation follow-up

kmike:
This PR has small changes to the existing docs;  see commit messages for more detailed list of changes.

I think the version we have here is not final, it seems there are still things to improve:

* tutorial doesn't provide enough motivation - it shows what to do, but fails to explain why
* tutorial doesn't go far enough (e.g. no listing page objects) -  but it only makes sense to have these sections after fixing the "motivation" bit
* advanced docs may need to be split up for docs needed to provider writers and docs needed for the users. Currently it is written for provider writers, but it seems e.g. DummyResponse can be  explained in a simpler way for people who just need to use it.

I haven't touched those, let's keep if out of scope for this PR.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -44,7 +44,7 @@ current scrapy-poet behavior.
 .. warning::
 
     Currently, scrapy-poet is only able to inject ``Response`` and
-    ``DummyResponse`` instances as *provider* dependencies.
+    :class:`~.DummyResponse` instances as *provider* dependencies.
 
 Ignoring requests
 =================
@@ -54,7 +54,7 @@ using a third-party API such as Auto Extract or querying a database.
 
 In cases like that, it makes no sense to send the request to Scrapy's downloader
 as it will only waste network resources. But there's an alternative to avoid
-making such requests, you could use ``DummyRequests`` type to annotate
+making such requests, you could use :class:`~.DummyResponse` type to annotate
 your response arguments.
 
 That could be done in the spider's parser method:
@@ -70,11 +70,11 @@ to not download scrapy Response as usual.
 
 This type annotation is already applied when you use the :func:`~.callback_for`
 helper: the callback which is created by ``callback_for`` doesn't use Response,
-it just calls page object's to_item method.
+it just calls page object's ``to_item`` method.
 
 If neither spider callback nor any of the input providers are using
-``Response``, InjectionMiddleware skips the download, returning a
-``DummyResponse`` instead. For example:
+``Response``, :class:`~.InjectionMiddleware` skips the download, returning a
+:class:`~.DummyResponse` instead. For example:
 
 .. code-block:: python
 
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -7,9 +7,10 @@ Advanced Usage
 Creating providers
 ==================
 
-Providers are responsible for building dependencies needed by Injectors. A very
-good example would be the ``ResponseDataProvider``, which builds and provides a
-``ResponseData`` instance for Injectors that need it, like the ``ItemWebPage``.
+Providers are responsible for building dependencies needed by Injectable
+objects. A good example would be the ``ResponseDataProvider``,
+which builds and provides a ``ResponseData`` instance for Injectables
+that need it, like the ``ItemWebPage``.
 
 .. code-block:: python
 
@@ -42,9 +43,8 @@ current scrapy-poet behavior.
 
 .. warning::
 
-    Currently, scrapy-poet is only able to inject ``Response`` instances as
-    provider dependencies. We should be able to overcome this limitation in the
-    future.
+    Currently, scrapy-poet is only able to inject ``Response`` and
+    ``DummyResponse`` instances as *provider* dependencies.
 
 Ignoring requests
 =================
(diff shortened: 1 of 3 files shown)


docs/intro/tutorial.rst
```diff
@@ -13,8 +13,8 @@ a website that lists books from famous authors.
 This tutorial will walk you through these tasks:
 
 1. Writing a :ref:`spider <scrapy:topics-spiders>` to crawl a site and extract data
-2. Configuring Scrapy project to include Injection Middleware
-3. Separating extraction logic from spider
+3. Separating extraction logic from the spider
```
victor-torres, line 16: 3 -> 2

docs/intro/tutorial.rst
```diff
@@ -83,44 +73,70 @@ outside our spider file.
...
-The parser method now receives a type annotated argument called ``book_page``.
-Our Injection Middleware will detect it and provide the required dependencies.
+``parse_book`` method now has a type annotated argument
+called ``book_page``. scrapy-poet detects this and makes sure
+a BookPage instance is created and passed to the callback.
 
-The spider should be looking like this:
+The full spider code would be looking like this:
 
 .. code-block:: python
```
victor-torres, line 142: Are we importing scrapy here as well?

docs/intro/tutorial.rst
```diff
@@ -160,38 +178,34 @@ returning the result of the ``to_item`` method call. We could make use of the
...
 ============
 
-At the end of our job, our spider should look like this:
+At the end of our job, the spider should look like this:
 
 .. code-block:: python
 
     import scrapy
-
-    from scrapy_poet import callback_for
```
victor-torres, line 178: I think imports are supposed to follow alphabetical order.

docs/advanced.rst
```diff
@@ -64,13 +64,16 @@ That could be done in the spider's parser method:
...
-``callback_for`` helper.
+Spider method that has its first argument annotated as :class:`~.DummyResponse`
+is signaling that it is not going to use the response, so it should be safe
+to not download scrapy Response as usual.
 
-If neither spider callback or any of the input providers are going to make use
-of a ``Response``, our Injection Middleware skips download returning a
+This type annotation is already applied when you use the :func:`~.callback_for`
+helper: the callback which is created by ``callback_for`` doesn't use Response,
+it just calls page object's to_item method.
```
victor-torres, line 73: Should we format this way?

```rst
``to_item``
```

docs/advanced.rst
```diff
@@ -64,13 +64,16 @@ That could be done in the spider's parser method:
...
+
+If neither spider callback nor any of the input providers are using
+``Response``, InjectionMiddleware skips the download, returning a
```
victor-torres, line 76: Same here: should we format this way?

```rst
``InjectionMiddleware``
```

Or even:

```rst
:class:`~.InjectionMiddleware`
```

docs/intro/tutorial.rst
rennerocha: It is better to use `1.` for all items in an enumerated list. It will be auto-generated with the right numbers and avoids problems like this when you include new items to the list.

docs/intro/tutorial.rst
kmike: Why? :) 

Let me explain the logic for the current imports. The rule in pep8 is the following:

> Imports should be grouped in the following order:
>
>    Standard library imports.
>    Related third party imports.
>    Local application/library specific imports.
>
> You should put a blank line between each group of imports.

Here, technically, all imports are "related third party imports", so pep8 doesn't specify any order.

I ordered them by "relatedness" - scrapy is almost considered stdlib here, and scrapy-poet is almost a local application/library specific import, as we're explaining how scrapy-poet works in the tutorial. I can admit this order is quite arbitrary; is it worse than other orders? I've seen people are ordering imports alphabetically, by length, etc.

docs/intro/tutorial.rst
kmike: Thanks, fixed!

docs/intro/tutorial.rst
kmike: added an import, thanks! Initially I haven't added it because this is not a complete example (it needs a BookPage to work).

docs/advanced.rst
kmike: done

docs/advanced.rst
kmike: Added more links for InjectionMiddleware and DummyResponse, thanks!

docs/intro/tutorial.rst
victor-torres: Sorry, I didn't mean there was a PEP or something.

I'm used to seeing alphabetical order across some Python projects and I like this approach because it's a concrete convention that's easy to be followed by any developer working on the project. Whereas "relatedness" is somewhat subjective to a certain point.

Also, I would hate to nitpick to this point. This was more like a suggestion. Up to you :)

kmike: Regarding alphabetical order - nothing against it, though for the guidelines which can be automated tooling like https://github.com/timothycrosley/isort may work better; you don't need to spend human brain cycles on something a machine can check & fix :)

kmike:
Thanks for the helpful comments @victor-torres and @rennerocha!

PR #30, opened 2020-05-21
Expose Scrapy Settings to Providers

victor-torres:
It would be nice to allow providers to access Scrapy Settings because they may need some configuration. For example, they may need to know concurrency limits.

Previously to this pull request, the Injection Middleware was able to inject Response instances only. It's now specifying a dependency map to the plan builder. Available types are:
- `scrapy.http.Request`
- `scrapy.http.Response`
- `scrapy.settings.Settings`
- ~`scrapy_poet.utils.DummyResponse`~

~A current limitation is that we need to specify the exact types in the dependency map. For example, it seems like [andi](https://github.com/scrapinghub/andi) is not able to understand that a `DummyResponse` might be replaced with a parent class like `Response` when the more specific is not present in the map. This could be the source of some problems if a provider makes use of `TextResponses`, for example.~

If your provider needs to use a subclass of Response, for example, TextResponse, make sure to annotate your response argument with the parent class.

This behavior is related to covariance. In simple words, this means a subclass could replace a parent class but a parent class could not substitute its subclasses. Example:

	class Animal:
		pass

	class Cat(Animal):
		def meow(self):
			pass

If our provider needs an Animal instance, we could provide either an Animal or a Cat (a more specific Animal subtype). On the other hand, if our provider needs a Cat instance (probably because meow is a requirement), it wouldn't be okay to pass it an Animal instance.

Our Injection Middleware makes no distinction between Responses and TextResponses, so you need to validate the response type by yourself. Example:

    @provides(MyCustomResponseData)
    class MyCustomResponseDataProvider(PageObjectInputProvider):

        def __init__(self, response: Response):
            assert issubclass(response, TextResponse)
            self.response = response


scrapy_poet/middleware.py
```diff
@@ -44,15 +49,23 @@ def process_response(self, request: Request, response: Response,
...
+        - :class:`~scrapy.settings.Settings`
         """
         # Find out the dependencies
         callback = utils.get_callback(request, spider)
-        plan, provider_instances = utils.build_plan(callback, response)
+        dependencies = {
+            Request: request,
+            Response: response,
+            Settings: spider.settings,
+            utils.DummyResponse: self.get_dummy_response(request),
```
victor-torres, line 66: We probably don't need DummyResponses in providers anymore because we can now make use of Request instances.

scrapy_poet/middleware.py
```diff
@@ -44,15 +49,23 @@ def process_response(self, request: Request, response: Response,
...
+        - :class:`~.DummyResponse`
+        - :class:`~scrapy.http.Request`
+        - :class:`~scrapy.http.Response`
+        - :class:`~scrapy.settings.Settings`
         """
         # Find out the dependencies
         callback = utils.get_callback(request, spider)
-        plan, provider_instances = utils.build_plan(callback, response)
+        dependencies = {
+            Request: request,
```
victor-torres, line 63: Try to add something in docs about covariance and why we should always specify Response instead of any of its sub-classes like TextResponse in providers' arguments.

tests/test_scrapy_settings.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@attr.s(auto_attribs=True)
+class MySettings:
+
+    robotstxt_obey: bool
+
+
+@provides(MySettings)
+class MySettingsProvider(PageObjectInputProvider):
```
ivanprado, line 36: It would be nice to have also a test with a provider which uses an injectable and inside the settings. For example:

```py

@attr.s(auto_attribs=True)
class SettingsWrapper(Injectable):
    settings: Settings

@provides(MySettings)
class MySettingsProvider(PageObjectInputProvider):
 
    def __init__(self, settings_wrapper: SettingsWrapper):
        self.settings = settings_wrapper.settings
```

Just to check that full recursive injection is also working.

scrapy_poet/middleware.py
```diff
@@ -44,15 +49,23 @@ def process_response(self, request: Request, response: Response,
...
+            Response: response,
+            Settings: spider.settings,
```
ivanprado, line 60: I think we can add here what @kmike told was missing: the crawler, the stats, etc.

scrapy_poet/middleware.py
ivanprado: +1 to remove `DummyResponse`

scrapy_poet/utils.py
```diff
@@ -121,27 +121,31 @@ def is_response_going_to_be_used(request, spider):
...
 
     return False
 
 
-def get_providers(plan: andi.Plan):
-    for obj, _ in plan:
-        provider = providers.get(obj)  # type: ignore
-        if not provider:
-            continue
+def get_providers(callback):
```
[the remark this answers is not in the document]
victor-torres: That's a good question! 

As you've noted, the previous approach used `build_plan`.  We refactored this function and it now depends on an external dependency list.  We don't have this list when we call `is_response_going_to_be_used`, which is the one using `get_providers`. 

The Injection Middleware calls this function before the download happens. This way we are able to skip it if the response is not going to be used.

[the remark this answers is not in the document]
ivanprado: I see. Still, I think `andi.plan` can be used, something like:

```py
def get_providers(callback):
  def providers_also_injectable(cls):
     return is_injectable(cls) or cls in providers.keys()

  plan = andi.plan(callback, is_injectable=providers_also_injectable, externally_provided={Request, Response, Settings})
  return list(dict(plan).keys() & providers.keys())
```

One reason to use this approach is that all the logic with `Union` is properly managed there. For example, what to do when you have two providers in a union (e.g. `data: Union[AutoExtractResponseData, ResponseData]`).  Probably the current code of `get_providers` can be change to solve properly this case, but better to keep all this kind of code in andi internals.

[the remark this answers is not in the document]
ivanprado: Well, even more. This piece of code https://github.com/victor-torres/scrapy-po/blob/provider-settings/scrapy_poet/utils.py#L124-L126

Can be replaced by:
```py
  def providers_also_injectable(cls):
     return is_injectable(cls) or cls in providers.keys()

  plan = andi.plan(callback, is_injectable=providers_also_injectable, externally_provided={Request, Response, Settings})
  if Request in dict(plan):
      return True
```

And then `get_providers` and `is_provider_using_response` could be removed. 

Well, and probably encapsulating this code in a function, of course.

tests/test_scrapy_settings.py
victor-torres: I've just refactored tests for better readability and to make sure recursive injection is working.

docs/advanced.rst
```diff
@@ -41,10 +41,56 @@ that need it, like the ``ItemWebPage``.
...
+- :class:`~scrapy.settings.Settings`
+- :class:`~scrapy.statscollectors.StatsCollector`
+
 .. warning::
 
-    Currently, scrapy-poet is only able to inject ``Response`` and
-    :class:`~.DummyResponse` instances as *provider* dependencies.
+    If your provider needs to use a subclass of Response,
+    for example, TextResponse,
+    make sure to annotate your response argument with the parent class type.
```
kmike, line 57: nitpick: it should be annotated with Response, not with a parent class. E.g. for HtmlResponse parent class is TextResponse.

scrapy_poet/utils.py
```diff
@@ -39,21 +39,21 @@ def parse(self, response: DummyResponse):
 
     If your ``PageObjectInputProvider`` doesn't need a request, you simply
     don't need to list it as a dependency. But if you need, for example, the
-    original request's URL, you can use ``DummyResponse`` instead of
+    original request's URL, you can use ``Request`` instead of
```
kmike, line 42: I'm not sure a section about Request for PageObjectInputProvider makes sense in DummyResponse docstring; DummyResponse has nothing to do with providers now, it is for spider writers who need to skip downloads.

scrapy_poet/utils.py
```diff
@@ -100,14 +100,14 @@ def is_provider_using_response(provider):
     for argument, possible_types in andi.inspect(provider.__init__).items():
         for cls in possible_types:
             if not issubclass(cls, Response):
-                # Type annotation is not a sub-class of Response.
+                # Type annotation is not a subclass of Response.
                 continue
 
             if issubclass(cls, DummyResponse):
```
kmike, line 106: Do we still need to support DummyResponse here?

kmike: It seems that if we drop DummyResponse support, is_provider_using_response function can be simplified

tests/test_scrapy_settings.py
```diff
@@ -0,0 +1,121 @@
...
+
+        return name
+
+    def to_item(self):
+        return {
+            "name": self.name,
+        }
+
+
+class ProductSettings(Injectable):
```
kmike, line 60: nice example

scrapy_poet/utils.py
```diff
@@ -150,14 +154,16 @@ def build_plan(callback, response
...
-    # find out what resources are available
+def build_providers(instances) -> Dict[Type, PageObjectInputProvider]:
     result = {}
     for cls, provider in providers.items():
-        if andi.inspect(provider.__init__):
-            result[cls] = provider(response)  # type: ignore
-        else:
-            result[cls] = provider()
+        kwargs = andi.plan(
+            provider.__init__,
```
kmike, line 161: Why is `.__init__` needed, can't just `provider` be passed here?

kmike:
A nice approach @victor-torres!

scrapy_poet/utils.py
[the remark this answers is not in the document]
victor-torres: I was not able to make it work using your suggestion. 

```python
(Pdb) andi.plan(callback, is_injectable=lambda cls: cls in providers.keys() or is_injectable(cls), externally_provided={Response})
*** andi.errors.NonProvidableError: Not possible to generate a plan for argument 'page' in '<bound method callback_for.<locals>.parse of <MySpider 'my_spider' at 0x10ae66128>>'. Causes:
    0. Any of [<class 'str'>] types are required for argument 'url' in '<class 'web_poet.page_inputs.ResponseData'>.__init__()' but none of them is injectable or externally providable
    1. Any of [<class 'str'>] types are required for argument 'html' in '<class 'web_poet.page_inputs.ResponseData'>.__init__()' but none of them is injectable or externally providable
```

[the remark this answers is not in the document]
victor-torres: In any case, andi's inspect function takes care of Union argument types.
See here: https://github.com/scrapinghub/andi/blob/master/andi/andi.py#L46

victor-torres:
Thank you for your thoughts, @ivanprado and @kmike. 
The pull request was updated. So were the open comments :)

scrapy_poet/utils.py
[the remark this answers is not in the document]
ivanprado: > I was not able to make it work using your suggestion.

Ahh, Ok. I see the problem. Please add `ResponseData` to `externally_provided` list and try again. This should fix this problem. 

> In any case, andi's inspect function takes care of Union argument types.

Yes, but there is some logic about what to do with Unions which is in `andi.plan` here: https://github.com/scrapinghub/andi/blob/2dba95e05dd603aed0d203602e688883fc9ce1f2/andi/andi.py#L299-L319. 

For example, imagine you have two providers with a `Union` - `data: Union[AutoExtractResponseData, ResponseData]`. 
* `andi.plan` will return only one of them in the plan:
  - `AutoExtractResponseData` if it is available (if it is in providers list, or in other words, if it is in `externally_provided`)
  - `ResponseData` if it isn't
* `get_providers` will always return both with the current implementation. This will force  `Response` to be considered as required even if it is not (`AutoExtractResponseData` does not require `Response`)

The later can be fixed in `get_providers` but I think is better if we keep this kind of logic internally into `andi.plan`

ivanprado:
I'm thinking about the removal of `DummyResponse` from providers... Maybe we cannot remove it. For example, let's say you have a provider that does not require a `Response` from Scrapy because it is accessing an external API. You still need there the `DummyResponse` to get the Url to fetch. 

```py
    @provides(ApiResponse)
    class ApiResponseProvider(PageObjectInputProvider):
        def __init__(self, response: DummyResponse):
            self.response = response

        async def __call__(self):
            data = await api_request(self.response.url)
            return ApiResponse(data=data)
```

So `DummyResponse` function in providers is:
* stating that you don't need the Scrapy full response
* being able to get the Url

An alternative is to allow only to annotate `Response` in the provider (no `DummyResponse`) and ignore providers dependencies. In some cases, it will be injected DummyResponse and in others, it will be a regular `Response` depending on the callback dependencies. The flaw  is that then requesting `RequestData` will fail in cases like the following:
```py
def parse(response: DummyResponse, data: ResponseData):
   ...
```
It will fail when trying to fulfil the `ResponseData` requirement.

So I think we should allow `DummyResponse` in providers. But always a `DummyResponse` should be received if this is what was required, even if finally a full `Request` was fetched. Otherwise, the contract would be broken. So we will have to fill `DummyResponses` from real `Responses` in some cases.

scrapy_poet/middleware.py
ivanprado: Well, I changed my opinion. See https://github.com/scrapinghub/scrapy-poet/pull/30#issuecomment-632580102

ivanprado:
I'm thinking about that even further. See the nice @victor-torres `Settings` example 
 https://github.com/victor-torres/scrapy-po/blob/provider-settings/tests/test_scrapy_settings.py#L31-L57

We are introducing a lot of boilerplate: we have to write a provider for `Settings` which receive `Settings` in the constructor, to finally being able to inject `Settings` anywhere in page objects. So I ask myself, why not directly allow injecting `Settings` directly, without any provider at all?

In other words, we have a shelf of things that already exist when the callback is going to run. This is the self: https://github.com/victor-torres/scrapy-po/blob/provider-settings/scrapy_poet/middleware.py#L61-L67

If we want them to be injectable, we have to write a provider for each of one of them. Seems a lot of boilerplate. I would just have them as "externally providable" and that's all, like some "non-injectable" objects that can be used as a dependency in any case because they are there. I would include a `DummyResponse` to this shelf as well. 

What do you think?

scrapy_poet/utils.py
```diff
@@ -121,27 +84,31 @@ def is_response_going_to_be_used(request, spider):
...
-    for obj, _ in plan:
-        provider = providers.get(obj)  # type: ignore
-        if not provider:
-            continue
+def get_providers(callback):
+    for argument, possible_types in andi.inspect(callback).items():
+        for cls in possible_types:
+            if is_injectable(cls):
+                yield from get_providers(cls)
+                continue
```
ejulio, line 99: Semantics..
`get_providers` and `providers.get`
Not sure if the distinction is clear.
Maybe import `providers` as `available_providers` and `get_providers` to `search_for_providers`, as it is not `get`, but most like a search in a tree

scrapy_poet/utils.py
```diff
@@ -121,27 +84,31 @@ def is_response_going_to_be_used(request, spider):
...
-        yield provider
+            provider = providers.get(cls)
+            if not provider:
+                continue
+
+            yield provider
 
 
-def build_plan(callback, response
+def build_plan(callback, instances
```
ejulio, line 108: :lipstick:  This is only styling preference, but better a single long line than a break..

tests/test_scrapy_settings.py
```diff
@@ -0,0 +1,121 @@
...
+            <a href="/food/sweets">Sweets</a>
+        </div>
+        <h1 class="name">Chocolate</h1>
+        <p>Price: <span class="price">22€</span></p>
+        <p class="description">The best chocolate ever</p>
+    </html>
+    """
+
+
+@provides(Settings)
```
ejulio, line 31: Is this one required?
Shouldn't it be by default now?

scrapy_poet/utils.py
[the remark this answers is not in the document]
victor-torres: > Please add ResponseData to externally_provided list and try again. This should fix this problem.

Yes, it should. But how are we supposed to discover all those unmet dependencies during runtime?

> For example, imagine you have two providers with a Union - data: Union[AutoExtractResponseData, ResponseData].

I don't think that makes sense because a provider is registered for a single class. Maybe you're trying to refer to page objects?

> `get_providers` will always return both with the current implementation.

Yes , that's true. But so far, I've not seen a single example making use of Union. It seems like we're selecting page objects (and therefore providers) using another method. Maybe @kmike could give a word here. 

> This will force Response to be considered as required even if it is not (AutoExtractResponseData does not require Response)

Again, providers need one – and only one – page input class. I don't see this issue. But I might be biased tho.

tests/test_scrapy_settings.py
victor-torres: thanks 😊

scrapy_poet/middleware.py
```diff
@@ -39,20 +42,30 @@ def process_request(self, request: Request, spider: Spider):
...
         """
         # Find out the dependencies
         callback = utils.get_callback(request, spider)
-        plan, provider_instances = utils.build_plan(callback, response)
+        dependencies = {
+            Request: request,
+            Response: response,
+            Crawler: spider.crawler,
+            Settings: spider.settings,
+            StatsCollector: spider.crawler.stats,
```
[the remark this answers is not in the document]
victor-torres: Is there any particular reason for that? Are you afraid of undesired API changes?

scrapy_poet/utils.py
victor-torres: Maybe `get_callback_providers` or `get_providers_from_callback`.

victor-torres: Oh, I see your point. Yes, it's more like a search indeed.

scrapy_poet/utils.py
victor-torres: Yeah, this is an issue that's more likely to happen with the use of type annotations.

Also prefer long lines in cases like this (function definitions). Let me try to make it work and see if our pipeline is complaining.

tests/test_scrapy_settings.py
victor-torres: I think @ivanprado has proposed the same thing: create some providers for the types we're injecting... Let me see how it looks like.

kmike:
@ivanprado wouldn't Request cover the use case you have in mind for DummyResponse?

```py
    @provides(ApiResponse)
    class ApiResponseProvider(PageObjectInputProvider):
        def __init__(self, request: Request):
            self.url = request.url

        async def __call__(self):
            data = await api_request(self.url)
            return ApiResponse(data=data)
```

If a provider doesn't need a response, it doesn't ask for it. Callbacks require Response implicitly, but they can opt-out using DummyResponse.

victor-torres:
Hey, @ivanprado. 

I don't think ResponseData should be used like this:

```python
def parse(response: DummyResponse, data: ResponseData):
   ...
```

I think ResponseData should be used by page objects, not directly by callbacks. 

Regarding DummyResponse, @kmike has already answered that. We should use Request instead.

docs/advanced.rst
```diff
@@ -41,10 +41,56 @@ that need it, like the ``ItemWebPage``.
...
+
+    This behavior is related to covariance.
+    In simple words,
+    this means a subclass could replace a parent class
+    but a parent class could not substitute its subclasses.
+    Example:
+
+    .. code-block:: python
+
+        class Animal:
```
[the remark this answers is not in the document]
victor-torres: I was trying to explain the concept using a simpler (and maybe funny) example  😄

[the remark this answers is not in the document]
victor-torres: I've slightly changed the docs text keeping the simple example. Let me know if you still think it should be replaced.

ivanprado:
Ey @victor-torres and @kmike, using `Request` here makes sense, good idea. You can also access `meta` and `cb_kwargs` and other stuff from there, right?

> I think ResponseData should be used by page objects, not directly by callbacks.

@victor-torres Sure, it was just an example. It would be the same even if it is inside on a page object that requires `ResponseData`

scrapy_poet/utils.py
victor-torres: I've just renamed it to `discover_callback_providers`. What do you think?

scrapy_poet/utils.py
victor-torres: Done.

scrapy_poet/utils.py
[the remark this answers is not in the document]
ivanprado: > But how are we supposed to discover all those unmet dependencies during runtime?

I don't get that. Could you put an example of unmet dependency?

Regarding the `Union` usage, is an original idea from @kmike so probably he can explain better. But the idea to assign `Union` an "optional" semantic. So let's say you need HTML, and you prefer it to come from the browser DOM, but you also are able to survive if it comes from direct download. You express it with the annotation `Union[SplashResponseData, ResponseData]`. In some environments, Splash will be available and in some others, it won't, but the page object will still work in both cases.

> Again, providers need one – and only one – page input class.

Yes, but this is not broken with the `Union`. The `Union` here means that one or the other will be supplied. But only one will be supplied at the end. So there are two possible providers, and only one will be invoked at the end.

scrapy_poet/middleware.py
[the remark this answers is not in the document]
ejulio: Just that, I prefer low coupling and something like `spider.crawler.stats` has 3 levels of coupling.
Not that is going to change, but we can keep concise on the middlewares' dependencies

tests/test_scrapy_settings.py
victor-torres: I've updated the pull request including some code to reduce boilerplate required by those Scrapy-provided dependencies.

victor-torres:
> @victor-torres Sure, it was just an example. It would be the same even if it is inside on a page object that requires ResponseData

@ivanprado, according to our latest discussion, we should not use those Scrapy objects inside Page Objetcs, but from Providers only. I am currently working trying to enforce this limitation to avoid bad design choices from developers when using this lib.

victor-torres:
@kmike and @ivanprado, I've just updated this pull request. 
Now, Scrapy dependencies are injected into Providers only.
They're not available for Page Objects.
Test cases were updated to reflect that.

Do you mind taking another look at things?
I'd like to see if there's anything else pending,
or if you have any suggestions about improving docs.

docs/advanced.rst
```diff
@@ -41,10 +41,57 @@ that need it, like the ``ItemWebPage``.
...
+- :class:`~scrapy.http.Response`
+- :class:`~scrapy.crawler.Crawler`
+- :class:`~scrapy.settings.Settings`
+- :class:`~scrapy.statscollectors.StatsCollector`
+
 .. warning::
 
-    Currently, scrapy-poet is only able to inject ``Response`` and
-    :class:`~.DummyResponse` instances as *provider* dependencies.
+    If your provider needs to use a subclass of Response,
```
ivanprado, line 55: I'm not sure this explanation is now that relevant in that form. The developer asks for a `Response` in the provider and can receive a subclass of it, but this is also the case in the regular Scrapy callbacks. So nothing new here, right?.

scrapy_poet/utils.py
```diff
@@ -99,19 +76,9 @@ def is_provider_using_response(provider):
     """Check whether injectable provider makes use of a valid Response."""
```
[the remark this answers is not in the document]
victor-torres: `discover_callback_providers` method was updated to make use of `andi.plan` again.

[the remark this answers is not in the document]
victor-torres: `is_provider_using_response ` method was updated to make use of andi.plan. I had to move the `SCRAPY_PROVIDED_CLASSES` constant to another file to avoid import problems.

docs/advanced.rst
victor-torres: Yes, that's correct. I've tried to simplify the docs mentioning something like this.

victor-torres: Could you check it, please?

scrapy_poet/page_input_providers.py
```diff
@@ -8,12 +8,25 @@
...
+from scrapy.http import Request, Response
+from scrapy.crawler import Crawler
+from scrapy.statscollectors import StatsCollector
+from scrapy.settings import Settings
 from web_poet.page_inputs import ResponseData
 
 # FIXME: refactor _providers / provides / register,  make a nicer API
 providers = {}
 
+SCRAPY_PROVIDED_CLASSES = {
```
kmike, line 21: What do you think about making this constant private (_SCRAPY_PROVIDED_CLASSES)? I'm not sure it should be a part of the public API, or do you see a use case for that?

kmike: Another idea: move it to utils. Currently this module contains provider themselves, while in utils there are functions to create providers and instances.

docs/advanced.rst
```diff
@@ -41,10 +41,36 @@ that need it, like the ``ItemWebPage``.
...
 
-    Currently, scrapy-poet is only able to inject ``Response`` and
-    :class:`~.DummyResponse` instances as *provider* dependencies.
+    Scrapy doesn't know when a Request is going to generate
+    a Response, a TextResponse, an HtmlResponse,
+    or any other type that inherits from Response.
+    Because of this,
+    you should always annotate your provider's response argument
+    with the Response type.
+    If your provider needs a TextResponse,
```
ivanprado, line 61:
```suggestion
    If your provider only supports TextResponse,
```

scrapy_poet/middleware.py
```diff
@@ -39,20 +43,33 @@ def process_request(self, request: Request, spider: Spider):
...
     def process_response(self, request: Request, response: Response,
                          spider: Spider):
-        """This method instantiates all ``Injectable`` sub-classes declared as
+        """This method instantiates all ``Injectable`` subclasses declared as
         request callback arguments and any other parameter with a provider for
         its type. Otherwise, this middleware doesn't populate
         ``request.cb_kwargs`` for this argument.
 
-        .. warning::
+        Currently, we are able to inject instances of the following
```
ivanprado, line 51: Consider moving this docu section about what is injectable in providers to `PageObjectInputProvider` class or its `__init__` method. I think people writing their own providers would go there to seek for documentation.

scrapy_poet/middleware.py
```diff
@@ -39,20 +43,33 @@ def process_request(self, request: Request, spider: Spider):
     @inlineCallbacks
     def process_response(self, request: Request, response: Response,
                          spider: Spider):
-        """This method instantiates all ``Injectable`` sub-classes declared as
+        """This method instantiates all ``Injectable`` subclasses declared as
```
ivanprado, line 46: It might be also interesting to mention what happens when there is a collision between an already set `cb_kwargs` and an injectable attribute: User-defined cb_kwargs takes precedence. 

See https://github.com/scrapinghub/scrapy-poet/pull/30/files#diff-49772e5c5730b71f0e27fcc38b4f9614R80

scrapy_poet/utils.py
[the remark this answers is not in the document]
ivanprado: Already addressed. Thank you @victor-torres ! Resolving.

scrapy_poet/page_input_providers.py
victor-torres: Done.

- moved to util.py
- turned into a private constant

scrapy_poet/middleware.py
victor-torres: Done.

scrapy_poet/middleware.py
victor-torres: Done.

scrapy_poet/middleware.py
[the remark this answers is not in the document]
victor-torres: Yeah, it might show a little bit redundant, but I guess it makes it easier to express yourself when declaring provider dependencies.

scrapy_poet/page_input_providers.py
```diff
@@ -19,7 +19,39 @@ class PageObjectInputProvider(abc.ABC):
...
+            the same way you would need to do when using Scrapy callbacks.
+            Example:
+
+            .. code-block:: python
+
+                @provides(MyCustomResponseData)
+                class MyCustomResponseDataProvider(PageObjectInputProvider):
+
+                    def __init__(self, response: Response):
+                        assert issubclass(response, TextResponse)
```
kmike, line 52:
```suggestion
                        assert isinstance(response, TextResponse)
```

victor-torres:
Thank you for your time, fellows. 🎉 
@ivanprado @kmike @ejulio

PR #32, opened 2020-06-01
Feat tox docs

ejulio:
Add docs configuration to `tox.ini`
Run the configuration in travis.
Had to fix the static path that was breaking the build


kmike:
Thanks @ejulio!

PR #33, opened 2020-06-02
improve documentation

kmike:
TODO:

- [x] overview
- [x] basic tutorial
- [x] advanced tutorial
- [x] check docstrings

I've made minor changes here and there, and added more links to web-poet docs. There are still sections can be explained a bit better (e.g. DummyResponse, from an user point of view), but I think this can be done later; having integrations available in third-party packages (e.g. for Splash and AutoExtract) could also allow to make the docs better later.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -8,16 +8,17 @@ web-poet_ defines a standard for writing reusable and portable
 extraction and crawling code; please check its docs_ to learn more.
 
 By using ``scrapy-poet`` you'll be organizing the spider code in a different
-way, which allows to separate extraction and crawling logic from the I/O,
-and from the Scrapy implementation details as well. It makes the code
-more testable and reusable.
+way, which separates extraction and crawling logic from the I/O,
+and from the Scrapy implementation details as well.
+It makes the code more testable and reusable.
 
 ``scrapy-poet`` also provides a way to integrate third-party APIs
 (like `Splash`_ and `AutoExtract`_) with the spider, without losing
-testability & reusability. Concrete integrations are not provided by
-``scrapy-poet``, but ``scrapy-poet`` makes them possbile.
+testability and reusability.
+Concrete integrations are not provided by ``scrapy-poet``, but
+``scrapy-poet`` makes them possbile.
 
-To get started, see :ref:`intro-install` and the :ref:`intro-tutorial`.
+To get started, see :ref:`intro-install` and :ref:`intro-tutorial`.
 
 :ref:`license` is BSD 3-clause.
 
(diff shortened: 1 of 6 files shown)


docs/index.rst
```diff
@@ -1,43 +1,41 @@
...
 =========================
 scrapy-poet documentation
 =========================
 
-scrapy-poet easily integrates Page Objects created using `web-poet`_ with
-Scrapy through the configuration of a dependency injection middleware.
+scrapy-poet allows to use `web-poet`_ Page Objects with Scrapy.
+
+web-poet_ defines a standard for writing reusable and portable
+extraction and crawling code; please check its docs_ to learn more.
```
ejulio, line 8: Does it allow crawling?

docs/index.rst
```diff
@@ -1,43 +1,41 @@
...
 
-`web-poet`_ is used to make reusable Page Objects that separates
-extraction logic from crawling. They could be easily tested and distributed
-across different projects. Also, they could make use of different backends,
-for example, acquiring data from `Splash`_ and `AutoExtract`_ API.
+By using ``scrapy-poet`` you'll be organizing the spider code in a different
+way, which allows to separate extraction and crawling logic from the I/O,
```
ejulio, line 11: > which separates extraction and crawling logic from I/O

docs/index.rst
```diff
@@ -1,43 +1,41 @@
...
+and from the Scrapy implementation details as well. It makes the code
```
ejulio, line 12: As markdown aggregates line-breaks in the same paragraph, maybe we could leave each phrase in it's own line, making it easier to review future changes.

docs/index.rst
```diff
@@ -1,43 +1,41 @@
...
+more testable and reusable.
 
-The goal of this project is to provide a bridge between 
-:ref:`Scrapy Spiders <scrapy:topics-spiders>` and Page Objects.
+``scrapy-poet`` also provides a way to integrate third-party APIs
+(like `Splash`_ and `AutoExtract`_) with the spider, without losing
+testability & reusability. Concrete integrations are not provided by
```
ejulio, line 17: personally, I prefer _and_, but _&_ is fine

ejulio: Same as above about the phrases, in case you agree with it

docs/index.rst
kmike: What I'm thinking about here is that we should be standardizing not only `to_item()`method, but methods related to crawling as well - e.g. pagination API, or API for lists of items. Crawling code can then use these building blocks.

kmike:
Thanks for the review @ejulio! I'm going to merge it, but we may need to discuss the point about crawling later.

PR #34, opened 2020-06-09
Add AutoExtract providers for scrapy-poet

victor-torres:
- we need to decide where we're storing the source code
- I've started the implementation here to bootstrap it quicker
- some dependencies are coming from private repositories and need to be moved


autoextract_poet/exceptions.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
+from typing import List
+
+
+@dataclass
+class QueryLevelError(Exception):
```
victor-torres, line 6: This exception has way more attributes on its original form. `request` attribute was removed because it was not being used. `url` and `pageType` attributes were removed because they're already contained within the `query` attribute.

autoextract_poet/inputs.py
```diff
@@ -0,0 +1,20 @@
+from dataclasses import dataclass
+from typing import Any, Dict, Optional
+
+
```
victor-torres, line 4: Are we moving those classes to the web-poet repository?

autoextract_poet/mixins.py
```diff
@@ -0,0 +1,19 @@
+from web_poet.mixins import ResponseShortcutsMixin
+from web_poet.page_inputs import ResponseData
+
+
+class AutoExtractResponseShortcutsMixin(ResponseShortcutsMixin):
+
```
victor-torres, line 6: Suggestions wanted.

autoextract_poet/providers.py
```diff
@@ -0,0 +1,49 @@
...
+    page_type: str
+    page_input_class: Type
+
+    def __init__(self, request: Request, stats: StatsCollector):
+        self.request = request
+        self.stats = stats
+
+    async def __call__(self):
+        try:
+            # FIXME: how to configure if you want full HTML or not?
```
victor-torres, line 28: We could receive Scrapy Settings on the `__init__` method, but do you think that's really necessary for now?

autoextract_poet/providers.py
```diff
@@ -0,0 +1,49 @@
...
+            self.stats.inc_value(f"autoextract/{self.page_type}/error")
+            raise
+
+        self.stats.inc_value(f"autoextract/{self.page_type}/success")
+        return self.page_input_class(data=data)
+
+
+@provides(AutoExtractProductResponseData)
+class AutoExtractProductResponseDataProvider(AutoExtractProvider):
+
```
victor-torres, line 40: Maybe rename to `autoextract_page_type`.

autoextract_poet/utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+        # A single url send, so a single element in the response
+        response = response_per_url[0]
+
+        if "error" in response:
+            raise QueryLevelError(query, response)
+
+        return response
+
+    # FIXME: move retry logic to this lib
```
victor-torres, line 21: This library is in a private repository. We need to move this code to this pull request and make sure it's using our implementation of the `QueryLevelError` exception.

tests/test_autoextract_poet.py
```diff
@@ -0,0 +1,21 @@
+from autoextract_poet.inputs import (
+    AutoExtractResponseData,
+    AutoExtractProductResponseData,
+    AutoExtractProductListResponseData,
+)
+from autoextract_poet.exceptions import QueryLevelError
+
```
victor-torres, line 7: This is just a temporary test file. We should add way more test cases here.

autoextract_poet/exceptions.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
```
victor-torres, line 1: `dataclasses` is not available on Python 3.6 😕

autoextract_poet/exceptions.py
ivanprado: > request attribute was removed because it was not being used

I think you mean "response". I see I'm not saving it in the QueryLevelError, but I think it could be useful to keep the full response there. 

> url and pageType attributes were removed because they're already contained within the query attribute.

Understand your point here. I'm unsure about what is best to do with a case like that. 
`query` contains all the parameters as you pointed but at the same time is not making explicit what is inside so it is more difficult for developers to know what it contains (they need to read the AutoExtract service documentation). That's what I opted to make attributes like `url` and `pageType` explicit.

autoextract_poet/mixins.py
```diff
@@ -0,0 +1,19 @@
...
+    @property
+    def response(self):
+        # FIXME: raise a custom Exception or create a separate method for errors
```
ivanprado, line 9: This comment is outdated. Errors are not expected at this stage. The response should be well-formed here.

autoextract_poet/mixins.py
```diff
@@ -0,0 +1,19 @@
...
+        data = self.autoextract_response.data
+
+        # FIXME: pageType could be different to item key in the response
```
ivanprado, line 12: I don't know what I'm saying here :stuck_out_tongue:

autoextract_poet/mixins.py
ivanprado: This was a trick to being able to use the HTML returned by AutoExtract the same way that is used in `ResponseData`. I see several problems here:
* HTML won't be always present. We need a way so that the user can configure what he wants. Having a set of new classes (e.g. AEProductPageHTML) just to configure that HTML is required seems explosive. 
* Probably it would be nice to have some kind of `AEResponseData` class for that. So that when you request `AEResponseData` your receive a `ResponseData` compatible class but with HTML coming from AutoExtract. The problem is that requiring both `AEResponseData` and `AEProductPage` should not raise two requests to AutoExtract, but only one. This breaks the current provider structure. We should think about it.

autoextract_poet/mixins.py
ivanprado: Ohh, I remember. I'm using pageType here to get the data from the `data` dict. For example, for `product` you have the product within the `product` key. But I'm not sure this is going to be the case for all pageTypes... We have to check the documentation https://doc.scrapinghub.com/autoextract.html

autoextract_poet/mixins.py
```diff
@@ -0,0 +1,19 @@
...
+        pageType = data["query"]["userQuery"]["pageType"]
+
+        # FIXME: maybe better to have a AEResponseData to not confuse with the ResponseData semantic
```
ivanprado, line 15: Ideally, we should have an `AEResponseData` that could be injectable. But at the same time we should group requests to AutoExtract. See my comment above.

autoextract_poet/providers.py
ivanprado: I'm using this in the spyder where I'm always requiring HTML, but the general case will be not requesting HTML (by default it is not supplied for regular accounts). I see three ways of configuring that:
* By configuring the provider from settings
* By configuring the provider from request meta
* By deducting it by annotating something as `AEResponseData`

Probably the best is the last one, but it requires changes in the providers infrastructure to support it. And there are complexities like what happen if you request `AEProductPage`, `AEArticlePage` and `AEResponseData`. What is inside `AEResponseData` in this case, the HTML from the product or from the article extraction?

autoextract_poet/utils.py
ivanprado: Yes, the retry logic and the exception, etc, should be part of the public repository.

autoextract_poet/exceptions.py
victor-torres: Yes, it's supposed to be `response` and not `request` 😅 
Yes, the values appear in a list here, that's how the query is built 😕  
But think it should easy to spot what's the URL and what's the AutoExtract page type.
Does that make sense?

autoextract_poet/mixins.py
victor-torres: What about leaving HTML content out of the scope for a while? I mean, AutoExtract biggest advantage is not having to parse HTML content. We could address it in another ticket, meanwhile, users could use regular responses from the spider callback.

autoextract_poet/providers.py
victor-torres: How common is it to request the HTML response from the AutoExtract API? Thinking about skipping it here on this pull request and address it later...

autoextract_poet/exceptions.py
victor-torres: I also have mixed feelings here:
- I'd like to ditch Python 3 since this is a new project
- Agree with you that this could represent a barrier when migrating existing spiders
- Although, I believe Python 3.6 spiders could be easily migrated to Python 3.7, as opposed to Python 2 spiders being migrated to Python 3.X

victor-torres: @kmike, we're welcome to share your two cents as well 😄

victor-torres:
Thank you for your thoughts here, @ivanprado. Tomorrow I'll be moving the retry logic from the private repository to this pull request. Agree that we could have more general improvements here, but maybe we could focus on a MVP first. Let's discuss them tomorrow.

autoextract_poet/pages.py
```diff
@@ -0,0 +1,38 @@
...
+    AutoExtractProductResponseData,
+    AutoExtractProductListResponseData,
+)
+
+
+@dataclass
+class AutoExtractProductPage(ItemPage, AutoExtractResponseShortcutsMixin):
+    """Generic product page."""
+
+    autoextract_response: AutoExtractProductResponseData
```
ejulio, line 16: In regular page objects we don't have access to the response structure.
So, maybe, to keep a consistent interface, we should only have access to `AutoExtractProductResponseData.data` in the page object.
Not that it is a problem, though

autoextract_poet/pages.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@dataclass
+class AEProductListPage(ItemPage, AutoExtractResponseShortcutsMixin):
+    """Generic product list page."""
+
+    autoextract_response: AutoExtractProductListResponseData
+
+    def breadcrumbs(self):
+        return self.autoextract_response.data.get("productList", {}).get("breadcrumbs", [])
```
ejulio, line 30: Just some rant :joy: 
Can we do better than `obj.get('some key', default).get('other key', default)` and have it as 
`obj['some key']['other key']`, basically having the defaults under the hood?
Not a problem here, just thinking about it..

autoextract_poet/providers.py
```diff
@@ -0,0 +1,49 @@
...
+    page_input_class: Type
+
+    def __init__(self, request: Request, stats: StatsCollector):
+        self.request = request
+        self.stats = stats
+
+    async def __call__(self):
+        try:
+            # FIXME: how to configure if you want full HTML or not?
+            data = await request(self.request.url, self.page_type)
```
ejulio, line 29: Since we're counting stats for success/error, maybe count a global one for requests?
This way, we can cross check the numbers, success + errors = global

autoextract_poet/providers.py
ejulio: I guess the context gives the idea of autoextract

autoextract_poet/utils.py
```diff
@@ -0,0 +1,22 @@
...
+from autocrawl.autoextract_retry import autoextract_retry
+
+from autoextract_poet.exceptions import QueryLevelError
+
+
+async def request(url: str, page_type: str, full_html=True):
+    query = [dict(url=url, pageType=page_type, fullHtml=full_html)]
+
+    async def _request():
+        response_per_url = await request_raw(query)
```
ejulio, line 11: Shouldn't we send some APIKEY here too?

autoextract_poet/pages.py
ejulio: Maybe, having `jmes` like we have `xpath` and `css` for regular HTML responses

autoextract_poet/inputs.py
```diff
@@ -0,0 +1,20 @@
...
+@dataclass
+class AutoExtractResponseData:
+    """Represent an AutoExtract API response containing returned data."""
+    data: Optional[Dict[str, Any]]
+
+
+@dataclass
+class AutoExtractProductResponseData(AutoExtractResponseData):
```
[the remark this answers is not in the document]
ivanprado: I think we already need different providers per page type.

autoextract_poet/providers.py
victor-torres: I've renamed class names.

autoextract_poet/utils.py
victor-torres: I've moved the code from the private repository to this pull request.

autoextract_poet/pages.py
victor-torres: I've refactored this class to make use of a private property. I was not including another dependency (jmes) for this simple use case xD But thank you for the suggestion.

ivanprado:
It seems is going to be important to be able to enable/disable parameters per request (e.g. `fullHtml`). A proposal could be: have key in `request.meta` that stores a list of objects that can be injected to the providers in the middleware. 

Example:
```py
class AutoExtractRequestParams(dict):
    pass

class AutoExtractProvider(PageObjectInputProvider):

    page_type: str
    page_input_class: Type

    def __init__(self, request: Request, stats: StatsCollector, params: Optional[AutoExtractRequestParams]):
       ...

def parse(response, product_list: AutoExtractProductListPage):
  params = AutoExtractRequestParams(fullHtml=True)
  meta = {PROVIDERS_ADDITIONAL_INJECTABLES: [params]}
  response.follow_all(product_list.urls(), callback=self.parse_product, meta=meta)
```

This have some limitations:

-  Same parameters shared to all AutoExtract requests within the same callback. This could be addressed by having a different params class for each pageType.

It might be better to address this case in a separate PR, though.

autoextract_poet/inputs.py
```diff
@@ -0,0 +1,20 @@
+from dataclasses import dataclass
+from typing import Any, Dict, Optional
+
+
+@dataclass
+class ResponseData:
```
ivanprado, line 6: Naming question. Two options:

1. Kind explicit in the class name. `ResponseData` (for scrapy data), `AutoExtractResponseData`, `SplashResponseData`
2. Use the module name to spot the kind `scrapy.ResponseData`, `autoextract.ResponseData`, `splash.ResponseData`

You choose 2. I think I agree, but not totally sure. Let's see if the rest have any opinion.

autoextract_poet/providers.py
```diff
@@ -0,0 +1,49 @@
...
+
+    page_type = "product"
+    page_input_class = ProductResponseData
+
+
+@provides(ProductListResponseData)
+class AutoExtractProductListResponseDataProvider(AutoExtractProvider):
+
+    page_type = "productList"
+    page_input_class = ProductListResponseData
```
ivanprado, line 49: `ProductListResponseData` has to be repeated twice here. What if we allow injecting the registered type so that it is received in the init. I mean creating an object:
```py
@dataclass
class ClassToProvide:
  cls: Type
```

And modify the Middleware so that you can receive it in the `__init__`:
```py
class AutoExtractProvider(PageObjectInputProvider):
    page_type: str

    def __init__(self, request: Request, stats: StatsCollector, cls_to_provide: ClassToProvide):
        self.request = request
        self.stats = stats
        self.page_input_class = cls_to_provide.cls

   ...
```

autoextract_poet/exceptions.py
kmike: There is a backport for Python 3.6, which can be conditionally installed in setup.py.

autoextract_poet/mixins.py
victor-torres: The mixin was entirely removed for now.

autoextract_poet/mixins.py
victor-torres: The mixin was entirely removed for now.

autoextract_poet/mixins.py
victor-torres: The mixin was entirely removed for now.

autoextract_poet/exceptions.py
victor-torres: Fixed

autoextract_poet/inputs.py
kmike: I think yes, that's the way to go. Probably makes sense to create web_poet.inputs.autoextract namespace/package, or maybe web_poet.autoextract.

autoextract_poet/inputs.py
kmike: I see an appeal of `autoextract.ResponseData`, but on the other hand, it has bitten me in past. Consider how it can be used. This is nice:

```py
from web_poet.inputs import autoextract

class Foo:
    data: autoextract.ResponseData
```

But there is also "autoextract" package (via https://github.com/scrapinghub/scrapinghub-autoextract), so it gets ugly when you need to import both:

```py
import autoextract
from web_poet.inputs import autoextract  # what to do?

# option 1:
from web_poet.inputs.autoextract import ResponseData  # clash with other ResponseData types

# option 2:
import web_poet.inputs.autoextract
data = web_poet.inputs.autoextract.ResponseData  # super verbose
```

You can also have `import autoextract` in `web_poet.inputs.__init__.py`; this allows the following:

```py
from web_poet import inputs
data = inputs.autoextract.ResponseData
```

which is less verbose, and not bad actually - though people will be importing ResponseData directly, making the code harder to read.

Alternative is, as @ivanprado said, to use `AutoExtractResponseData` name instead. I don't have a strong opinion on `AutoExtractResponseData` vs `inputs.autoextract.ResponseData`, but other options look worse to me (`ResponseData`, `web_poet.inputs.autoextract.ResponseData` and `autoextract.ResponseData`).

autoextract_poet/inputs.py
```diff
@@ -0,0 +1,20 @@
...
+    """Represent an AutoExtract API response containing returned data."""
+    data: Optional[Dict[str, Any]]
+
+
+@dataclass
+class ProductResponseData(ResponseData):
+    """Represent an AutoExtract Product response."""
```
kmike, line 13:
```suggestion
    """ AutoExtract Product response."""
```

and the same for other data types :)

autoextract_poet/inputs.py
```diff
@@ -0,0 +1,20 @@
...
+    pass
+
+
+@dataclass
+class ProductListResponseData(ResponseData):
```
kmike, line 18: I like how they all inherit from a base class.

autoextract_poet/pages.py
```diff
@@ -0,0 +1,41 @@
...
+from web_poet.pages import ItemPage
+
+from autoextract_poet.inputs import (
+    ProductResponseData,
+    ProductListResponseData,
+)
+
+
+@dataclass
+class ProductPage(ItemPage):
```
kmike, line 12: I think that's fine to keep base classes for pages out of scope for this PR. This looks like a separate task.

autoextract_poet/providers.py
```diff
@@ -0,0 +1,49 @@
...
+
+    async def __call__(self):
+        try:
+            # FIXME: how to configure if you want full HTML or not?
+            data = await request(self.request.url, self.page_type)
+        except Exception:
+            self.stats.inc_value(f"autoextract/{self.page_type}/error")
+            raise
+
+        self.stats.inc_value(f"autoextract/{self.page_type}/success")
```
kmike, line 34: nice use of stats

autoextract_poet/retry.py
```diff
@@ -0,0 +1,162 @@
...
+from autoextract_poet.exceptions import QueryLevelError
+
+logger = logging.getLogger(__name__)
+
+_DOMAIN_OCCUPIED_RE = re.compile(
+    r".*domain .* is occupied, please retry in (.*) seconds.*", re.IGNORECASE
+)
+
+
+def _domain_occupied(msg: Optional[str]) -> Optional[float]:
```
kmike, line 42: That's awful we need to write such code in a client library; API should have exposed this in a structured format. That's good to ahve support for it in the library now, but I hope we'll be able to remove it in future.

autoextract_poet/utils.py
kmike: I think it'd be better to take the API key from Scrapy settings, as it'd be consistent with scrapy-autoextract. To do it, request function could accept api_key argument, and a provider may require Settings.

setup.py
```diff
@@ -1,5 +1,22 @@
...
+    requirements = [
+        'andi>=0.3',
+        'attrs',
+        'parsel',
+        'tenacity',
+        'web-poet',
+    ]
+
+    if not sys.version_info >= (3, 7):
+        requirements.append('dataclasses')
```
kmike, line 15: I think that's better to use https://www.python.org/dev/peps/pep-0508/#environment-markers; dynamic code in setup.py may not be the best for wheels - I'm not sure it would cause an actual issue here though.

autoextract_poet/retry.py
```diff
@@ -0,0 +1,162 @@
...
+        exc = retry_state.outcome.exception()
+        if _is_domain_occupied_error(exc):
+            return stop_never
+        elif _is_retriable_error(exc):
+            return self.stop_on_15_minutes(retry_state)
+        else:
+            raise RuntimeError("Invalid retry state exception: %s" % exc)
+
+
+autoextract_retry = retry(
```
kmike, line 156: After seeing the code again, I think it can be better to move it to scrapinghub-autoextract eventually (//cc @ivanprado). At least move retry definitions there, but maybe not enable them by default for lower-level usage.

autoextract_poet/mixins.py
victor-torres: Removed for now.

autoextract_poet/exceptions.py
victor-torres: After some refactoring, exception should look like this on console:

```python
autoextract_poet.utils.QueryLevelError: (Query(url='example.com', page_type='myPageType', full_html=True), 'error')
```

autoextract_poet/providers.py
victor-torres: Done.

autoextract_poet/inputs.py
victor-torres: Renamed to `AutoExtractResponseData` as it doesn't hurt :)

autoextract_poet/inputs.py
victor-torres: Done.

autoextract_poet/providers.py
victor-torres: That's @ivanprado's idea :D

setup.py
victor-torres: #TIL I'll give it a try

autoextract_poet/inputs.py
victor-torres: Actually, we're only using this `ResponseData` class within our own code. As we're planning to move this to `web_poet`, I've changed this back to `ResponseData` and we should use it as a namespace as suggested by @ivanprado and @kmike.

setup.py
victor-torres: Done.

autoextract_poet/pages.py
victor-torres: Done.

autoextract_poet/inputs.py
ejulio: Wouldn't it create some sort of coupling of webpoet and our technology (AutoExtract)?
Maybe a new package?

ejulio:
About @ivanprado comment
> It seems is going to be important to be able to enable/disable parameters per request (e.g. fullHtml). A proposal could be: have key in request.meta that stores a list of objects that can be injected to the providers in the middleware.

We need to be careful that, the data that goes in `meta` must be serializable and we need to test it in `dash` in a given moment.
I recall that we can send `loader` in `meta` while working locally, but once it is the requests fail to serialize in `dash`, as it is a different routine or so..

autoextract_poet/providers.py
ejulio: Another approach could be by using types.
We could've a `AutoExtractFullHtmlProvider`  and `ProductResponseFullHtmlData`.
Not sure if it is better, but it gives control on the request level to enable/disable HTML response.
For example, for product listing I don't want the HTML only what AE returns, but for product pages I want everything.
The problem is that it creates more classes and granularity, which can be a bit convoluted

autoextract_poet/providers.py
ejulio: I guess it's a good idea, just don't know if "too magic", like hard to trace where things are happening.
Also, what would happen if the user requests for a class that is not available in AutoExtract (something that is not a product, listing, ...)?

autoextract_poet/retry.py
victor-torres: I've started the move https://github.com/scrapinghub/scrapinghub-autoextract/pull/7

victor-torres:
Quick update about this PR:
- This PR is almost ready to be merged scrapinghub/scrapinghub-autoextract#7
- This PR is related to this one scrapinghub/web-poet#8

It seems like I'll be working on it this or next week.

autoextract_poet/query.py
```diff
@@ -0,0 +1,21 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class Query:
+
+    url: str
+    page_type: str
+    # FIXME: is this resource available for everyone?
+    # FIXME: decide if this should be True or False by default
```
victor-torres, line 10: cc @ivanprado @kmike @ejulio

victor-torres: The resource in question is `full_html`.

scrapy_poet/repository.py
```diff
@@ -0,0 +1,55 @@
...
+
+providers = {}
+
+
+def register(provider_class: typing.Type[PageObjectInputProvider],
+             provided_class: typing.Optional[typing.Type] = None):
+    """This method registers a Page Object Input Provider in the providers
+    registry. It could be replaced by the use of the ``provides`` decorator.
+
+    If ``provided_class`` is not specified, we get it from ``provider_class``.
```
ivanprado, line 25: Maybe it makes sense to force the providers to have `provided_class`  always defined. And then change the decorator `provides` by `provider`. And the decorator also looks for `provided_class`. This way you could remove the lines `register(...)` below and replace them by a decorator in top of the provider classes.

scrapy_poet/repository.py
```diff
@@ -0,0 +1,55 @@
...
+import typing
+
+from scrapy_poet import autoextract
+from scrapy_poet.providers import (
+    PageObjectInputProvider,
+    ResponseDataProvider,
+)
+# FIXME: refactor _providers / provides / register,  make a nicer API
+
+providers = {}
```
ivanprado, line 17: I'm thinking that we should maybe have a registry per spider/crawler in addition to the global one. So when you start a Spider, the global providers are copied to the local spider/crawler registry. And we could also provide some hooks to modify the spider/crawler providers. This can be useful for unit testing for example, where you can be mocking the providers to test an Spider.

Another use case could be to be able to generate providers on the fly. But not sure when would that be required. 

It can also be useful to disable some providers if you don't want them available.

scrapy_poet/autoextract.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@attr.s(auto_attribs=True, slots=True)
+class Query:
+
+    url: str
+    page_type: str
+    # FIXME: is this resource available for everyone?
+    # FIXME: decide if this should be True or False by default
+    full_html: bool = True
```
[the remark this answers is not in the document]
victor-torres: Done

scrapy_poet/autoextract.py
```diff
@@ -0,0 +1,71 @@
...
+    ArticleResponseData,
+)
+
+
+@attr.s(auto_attribs=True, slots=True)
+class Query:
+
+    url: str
+    page_type: str
+    # FIXME: is this resource available for everyone?
```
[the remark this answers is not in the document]
victor-torres: Hey, @ivanprado. 

I've added the new attributes `meta` and `articleBodyRaw`. 
I've also added a new parameter called `extra`, an optional dictionary that will be expanded to update the final query object.

Let me know if you find that we can drop the `articleBodyRaw`, for example, now that we have the `extra` attribute.

scrapy_poet/autoextract.py
```diff
@@ -0,0 +1,71 @@
...
+    # FIXME: decide if this should be True or False by default
+    full_html: bool = True
+
+    @property
+    def autoextract_query(self):
+        return [
```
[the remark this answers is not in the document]
victor-torres: Done.

scrapy_poet/repository.py
victor-torres: That's a good idea. I see the potential coming from it. It's just probably out of scope here, would you mind creating an issue/ticket for it?

scrapy_poet/repository.py
victor-torres: It's very tempting. But see that the main reason for calling `register` bellow is to avoid circular imports and weird usages like having to import classes or modules without using them just to make the providers register themselves through the use of a decorator. 

@kmike has suggested making it somehow mandatory like we having a place on our spider settings where we should list and register all required providers. This is actually following the opposite path of your suggestion. 

I think we should discuss pros and cons together.

scrapy_poet/autoextract.py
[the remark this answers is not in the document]
ivanprado: I would keep `articleBodyRaw` given that is clearly stated in the documentation: https://doc.scrapinghub.com/autoextract.html

scrapy_poet/repository.py
ivanprado: >  like having to import classes or modules without using them just to make the providers register themselves

I understand. Makes sense. But this does not affect the idea of having the property `provided_class` always defined, right?

> Having a place on our spider settings where we should list and register all required providers

Yes, I like something like that and is in sync with my idea exposed here: https://github.com/scrapinghub/scrapy-poet/pull/34#discussion_r466527309

scrapy_poet/repository.py
ivanprado: Not sure it is out of the scope given the idea exposed at https://github.com/scrapinghub/scrapy-poet/pull/34#discussion_r466707862 which is basically forget about having a static registry of providers and have it in the spider. How and where to register the providers seems an important question and I would probably tackle it here. But I also understand if this wants to be done in two parts, first merging this PR, and then open a new one to work on spider lifecycle providers.

autoextract_poet/query.py
victor-torres: This has been resolved since we've changed `fullHtml` to `False` by default.

scrapy_poet/repository.py
victor-torres: > I understand. Makes sense. But this does not affect the idea of having the property provided_class always defined, right?

You're right. I've just refactored this decorator and the register method as well. Could you check it, please?

scrapy_poet/repository.py
victor-torres: I'd really like to implement it. I do think it's better to manually register the providers, similarly to what we have to do with Django applications. Let's just leave it to the next pull request because we need to agree where this will take place and that could prevent us from merging the other benefits of this pull request.

ivanprado: Makes sense to me :+1:

scrapy_poet/repository.py
```diff
@@ -17,37 +17,27 @@
...
-    registry. It could be replaced by the use of the ``provides`` decorator.
+    registry.
 
-    If ``provided_class`` is not specified, we get it from ``provider_class``.
-
-    Examples:
-
-        register(ResponseDataProvider)
-        register(ResponseDataProvider, ResponseData)
+    It could be replaced by the use of the ``provides`` decorator but you need
```
ivanprado, line 24:
```suggestion
    It could be replaced by the use of the ``provider`` decorator but you need
```

scrapy_poet/repository.py
```diff
@@ -17,37 +17,27 @@
...
+    to make sure the module containing those classes are imported during
```
ivanprado, line 25:
```suggestion
    to make sure the module containing those classes is imported at
```

scrapy_poet/repository.py
```diff
@@ -17,37 +17,27 @@
...
+    runtime.
     """
-    if provided_class is None:
-        provided_class = provider_class.provided_class
-
-    providers[provided_class] = provider_class
+    providers[provider_class.provided_class] = provider_class
```
ivanprado, line 28: I would check here if `provided_class` is actually in `provider_class`. If not I would raise a meaningful exception.

scrapy_poet/repository.py
ivanprado: Checked. I left some comments.

autoextract_poet/inputs.py
victor-torres: We'll now have  `autoextract-poet` repository for those classes.

autoextract_poet/pages.py
victor-torres: This was addressed and now we have `data`.

autoextract_poet/utils.py
victor-torres: Now, Scrapy Settings is listed as a dependency of this provider. 
We're trying to get the API Key from the settings under the key name `SCRAPINGHUB_AUTOEXTRACT_KEY`.

autoextract_poet/providers.py
victor-torres: There's also some indication from @kmike that we should be registering those providers manually (probably in the Scrapy Settings).

scrapy_poet/repository.py
victor-torres: I've just added an assertion above this line. It should provide enough information about what's happening.

scrapy_poet/autoextract.py
```diff
@@ -0,0 +1,87 @@
...
+
+from autoextract.aio.client import request_raw
+from autoextract_poet.page_inputs import (
+    ProductResponseData,
+    ArticleResponseData,
+)
+
+
+@attr.s(auto_attribs=True, slots=True)
+class Query:
```
kmike, line 18: There is https://github.com/scrapinghub/scrapinghub-autoextract/blob/590e755710e3d4408b8ec1362e99a5f8af113363/autoextract/request.py#L10, which is supported by the library directly; I wonder if you can use it instead. `extra` support could be also added there.

victor-torres: Documenting: this was moved to https://github.com/scrapinghub/scrapinghub-autoextract/pull/14/files

victor-torres:
This draft is being closed because the providers' code will be put on the scrapy-autoextract library.

PR #35, opened 2020-07-17
Switch PYPI key

imduffy15:
This should fix issues with travis ci and pypi pushing after the rename of scrapy-po to scrapy-poet

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: python
-
 branches:
   only:
   - master
-
 matrix:
   include:
   - python: 3.6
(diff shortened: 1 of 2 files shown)


kmike:
Thanks @imduffy15!

.travis.yml
```diff
@@ -16,20 +14,16 @@ matrix:
...
-
 script: tox
-
 after_success:
 - codecov
-
 deploy:
   provider: pypi
-  user: kmike
+  user: scrapinghub
```
kmike, line 24: @imduffy15 should it be `__token__` as well?

imduffy15: To my surprise yes, pypi auto figures the user from the token.

PR #38, opened 2020-09-14
Add before and after hooks to providers

victor-torres:
Before and After hooks could be very helpful when implementing providers that are – somehow – aware of each other.

For example, you can aggregate similar providers that require external API calls and combine those requests into a single one, optimizing resources and reducing network overhead.

One could think that these hooks are against the web-poet proposal, but providers are specified by scrapy-poet and not by the former library. web-poet makes no assumptions about dependencies and how they should be fulfilled.

- [x] write tests
- [x] implement feature
- [ ] update documentation

### DIFF
--- a/scrapy_poet/utils.py
+++ b/scrapy_poet/utils.py
@@ -154,11 +154,23 @@ def build_providers(instances) -> Dict[Type, PageObjectInputProvider]:
 def build_instances(plan: andi.Plan, providers):
     """Build the instances dict from a plan."""
     instances = {}
+
+    for cls, kwargs_spec in plan:
+        provider = providers.get(cls)
+        if provider and hasattr(provider, "__before__"):
+            provider.__before__()
+
     for cls, kwargs_spec in plan:
         if cls in providers:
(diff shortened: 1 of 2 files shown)


ejulio:
The code looks good, but I don't see how you could do
> For example, you can aggregate similar providers that require external API calls and combine those requests into a single one, optimizing resources and reducing network overhead.

Can you provide an example?

victor-torres:
Thank you for your review, @ejulio. 

Let's keep this pull request on hold for a while. It might not even be necessary anymore after a couple of suggestions from @ivanprado on [scrapy-autoextract repo](https://github.com/scrapinghub/scrapy-autoextract/pull/15#issuecomment-692610535).

victor-torres:
@ejulio, the example you need is actually the pull request I've mentioned above.

victor-torres:
We won't need those hooks if this new provider interface proposal is accepted: https://github.com/scrapinghub/scrapy-poet/pull/40

ivanprado:
Closing in favour of https://github.com/scrapinghub/scrapy-poet/pull/41

PR #40, opened 2020-09-18
New provider interface

victor-torres:
This new interface expects providers to specify a class method with a set of provided classes. 

When calling providers with this new interface, injection middleware should pass an argument with a sub-set of those provided classes that are currently being provided. This way, the provider could return a dictionary with its keys being the classes specified in this sub-set and the values being the desired output.

Providers that need to provide instances for a single class should just go ahead and specify a set with a single element.

- [ ] update documentation

### DIFF
--- a/scrapy_poet/middleware.py
+++ b/scrapy_poet/middleware.py
@@ -74,9 +74,11 @@ class InjectionMiddleware:
         }
         assert set(external_dependencies.keys()) == _SCRAPY_PROVIDED_CLASSES
 
+        providers = spider.settings["SCRAPY_POET_PROVIDERS"]
+
         # Build all instances declared as dependencies
-        plan = utils.build_plan(callback)
-        instances = yield from utils.build_instances(plan, external_dependencies)
+        plan = utils.build_plan(callback, providers)
+        instances = yield from utils.build_instances(plan, providers, external_dependencies)
 
         # Fill the callback arguments with the created instances
         for arg, value in plan.final_kwargs(instances).items():
--- a/scrapy_poet/middleware.py
+++ b/scrapy_poet/middleware.py
@@ -2,7 +2,6 @@
 responsible for injecting Page Input dependencies before the request callbacks
 are executed.
 """
-
 from scrapy import Spider
 from scrapy.crawler import Crawler
 from scrapy.http import Request, Response
@@ -11,7 +10,47 @@ from scrapy.statscollectors import StatsCollector
 from twisted.internet.defer import inlineCallbacks, returnValue
 
 from scrapy_poet import utils
-from scrapy_poet.utils import _SCRAPY_PROVIDED_CLASSES
+from scrapy_poet.page_input_providers import PageObjectInputProvider
+from scrapy_poet.utils import _SCRAPY_PROVIDED_CLASSES, load_provider_classes
+
+
+Provider = PageObjectInputProvider
+
+
+class ProviderManager:
+
+    def __init__(self, crawler: Crawler):
+        self.crawler = crawler
+        self.providers = self.load_providers()
+
+    @inlineCallbacks
+    def __call__(self, spider: Spider, request: Request, response: Response):
+        callback = utils.get_callback(request, spider)
+        plan = utils.build_plan(callback, self.providers)
+
+        scrapy_provided_dependencies = {
+            Spider: spider,
+            Request: request,
+            Response: response,
+            Crawler: spider.crawler,
+            Settings: spider.settings,
+            StatsCollector: spider.crawler.stats,
+        }
+        assert scrapy_provided_dependencies.keys() == _SCRAPY_PROVIDED_CLASSES
+
+        provider_instances = yield from utils.build_instances(
+            plan,
+            self.providers,
+            scrapy_provided_dependencies
+        )
+        final_kwargs = plan.final_kwargs(provider_instances)
+        raise returnValue(final_kwargs)
+
+    def load_providers(self):
+        return [
+            cls(self.crawler)
+            for cls in load_provider_classes(self.crawler.settings)
+        ]
 
 
 class InjectionMiddleware:
@@ -20,6 +59,14 @@ class InjectionMiddleware:
     * check if request downloads could be skipped
     * inject dependencies before request callbacks are executed
     """
+    def __init__(self, crawler):
+        self.crawler = crawler
+        self.provider_manager = ProviderManager(crawler)
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler)
+
     def process_request(self, request: Request, spider: Spider):
         """This method checks if the request is really needed and if its
         download could be skipped by trying to infer if a ``Response``
@@ -34,7 +81,8 @@ class InjectionMiddleware:
         actually using another source like external APIs such as Scrapinghub's
         Auto Extract.
         """
-        if utils.is_response_going_to_be_used(request, spider):
+        providers = self.provider_manager.providers
+        if utils.is_response_going_to_be_used(request, spider, providers):
             return
 
         spider.logger.debug(f'Skipping download of {request}')
@@ -63,25 +111,13 @@ class InjectionMiddleware:
         - :class:`~scrapy.statscollectors.StatsCollector`
         """
         # Find out the dependencies
-        callback = utils.get_callback(request, spider)
-        external_dependencies = {
-            Spider: spider,
-            Request: request,
-            Response: response,
-            Crawler: spider.crawler,
-            Settings: spider.settings,
-            StatsCollector: spider.crawler.stats,
-        }
-        assert set(external_dependencies.keys()) == _SCRAPY_PROVIDED_CLASSES
-
-        providers = spider.settings["SCRAPY_POET_PROVIDERS"]
-
-        # Build all instances declared as dependencies
-        plan = utils.build_plan(callback, providers)
-        instances = yield from utils.build_instances(plan, providers, external_dependencies)
-
+        final_kwargs = yield from self.provider_manager(
+            spider,
+            request,
+            response
+        )
         # Fill the callback arguments with the created instances
-        for arg, value in plan.final_kwargs(instances).items():
+        for arg, value in final_kwargs.items():
             # Precedence of user callback arguments
             if arg not in request.cb_kwargs:
                 request.cb_kwargs[arg] = value
--- a/scrapy_poet/middleware.py
+++ b/scrapy_poet/middleware.py
@@ -10,11 +10,7 @@ from scrapy.statscollectors import StatsCollector
 from twisted.internet.defer import inlineCallbacks, returnValue
 
 from scrapy_poet import utils
-from scrapy_poet.page_input_providers import PageObjectInputProvider
 from scrapy_poet.utils import _SCRAPY_PROVIDED_CLASSES, load_provider_classes
-
-
-Provider = PageObjectInputProvider
 
 
 class ProviderManager:
--- a/scrapy_poet/middleware.py
+++ b/scrapy_poet/middleware.py
@@ -64,7 +64,7 @@ class InjectionMiddleware:
         """
         # Find out the dependencies
         callback = utils.get_callback(request, spider)
-        dependencies = {
+        external_dependencies = {
             Spider: spider,
             Request: request,
             Response: response,
@@ -72,13 +72,11 @@ class InjectionMiddleware:
             Settings: spider.settings,
             StatsCollector: spider.crawler.stats,
         }
-        assert set(dependencies.keys()) == _SCRAPY_PROVIDED_CLASSES
-
-        plan = utils.build_plan(callback)
-        provider_instances = utils.build_providers(dependencies)
+        assert set(external_dependencies.keys()) == _SCRAPY_PROVIDED_CLASSES
 
         # Build all instances declared as dependencies
-        instances = yield from utils.build_instances(plan, provider_instances)
+        plan = utils.build_plan(callback)
+        instances = yield from utils.build_instances(plan, external_dependencies)
 
         # Fill the callback arguments with the created instances
         for arg, value in plan.final_kwargs(instances).items():
--- a/scrapy_poet/middleware.py
+++ b/scrapy_poet/middleware.py
@@ -73,11 +73,12 @@ class InjectionMiddleware:
             StatsCollector: spider.crawler.stats,
         }
         assert set(dependencies.keys()) == _SCRAPY_PROVIDED_CLASSES
-        plan, provider_instances = utils.build_plan(callback, dependencies)
+
+        plan = utils.build_plan(callback)
+        provider_instances = utils.build_providers(dependencies)
 
         # Build all instances declared as dependencies
-        instances = yield from utils.build_instances(
-            plan.dependencies, provider_instances)
+        instances = yield from utils.build_instances(plan, provider_instances)
 
         # Fill the callback arguments with the created instances
         for arg, value in plan.final_kwargs(instances).items():
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -11,8 +11,6 @@ import typing
 from scrapy.http import Response
 from web_poet.page_inputs import ResponseData
 
-providers = set()
-
 
 class PageObjectInputProvider(abc.ABC):
     """This is an abstract class for describing Page Object Input Providers."""
@@ -57,10 +55,6 @@ class PageObjectInputProvider(abc.ABC):
     @abc.abstractmethod
     def __call__(self, provided_classes: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
         """This method is responsible for building Page Input dependencies."""
-
-    @classmethod
-    def register(cls):
-        providers.add(cls)
 
 
 class ResponseDataProvider(PageObjectInputProvider):
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -53,7 +53,7 @@ class PageObjectInputProvider(abc.ABC):
         """
 
     @abc.abstractmethod
-    def __call__(self, provided_classes: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
+    def __call__(self, to_provide: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
         """This method is responsible for building Page Input dependencies."""
 
 
@@ -66,7 +66,7 @@ class ResponseDataProvider(PageObjectInputProvider):
         """This class receives a Scrapy ``Response`` as a dependency."""
         self.response = response
 
-    def __call__(self, provided_classes: typing.Set[typing.Type]):
+    def __call__(self, to_provide: typing.Set[typing.Type]):
         """Builds a ``ResponseData`` instance using a Scrapy ``Response``."""
         return {
             ResponseData: ResponseData(
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -83,7 +83,7 @@ class ResponseDataProvider(PageObjectInputProvider):
         """This class receives a Scrapy ``Response`` as a dependency."""
         self.response = response
 
-    def __call__(self, *args, **kwargs):
+    def __call__(self, provided_classes: typing.Set[typing.Type]):
         """Builds a ``ResponseData`` instance using a Scrapy ``Response``."""
         return {
             ResponseData: ResponseData(
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -9,6 +9,7 @@ import abc
 import typing
 
 from scrapy.http import Response
+from scrapy.crawler import Crawler
 from web_poet.page_inputs import ResponseData
 
 
@@ -17,7 +18,7 @@ class PageObjectInputProvider(abc.ABC):
 
     provided_classes: typing.Set[typing.Type]
 
-    def __init__(self):
+    def __init__(self, crawler: Crawler):
         """You can override this method to receive external dependencies.
 
         Currently, scrapy-poet is able to inject instances of the following
@@ -51,6 +52,7 @@ class PageObjectInputProvider(abc.ABC):
                         assert isinstance(response, TextResponse)
                         self.response = response
         """
+        self.crawler = crawler
 
     @abc.abstractmethod
     def __call__(self, to_provide: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
@@ -62,15 +64,11 @@ class ResponseDataProvider(PageObjectInputProvider):
 
     provided_classes = {ResponseData}
 
-    def __init__(self, response: Response):
-        """This class receives a Scrapy ``Response`` as a dependency."""
-        self.response = response
-
-    def __call__(self, to_provide: typing.Set[typing.Type]):
+    def __call__(self, to_provide: typing.Set[typing.Type], response: Response):
         """Builds a ``ResponseData`` instance using a Scrapy ``Response``."""
         return {
             ResponseData: ResponseData(
-                url=self.response.url,
-                html=self.response.text
+                url=response.url,
+                html=response.text
             )
         }
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -11,12 +11,13 @@ import typing
 from scrapy.http import Response
 from web_poet.page_inputs import ResponseData
 
-# FIXME: refactor _providers / provides / register,  make a nicer API
-providers = {}
+providers = set()
 
 
 class PageObjectInputProvider(abc.ABC):
     """This is an abstract class for describing Page Object Input Providers."""
+
+    provided_classes: typing.Set[typing.Type]
 
     def __init__(self):
         """You can override this method to receive external dependencies.
@@ -54,23 +55,11 @@ class PageObjectInputProvider(abc.ABC):
         """
 
     @abc.abstractmethod
-    def __call__(self):
+    def __call__(self, provided_classes: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
         """This method is responsible for building Page Input dependencies."""
 
 
-def register(provider_class: typing.Type[PageObjectInputProvider],
-             provided_class: typing.Type):
-    """This method registers a Page Object Input Provider in the providers
-    registry. It could be replaced by the use of the ``provides`` decorator.
-
-    Example:
-
-        register(ResponseDataProvider, ResponseData)
-    """
-    providers[provided_class] = provider_class
-
-
-def provides(provided_class: typing.Type):
+def provides():
     """This decorator should be used with classes that inherits from
     ``PageObjectInputProvider`` in order to automatically register them as
     providers.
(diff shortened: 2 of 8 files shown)


scrapy_poet/page_input_providers.py
```diff
@@ -54,49 +55,39 @@ def __init__(self, response: Response):
...
     providers.
 
     See ``ResponseDataProvider``'s implementation for an example.
     """
     def decorator(provider_class: typing.Type[PageObjectInputProvider]):
-        register(provider_class, provided_class)
+        providers.add(provider_class)
         return provider_class
 
     return decorator
```
kmike, line 73: Returning decorator doesn't make sense now, one can add to providers and return class as-is. I'm not sure decorator interface makes sense with that, but that can be discussed & refactored separately.

scrapy_poet/page_input_providers.py
```diff
@@ -54,49 +55,39 @@ def __init__(self, response: Response):
...
 
-    def __call__(self):
-        """This method builds a ``ResponseData`` instance using a Scrapy
-        ``Response``.
-        """
-        return ResponseData(
-            url=self.response.url,
-            html=self.response.text
-        )
+    def __call__(self, *args, **kwargs):
```
kmike, line 86: It could be better to receive provided_classes explicitly, what do you think?

scrapy_poet/utils.py
```diff
@@ -102,14 +102,12 @@ def discover_callback_providers(callback):
...
-        externally_provided=providers.keys(),
+        externally_provided=set.union(*(p.provided_classes for p in providers)),
     )
     for obj, _ in plan:
-        provider = providers.get(obj)
-        if not provider:
-            continue
-
-        yield provider
+        for provider in providers:
```
kmike, line 108: With providers being a set an order is a bit arbitrary now. Does it make sense to change it to a list, to ensure priorities are explicit?

kmike:
Thanks @victor-torres! The approach looks fine. I haven't checked an implementation in detail. What do you think about merging it only after a provider for html in scrapy-autoextract is implemented, to ensure the API proposed here is flexible enough, and that the implementation works in real-world cases?

scrapy_poet/page_input_providers.py
victor-torres: That was both my initial thought and conclusion: we can deal with it in another pull request.

scrapy_poet/page_input_providers.py
victor-torres: Absolutely, there's a lot of room for improvement on this pr as I was just pushing things from my local machine 😂

victor-torres: @kmike, before committing this change, a couple of questions:

- [how] should I type annotate the `provided_classes` argument?
- should I assert `provided_classes == self.provided_classes` or `provided_classes == {ResponseData}`

Thinking here if it makes sense to have a class attribute named `provided_class` for single-object providers. Our injection middleware could check if wither attribute is defined and raise an error if more than one is defined. This could simplify things a little bit (or make them more complex, never know). What do you think?

scrapy_poet/utils.py
victor-torres: - sets are a good option because they enforce uniqueness 
- lists on the other hand could make it possible to play with providers' order but
  - they don't guarantee uniqueness, what could lead to problems (maybe)
  - we don't depend on providers' order at the moment, so we can probably postpone this task

victor-torres:
> Thanks @victor-torres! The approach looks fine. 

Thanks, @kmike. I've been working non-stop trying to push this code before EOD. Hope my implementation gives a good idea about @ivanprado's proposal.

> What do you think about merging it only after a provider for html in scrapy-autoextract is implemented, to ensure the API proposed here is flexible enough, and that the implementation works in real-world cases?

Sure, this is was my idea. This pull request was opened because:
- I needed to write this specification first before implementing the HTML Data provider using it
- you can give early feedback whilst I work on the other part of this task 😄

ivanprado (commented):
Thank you @victor-torres for the iteration :+1: !  I went through them and left some comments. To me the new interface makes sense, it seems round enough. I left a pair of mayor comments:

- Each provider must check remaining work in received `to_provide_remaining` and only create instances if they are in the list. In the same way, the loop invoking the providers should update the `to_provide_remaining` on each loop iteration. 
- Preserving order of providers in a list makes sense to me.

scrapy_poet/page_input_providers.py
victor-torres: >  Indeed I would probably change this name to something more specific like to_provide_remaining, which is what the set represents: what classes has not been yet provided.

I don't like naming it as `to_provide_remaining` because it gives the idea that the provider might be called multiple times and it's not what's intended. Actually, calling the provider multiple times could be the source of issues since we "think" we're calling it with every dependency but suddenly, we're not.

> If there are no classes in the to_provide_remaining that the provider can provide, then it should do nothing. Rerturn the empty dict. This logic is not present in this provider.

I like this proposal. I'll try to implement it. 

> The reason for that is that there can be redundant providers. More than one provider could be providing the same type (e.g. Html data). Once one provider has provided this type, the rest should not try to provide it.

I'm not sure it should be possible to register multiple providers for the same type because that could make things a little bit more complicated or cause confusion: which one is going to be executed?

scrapy_poet/utils.py
victor-torres: Not sure about your example. 

Isn't it better to override the original AutoExtract provider and register / install this variation instead? I suppose it would lead to less bugs and would give better accountability for which providers are responsible for what.

scrapy_poet/page_input_providers.py
```diff
@@ -54,49 +55,39 @@ def __init__(self, response: Response):
         """
 
     @abc.abstractmethod
-    def __call__(self):
+    def __call__(self, provided_classes: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
```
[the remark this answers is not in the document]
victor-torres: It's possible, but I think the way it's implemented now is more convenient.

scrapy_poet/utils.py
```diff
@@ -125,40 +123,56 @@ def is_response_going_to_be_used(request, spider):
...
-def build_instances(plan: andi.Plan, providers):
+def build_instances(plan: andi.Plan, provider_instances: Set[PageObjectInputProvider]):
     """Build the instances dict from a plan."""
     instances = {}
-    for cls, kwargs_spec in plan:
-        if cls in providers:
-            instances[cls] = yield maybeDeferred_coro(providers[cls])
-        else:
+
+    dependencies = dict(plan.dependencies)
```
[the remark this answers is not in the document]
victor-torres: This is what the previous code was using. Any reason for changing that?

scrapy_poet/utils.py
```diff
@@ -125,40 +123,56 @@ def is_response_going_to_be_used(request, spider):
...
+        if provided_classes:
+            # Create a single instance of this provider and associate it
+            # with all classes being currently provided by it
+            instance = yield maybeDeferred_coro(provider, provided_classes)
+
+        for provided_class in provided_classes:
+            instances[provided_class] = instance[provided_class]
+
+    # Build missing dependencies not addressed by previous step
+    for cls, kwargs_spec in dependencies.items():
```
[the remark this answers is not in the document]
victor-torres: Will it make any difference here in this case? 🤔

scrapy_poet/utils.py
```diff
@@ -102,14 +102,12 @@ def discover_callback_providers(callback):
...
     )
     for obj, _ in plan:
-        provider = providers.get(obj)
-        if not provider:
-            continue
-
-        yield provider
+        for provider in providers:
+            if obj in provider.provided_classes:
+                yield provider
```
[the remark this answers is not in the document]
victor-torres: Done. It's returning a set instead of a generator now.

scrapy_poet/utils.py
```diff
@@ -125,40 +123,56 @@ def is_response_going_to_be_used(request, spider):
...
     """Build the instances dict from a plan."""
     instances = {}
-    for cls, kwargs_spec in plan:
-        if cls in providers:
-            instances[cls] = yield maybeDeferred_coro(providers[cls])
-        else:
+
+    dependencies = dict(plan.dependencies)
+
+    for provider in provider_instances:
```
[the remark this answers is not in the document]
victor-torres: Perhaps that was my first try. I'm afraid you cannot change a list/dictionary while iterating over it. 

I've just performed other improvements on this function, please check if they make sense.

scrapy_poet/utils.py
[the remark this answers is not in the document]
victor-torres: I've just tried to change it from `plan.dependencies` to `plan` and lots of test cases broke. AFAIK test cases are describing the correct functionality of the source code. Let me know if there's a reason for changing it and then we need to update our tests cases as well.

scrapy_poet/utils.py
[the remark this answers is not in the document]
victor-torres: FYI I'm using the `plan.dependencies` list now because of a refactor to simplify this code.

scrapy_poet/page_input_providers.py
victor-torres: > If there are no classes in the to_provide_remaining that the provider can provide, then it should do nothing. Rerturn the empty dict. This logic is not present in this provider.

I was going to implement this suggestion, but in the end I don't think it's necessary because the injection middleware (and its utils) take care of this use case.

`build_instances` method will only call a provider instance if a dependency is listed on its provided classes attribute, otherwise, it will just skip this provider. So no problem here and no need to handle this case in the provider itself because it won't ever be called.

scrapy_poet/utils.py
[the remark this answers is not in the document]
victor-torres: I'm back to dictionaries because of another refactor. It made sense just to use them. 

Regarding your concern with the items order, since Python 3.6 dicts keep the insertion order. 

```python
>>> dict([(1, 1), (2, 2), (3, 3)])
{1: 1, 2: 2, 3: 3}
>>> dict([(2, 2), (1, 1), (3, 3)])
{2: 2, 1: 1, 3: 3}
``` 

This project supports Python 3.6 onwards.

scrapy_poet/page_input_providers.py
```diff
@@ -54,49 +55,28 @@ def __init__(self, response: Response):
...
-            html=self.response.text
-        )
+    def __call__(self, provided_classes: typing.Set[typing.Type]):
+        """Builds a ``ResponseData`` instance using a Scrapy ``Response``."""
+        return {
+            ResponseData: ResponseData(
+                url=self.response.url,
+                html=self.response.text
+            )
+        }
```
victor-torres, line 82: You may think that we should:

- check if `provided_classes` are contained in `self.provided_classes` but our injection middleware already takes care of this so this method shouldn't be called with classes that it does not provide
- return a blank dictionary if there's no class being specified but again, our injection middleware already takes care of this so no provider should be initialized nor called if it doesn't provide any of the required types

victor-torres:
I think this PR is ready for another round of reviews @ivanprado @kmike @ejulio

scrapy_poet/page_input_providers.py
ivanprado: > build_instances method will only call a provider instance if a dependency is listed on its provided classes attribute, otherwise, it will just skip this provider. So no problem here and no need to handle this case in the provider itself because it won't ever be called.

An example where this won't be true:

```py 
class ProductProvider(PageObjectInputProvider):
  provided_classes = [Product, Html]

  def _call_(...):
    ...

class ArticleProvider(PageObjectInputProvider):
  provided_classes = [Article, Html]

  def _call_(...):
    ...
```

And then you have a callback in a spider:
```py
class MySpider(Spider):
  def parse(response, product: Product, article: Article, html: Html):
    pass
```

In such a case, `build_instances` will use both providers, and both providers will ask for Html. This is not right, only the first provider (whichever it is) should ask for the Html in this case. And for that you need the `to_provide_remaining` mechanism. 

There are other cases: overriding a provider. Imagine that you have a single provider that provides all the AutoExtract page objects:
```py
class AutoExtractProvider(PageObjectInputProvider):
  provided_classes = [Article, Product, JobPosting, Html]

  ....
```
But you want to override the provider for Product, because you are obtaining the product by other means. The overriding can be obtained just but placing your new provider before the original provider. But you need the `to_provide_remaining` mechanism to avoid both providers performing the request. 

IMO this mechanism is important thus. 

> I'm not sure it should be possible to register multiple providers for the same type because that could make things a little bit more complicated or cause confusion: which one is going to be executed?

I don't think is complicated. Simply the order will be respected. Is the same than with Scrapy extensions. Maybe we could use numbers like in Scrapy to make simpler the ordering. This can be done in a different PR, but in this one it think at least the ordering should be respected. 

Also note that not having order is even worse in terms of confusion, because in this case you'll never know which one will be executed. So results won't be necessary deterministic.

scrapy_poet/utils.py
ivanprado: @victor-torres there is an example above https://github.com/scrapinghub/scrapy-poet/pull/40#discussion_r492556338 . 

An alternative would make providers disjoint. That is, you don't allow having two providers providing the same class, and if this happens you raise an error. I think this is what you are thinking on. And then ordering is not important because is impossible for two providers to provide objects of the same class. 

So let's go back to an example I put before. In the case you have a provider like:
```py
class AutoExtractProvider(PageObjectInputProvider):
  provided_classes = [Article, Product, JobPosting, Html]

  def __init__(request: DummyRequest, stats: StatsCollector)

  def __call__(...):
    ....
```

If you want to override let's say product, you would have to write your own provider:

```py
class OverridenAutoExtractProvider(PageObjectInputProvider):
  provided_classes = AutoExtractProvider.provided_classes

  # Note that you have to mimic the annotations from the original provider so that the injection works. 
  def __init__(request: DummyRequest, stats: StatsCollector):
    self.origina_provider = AutoExtractProvider(request, stats)
    ...

  def __call__(to_provide)
    # here you have to provide product using your own custom code, and then the rest calling to
    # self.original_provider.__call__, put all together, and return it. 
  ....
```
And then replace the old one with the new one. 

I see now better the idea :+1: . Still not sure about it. I don't see the difficulties of having a ordered list of providers and keep the `to_provide_remaining` list. And it would make the overriding simpler.

scrapy_poet/utils.py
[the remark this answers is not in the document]
ivanprado: But note that it didn't become a Python standard until Python 3.7, so different 3.6 CPython implementations could have not implemented them. So I think better not to rely on that, given that is easy to iterate over the list of tuples in any case.

scrapy_poet/utils.py
```diff
@@ -125,40 +131,47 @@ def is_response_going_to_be_used(request, spider):
...
+
+    # Build dependencies handled by registered providers
+    plan_dependencies = dict(plan.dependencies)
+    for provider in providers:
+        provided_classes = plan_dependencies.keys() & provider.provided_classes
+        if not provided_classes:
+            continue
+
+        provider_instance = build_provider(provider, external_dependencies)
+        results = yield maybeDeferred_coro(provider_instance, provided_classes)
```
ivanprado, line 167: I see that you directly send in `provided_classes` these classes that the provider must deal with. Makes sense :+1:

scrapy_poet/utils.py
```diff
@@ -125,40 +131,47 @@ def is_response_going_to_be_used(request, spider):
...
+        instances.update(results)
+
+    # Build remaining dependencies
+    for cls, kwargs_spec in plan_dependencies.items():
```
ivanprado, line 171: First, my proposal about not using a dictionary because of the ordering concern would require just changing the line with:

```py
for cls, kwargs_spec in plan.dependencies:
```

Second, note that in the original code this line was:
```py
    for cls, kwargs_spec in plan:
```

I'm not saying this is wrong, but it is not the same iterating over the `plan` than iterating over the `plan.dependencies` (See the doc here: https://github.com/scrapinghub/andi/blob/master/andi/andi.py#L115, `plan.dependecies` is the plan without the last element. So I'm just trying to understand why here we are changing that. @victor-torres you have it clear?

tests/conftest.py
```diff
@@ -1,6 +1,10 @@
 import pytest
 from scrapy.settings import Settings
 
+from scrapy_poet.page_input_providers import ResponseDataProvider
+
+ResponseDataProvider.register()
```
ivanprado, line 6: Don't remember this well. Didn't we have a method `install()` somewhere where all the providers where installed?

I'm a little bit concerned about having a static list for the providers. I have the feeling that is better to have it somewhere else (e.g. the middleware), and then have in the modules some kind of installation function that can be used to fill it.

ivanprado: Ohh, I see this is the test.

ivanprado (commented):
@victor-torres I went trhough the changes and left some comments. 

It seems that the major discussion now is whether we should enforce providers to be disjoint  by design - then the order is unimportant because overlapping is impossible -  or instead allow overlapping wich would open the door to providers overriding if the order of the providers is considered. 

In terms of implementation, almost nothing changes:

- For the disjoint proposal, just raising an exception if overlapping happens
- For the overriding proposal, just change a line in the code to update the `to_provide` set that the `__call__` method receives. 

So this is just a pure design question. 

I like slightly more the option that allows overrides, because It allows to do all the things you can do in the disjoint case and it also allows to override without introduce complexity (API would be exactly the same).

@kmike what is your opinion on the matter?

scrapy_poet/page_input_providers.py
ivanprado: I finished looking at the code and I see that `__call__` receives only the list of classes that the provider must take care of. Makes sense. Implementing then the proposal that allows overriding is just a matter of removing from this list those for which you already have an instance. Just modifying one line in the code. And `to_provide` is then a much better name :-)

scrapy_poet/utils.py
victor-torres: 🤦🏻 on the master branch the plan argument received by this function is actually plan.dependencies, look https://github.com/scrapinghub/scrapy-poet/blob/master/scrapy_poet/middleware.py#L80

scrapy_poet/page_input_providers.py
victor-torres: - renamed `provided_classes` argument to `to_provided` 
- respecting order of providers

victor-torres:
@ivanprado, I'm now respecting the order of providers and our injection middleware should not ask for a provider to fulfill a type that was already fulfilled by another provider (see https://github.com/scrapinghub/scrapy-poet/pull/40/commits/b7b7ee4fdb196f0b1a905a90c3d8be6483e7a9db).

scrapy_poet/middleware.py
```diff
@@ -64,20 +64,21 @@ def process_response(self, request: Request, response: Response,
         """
         # Find out the dependencies
         callback = utils.get_callback(request, spider)
-        dependencies = {
+        external_dependencies = {
```
ivanprado, line 67: There are two types of external dependencies:

- The external dependencies required to build the providers
- The external dependencies required to invoke the callback (which are provided by the providers). 

The `external_dependencies` name is ambiguous and confused me when reading the code. It is now passed as argument to some methods and was thinking about which kind of dependencies they were. 

I propose then to change the name to make this more clear. What for example `provider_dependencies`? `provider_external_dependencies` would probably be the most accurate, but too long I guess. Anyway, this is very minor.

scrapy_poet/utils.py
```diff
@@ -118,47 +120,57 @@ def is_response_going_to_be_used(request, spider):
...
-def build_plan(callback, instances) -> Tuple[andi.Plan, Dict[Type, PageObjectInputProvider]]:
+def build_plan(callback: Callable, providers: List[Type[PageObjectInputProvider]]) -> andi.Plan:
     """Build a plan for the injection in the callback."""
-    provider_instances = build_providers(instances)
-    plan = andi.plan(
+    return andi.plan(
         callback,
         is_injectable=is_injectable,
-        externally_provided=provider_instances.keys()
+        externally_provided=get_provided_classes_from_providers(providers),
```
ivanprado, line 136: Initializing the providers from settings or getting the provided classes is something that is performed on each request and several times for each. It would be nice to cache these values so that they are initialized once at the start and then reusing. 

I propose to encapsulate all these functions (`build_plan`, `discovery_callback_providers`, maybe some of them present on the `InjectionMiddleware` itself, etc) into a class (e.g. `Injector`). This class is initialized in the Injection middleware. On initialization, it gets the settings, and caches some values (like the provider list and the list from `get_provided_classes_from_providers(providers)`) so that it can be reused.

scrapy_poet/middleware.py
kmike: data point: I only understood what dependencies vs external_dependencies are after a few hours of code review, and forgot about it a few hours later :)

scrapy_poet/utils.py
```diff
@@ -118,47 +120,57 @@ def is_response_going_to_be_used(request, spider):
...
+    dependencies_set = set(cls for cls, kwargs_spec in plan.dependencies)
+    for provider in providers:
+        provided_classes = dependencies_set & provider.provided_classes
+        provided_classes -= instances.keys()  # ignore already provided types
+        if not provided_classes:
+            continue
+
+        provider_instance = build_provider(provider, external_dependencies)
+        results = yield maybeDeferred_coro(provider_instance, provided_classes)
+        instances.update(results)
```
ivanprado, line 167: Wonder if we should check that `results.keys() - provided_classes == {}` and raise an exception otherwise. This will help if somebody writes a wrong provider returning the wrong object in the dict keys.

tests/test_utils.py
```diff
@@ -3,15 +3,16 @@
 
 import scrapy
 from scrapy.http import TextResponse
+from scrapy.settings import Settings
```
ivanprado, line 6: It would be nice to have a test that checks that overriding on providers works. Something like having two providers with some overlapping class, and then provide them in a different order and be sure that the response comes from the proper provider.

ivanprado (commented):
Thank you @victor-torres, great work!. I get got a good aggrement about the API :smiley: . I left some comments but are more related with code organization/optimizations. Major ideas:
* Create a class to encapsulate all the injection logic (e.g. `Injector`). This will allow to improve efficiency by caching some recurrent information and also encapsulate better the code. Now, most of the logic is in `utils` which I think it shouldn't. Utils to me is something like auxiliary functions not directly related with the main logic, but actually the main logic in this case is in the `utils` module. 
* Get providers from settings as Scrapy does for extensions and middlewares. 
* I wonder how we would do when adding a module to a spider (e.g. `scrapy-autoextract`). Having to manually add the providers from the module into the list of settings seems uncomfortable. Could we have some kind function wich updates the providers list all together? Or maybe is Ok given that usually we'll add a single provider which handle many types?
* Add some new unit tests to check providers overriding and new functionalities included.

tests/conftest.py
ivanprado: No static list of providers anymore. Marking as fixed.

scrapy_poet/utils.py
ivanprado: Ohh, I see now. Thank you :smiley:

scrapy_poet/utils.py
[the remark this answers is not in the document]
ivanprado: Fixed. Marking as resolved.

scrapy_poet/utils.py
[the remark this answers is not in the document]
ivanprado: This was addressed. Resolving.

scrapy_poet/utils.py
```diff
@@ -125,40 +123,56 @@ def is_response_going_to_be_used(request, spider):
...
     )
-    return plan, provider_instances
 
 
-def build_providers(instances) -> Dict[Type, PageObjectInputProvider]:
-    result = {}
-    for cls, provider in providers.items():
+def build_providers(instances) -> Set[PageObjectInputProvider]:
+    result = set()
+    for provider in providers:
```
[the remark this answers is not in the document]
ivanprado: Addressed.

scrapy_poet/page_input_providers.py
ivanprado: Addressed. Resolving.

scrapy_poet/middleware.py
victor-torres: Agree with you guys, just couldn't think about a better name.

ivanprado: Thinking on more names: `externally_available_for_providers` and `externally_available_for_callbacks` Quite long, uhh!

`external_types_for_providers` and `external_types_for_callbacks` are shorter. 

`supported_init_types_for_providers` and `types_supplied_by_providers`

scrapy_poet/utils.py
victor-torres: I would go further and raise an exception whenever the returned class set is different than the set the provider was asked. What do you think?

ivanprado: Yes, to me it makes sense. Will help the developer to spot problems. I don't see the case where returning something different in the keys would make sense.

victor-torres: > I would go further and raise an exception whenever the returned class set is different than the set the provider was asked. What do you think?

If we do that we will loose a powerful behavior in my opinion which is:

- imagine a SplashScreenshotProvider
- this provider provides SplashScreenshot and ResponseData
- it has priority over the ResponseDataProvider

Now let's think in two possible scenarios:

- callback needs SplashScreenshot and ResponseData
- callback only needs ResponseData

If callback needs both, okay, we can proceed and build the two objects as usual.

If callback only needs ResponseData, our SplashScreenshotProvider could then return a blank set of instances. This will force our injection middleware to ask for the next provider that's able to provide this information, that in our case is the ResponseDataProvider.

What do you think about it?

scrapy_poet/middleware.py
victor-torres: Why not `scrapy_objects` then?

scrapy_poet/utils.py
ivanprado: Yeah, this makes sense. Good example :+1: . But this is not incompatible with the idea I had in mind: the error should only be raised if any of the keys in the returned `dict` is not in `Provider.provided_clases`. An empty `dict` will always pass this check.

scrapy_poet/middleware.py
ivanprado: Yeah, sounds good.

scrapy_poet/utils.py
victor-torres: Cool. Raising an exception whenever a provider returns classes that it was not supposed to provide.

scrapy_poet/middleware.py
```diff
@@ -11,7 +10,47 @@
 from twisted.internet.defer import inlineCallbacks, returnValue
 
 from scrapy_poet import utils
-from scrapy_poet.utils import _SCRAPY_PROVIDED_CLASSES
+from scrapy_poet.page_input_providers import PageObjectInputProvider
+from scrapy_poet.utils import _SCRAPY_PROVIDED_CLASSES, load_provider_classes
+
+
+Provider = PageObjectInputProvider
```
victor-torres, line 17: This is not being used.

victor-torres: We could take the chance to rename this class into something more simple, like the assign statement in this line.

scrapy_poet/utils.py
```diff
@@ -29,6 +26,22 @@
 }
 
 
+def load_provider_classes(settings: Settings):
+    result = []
+    for cls in settings.getlist('SCRAPY_POET_PROVIDER_CLASSES'):
+        if not callable(cls):
```
ejulio, line 32: I thing you check for `str` here, not `callable`

scrapy_poet/utils.py
```diff
@@ -87,78 +100,89 @@ def is_callback_using_response(callback: Callable):
...
-        externally_provided=providers.keys(),
+        externally_provided=get_provided_classes_from_providers(providers),
     )
+    result = set()
     for obj, _ in plan:
-        provider = providers.get(obj)
-        if not provider:
-            continue
+        for provider in providers:
+            if obj in provider.provided_classes:
```
ejulio, line 123: `cls` instead of `obj`

scrapy_poet/utils.py
```diff
@@ -98,18 +99,19 @@ def is_provider_using_response(provider):
...
-def discover_callback_providers(callback):
+def discover_callback_providers(callback: Callable, providers: List[Type[PageObjectInputProvider]]):
     plan = andi.plan(
         callback,
         is_injectable=is_injectable,
-        externally_provided=providers.keys(),
+        externally_provided=get_provided_classes_from_providers(providers),
     )
+    result = set()
     for obj, _ in plan:
```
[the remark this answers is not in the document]
ejulio: Agree with this one.
Also, by changing to dict (default `OrderedDict`) we get the uniqueness and priority that was discussed in the other thread https://github.com/scrapinghub/scrapy-poet/pull/40/files#r491110196

ejulio (changes requested):

ivanprado:
Closing in favour of https://github.com/scrapinghub/scrapy-poet/pull/41

PR #41, opened 2020-11-10
[WIP] Even newer providers interface

ivanprado:
This PR continues with the great work done by @victor-torres in the PR https://github.com/scrapinghub/scrapy-poet/pull/40. 

It introduces an entirely new providers interface:
- One provider can provide instances for several classes
- The list of providers is configured in Scrapy settings
- Providers are initialized at the spider start. So there exist only a singleton instance per provider during the full lifecycle of the spider. This reduces the creation/destruction of objects and also simplifies providers that require a state.

- [ ] Refactoring code to create some kind of `Injector`. Code will be cleaner and caching of some information will make the code more efficient
- [ ] Update documentation
- [ ] Make sure that examples are right

### DIFF
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -11,8 +11,6 @@ import typing
 from scrapy.http import Response
 from web_poet.page_inputs import ResponseData
 
-providers = set()
-
 
 class PageObjectInputProvider(abc.ABC):
     """This is an abstract class for describing Page Object Input Providers."""
@@ -57,10 +55,6 @@ class PageObjectInputProvider(abc.ABC):
     @abc.abstractmethod
     def __call__(self, provided_classes: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
         """This method is responsible for building Page Input dependencies."""
-
-    @classmethod
-    def register(cls):
-        providers.add(cls)
 
 
 class ResponseDataProvider(PageObjectInputProvider):
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -53,7 +53,7 @@ class PageObjectInputProvider(abc.ABC):
         """
 
     @abc.abstractmethod
-    def __call__(self, provided_classes: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
+    def __call__(self, to_provide: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
         """This method is responsible for building Page Input dependencies."""
 
 
@@ -66,7 +66,7 @@ class ResponseDataProvider(PageObjectInputProvider):
         """This class receives a Scrapy ``Response`` as a dependency."""
         self.response = response
 
-    def __call__(self, provided_classes: typing.Set[typing.Type]):
+    def __call__(self, to_provide: typing.Set[typing.Type]):
         """Builds a ``ResponseData`` instance using a Scrapy ``Response``."""
         return {
             ResponseData: ResponseData(
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -83,7 +83,7 @@ class ResponseDataProvider(PageObjectInputProvider):
         """This class receives a Scrapy ``Response`` as a dependency."""
         self.response = response
 
-    def __call__(self, *args, **kwargs):
+    def __call__(self, provided_classes: typing.Set[typing.Type]):
         """Builds a ``ResponseData`` instance using a Scrapy ``Response``."""
         return {
             ResponseData: ResponseData(
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -9,6 +9,7 @@ import abc
 import typing
 
 from scrapy.http import Response
+from scrapy.crawler import Crawler
 from web_poet.page_inputs import ResponseData
 
 
@@ -17,7 +18,7 @@ class PageObjectInputProvider(abc.ABC):
 
     provided_classes: typing.Set[typing.Type]
 
-    def __init__(self):
+    def __init__(self, crawler: Crawler):
         """You can override this method to receive external dependencies.
 
         Currently, scrapy-poet is able to inject instances of the following
@@ -51,6 +52,7 @@ class PageObjectInputProvider(abc.ABC):
                         assert isinstance(response, TextResponse)
                         self.response = response
         """
+        self.crawler = crawler
 
     @abc.abstractmethod
     def __call__(self, to_provide: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
@@ -62,15 +64,11 @@ class ResponseDataProvider(PageObjectInputProvider):
 
     provided_classes = {ResponseData}
 
-    def __init__(self, response: Response):
-        """This class receives a Scrapy ``Response`` as a dependency."""
-        self.response = response
-
-    def __call__(self, to_provide: typing.Set[typing.Type]):
+    def __call__(self, to_provide: typing.Set[typing.Type], response: Response):
         """Builds a ``ResponseData`` instance using a Scrapy ``Response``."""
         return {
             ResponseData: ResponseData(
-                url=self.response.url,
-                html=self.response.text
+                url=response.url,
+                html=response.text
             )
         }
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -11,12 +11,13 @@ import typing
 from scrapy.http import Response
 from web_poet.page_inputs import ResponseData
 
-# FIXME: refactor _providers / provides / register,  make a nicer API
-providers = {}
+providers = set()
 
 
 class PageObjectInputProvider(abc.ABC):
     """This is an abstract class for describing Page Object Input Providers."""
+
+    provided_classes: typing.Set[typing.Type]
 
     def __init__(self):
         """You can override this method to receive external dependencies.
@@ -54,23 +55,11 @@ class PageObjectInputProvider(abc.ABC):
         """
 
     @abc.abstractmethod
-    def __call__(self):
+    def __call__(self, provided_classes: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
         """This method is responsible for building Page Input dependencies."""
 
 
-def register(provider_class: typing.Type[PageObjectInputProvider],
-             provided_class: typing.Type):
-    """This method registers a Page Object Input Provider in the providers
-    registry. It could be replaced by the use of the ``provides`` decorator.
-
-    Example:
-
-        register(ResponseDataProvider, ResponseData)
-    """
-    providers[provided_class] = provider_class
-
-
-def provides(provided_class: typing.Type):
+def provides():
     """This decorator should be used with classes that inherits from
     ``PageObjectInputProvider`` in order to automatically register them as
     providers.
@@ -78,25 +67,27 @@ def provides(provided_class: typing.Type):
     See ``ResponseDataProvider``'s implementation for an example.
     """
     def decorator(provider_class: typing.Type[PageObjectInputProvider]):
-        register(provider_class, provided_class)
+        providers.add(provider_class)
         return provider_class
 
     return decorator
 
 
-@provides(ResponseData)
+@provides()
 class ResponseDataProvider(PageObjectInputProvider):
     """This class provides ``web_poet.page_inputs.ResponseData`` instances."""
+
+    provided_classes = {ResponseData}
 
     def __init__(self, response: Response):
         """This class receives a Scrapy ``Response`` as a dependency."""
         self.response = response
 
-    def __call__(self):
-        """This method builds a ``ResponseData`` instance using a Scrapy
-        ``Response``.
-        """
-        return ResponseData(
-            url=self.response.url,
-            html=self.response.text
-        )
+    def __call__(self, *args, **kwargs):
+        """Builds a ``ResponseData`` instance using a Scrapy ``Response``."""
+        return {
+            ResponseData: ResponseData(
+                url=self.response.url,
+                html=self.response.text
+            )
+        }
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -5,58 +5,103 @@ repository of ``PageObjectInputProviders``.
 You could implement different providers in order to acquire data from multiple
 external sources, for example, Splash or Auto Extract API.
 """
-import abc
-import typing
+from typing import Set, Type, Union, Callable
 
 from scrapy.http import Response
 from scrapy.crawler import Crawler
+from scrapy_poet.errors import InjectionError
 from web_poet.page_inputs import ResponseData
 
 
-class PageObjectInputProvider(abc.ABC):
-    """This is an abstract class for describing Page Object Input Providers."""
+class PageObjectInputProvider:
+    """
+    This is the base class for creating Page Object Input Providers.
 
-    provided_classes: typing.Set[typing.Type]
+    A Page Object Input Provider (POIP) takes responsibility for providing
+    instances of some types to Scrapy callbacks. The types a POIP provide must
+    be declared in the class attribute ``provided_classes``.
+
+    POIPs are initialized at the spider start by invoking the ``__init__`` method,
+    which receives the crawler instance as argument.
+
+    The ``__call__`` method must be overridden, and it is inside this method
+    where the actual instances must be build. The default ``__call__`` signature
+    is as follows:
+
+    .. code-block:: python
+
+        def __call__(self, to_provide: Set[Type]) -> Dict[Type, Any]:
+
+    Therefore, it receives a list of types to be provided and return a dictionary
+    with the instances created indexed by type.
+
+    Additional dependencies can be declared in the ``__call__`` signature
+    that will be automatically injected. Currently, scrapy-poet is able
+    to inject instances of the following classes:
+
+    - :class:`~scrapy.http.Request`
+    - :class:`~scrapy.http.Response`
+    - :class:`~scrapy.crawler.Crawler`
+    - :class:`~scrapy.settings.Settings`
+    - :class:`~scrapy.statscollectors.StatsCollector`
+
+    Finally, ``__call__`` function can execute asynchronous code. Just
+    either prepend the declaration with ``async`` or annotate it with
+    ``@inlineCallbacks`` depending on how Scrapy ``TWISTED_REACTOR``
+    is configured.
+
+    The available POIPs are configured in the spider settings
+    ``SCRAPY_POET_PROVIDER_CLASSES``
+
+    A simple example of a provider:
+
+    .. code-block:: python
+
+        class BodyHtml(str): pass
+
+        class BodyHtmlProvider(PageObjectInputProvider):
+            provided_classes = {BodyHtml}
+
+            def __call__(self, to_provide, response: Response):
+                return {BodyHtml: BodyHtml(response.css("html body").get())}
+
+    The **provided_classes** class attribute is the ``set`` of classes
+    that this provider provides.
+    Alternatively, it can be a function with type ``Callable[[Callable], bool]`` that
+    returns ``True`` if and only if the given type, which must be callable,
+    is provided by this provider.
+    """
+
+    provided_classes: Union[Set[Callable], Callable[[Callable], bool]]
+
+    @classmethod
+    def is_provided(cls, type_: Callable):
+        """
+        Return ``True`` if the given type is provided by this provider, based
+        on the value of the attribute ``provided_classes``
+        """
+        if isinstance(cls.provided_classes, Set):
+            return type_ in cls.provided_classes
+        elif callable(cls.provided_classes):
+            return cls.provided_classes(type_)
+        else:
+            raise InjectionError(
+                f"Unexpected type '{type(cls.provided_classes)}' for "
+                f"'{cls}.provided_classes'. Expected either 'set' or 'callable'")
 
     def __init__(self, crawler: Crawler):
-        """You can override this method to receive external dependencies.
+        pass
 
-        Currently, scrapy-poet is able to inject instances of the following
-        classes as *provider* dependencies:
-
-        - :class:`~scrapy.http.Request`
-        - :class:`~scrapy.http.Response`
-        - :class:`~scrapy.crawler.Crawler`
-        - :class:`~scrapy.settings.Settings`
-        - :class:`~scrapy.statscollectors.StatsCollector`
-
-        .. warning::
-
-            Scrapy doesn't know when a Request is going to generate
-            a Response, a TextResponse, an HtmlResponse,
-            or any other type that inherits from Response.
-            Because of this,
-            you should always annotate your provider's response argument
-            with the Response type.
-            If your provider needs a TextResponse,
-            you need to validate it by yourself,
-            the same way you would need to do when using Scrapy callbacks.
-            Example:
-
-            .. code-block:: python
-
-                @provides(MyCustomResponseData)
-                class MyCustomResponseDataProvider(PageObjectInputProvider):
-
-                    def __init__(self, response: Response):
-                        assert isinstance(response, TextResponse)
-                        self.response = response
-        """
-        self.crawler = crawler
-
-    @abc.abstractmethod
-    def __call__(self, to_provide: typing.Set[typing.Type]) -> typing.Dict[typing.Type, typing.Any]:
-        """This method is responsible for building Page Input dependencies."""
+    # Remember that is expected for all children to implement the ``__call__``
+    # method. The simplest signature for it is:
+    #
+    #   def __call__(self, to_provide: Set[Type]) -> Dict[Type, Any]:
+    #
+    # But some adding some other injectable attributes are possible
+    # (see the class docstring)
+    #
+    # The technical reason why this method was not declared abstract is that
+    # injection breaks the method overriding rules and mypy then complains.
 
 
 class ResponseDataProvider(PageObjectInputProvider):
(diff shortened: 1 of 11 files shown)


sortafreel (approved):

scrapy_poet/injection.py
```diff
@@ -0,0 +1,323 @@
...
+    def __init__(self, crawler: Crawler):
+        self.crawler = crawler
+        self.spider = crawler.spider
+        self.load_providers()
+
+    def load_providers(self):
+        self.providers = [
+            cls(self.crawler)
+            for cls in load_provider_classes(self.spider.settings)
+        ]
```
kmike, line 40: I wonder if that'd be better to follow "scrapy way" here, and configure providers similarly to middlewares / pipelines / extensions. In Scrapy they're configured with `{cls: priority}` dicts in the user code, and there is also a base setting as well. So, you'd have

`SCRAPY_POET_PROVIDERS_BASE = {ResponseDataProvider: 500}`,

and then users can do `SCRAPY_POET_PROVIDERS = {MyProvider: 100, 'path.to.MyPovider2': 200}`.

The code to handle it is implemented here: 

* https://github.com/scrapy/scrapy/blob/85604e1078b927c2a875040e29c58b4594c8d386/scrapy/core/spidermw.py#L31
* https://github.com/scrapy/scrapy/blob/85604e1078b927c2a875040e29c58b4594c8d386/scrapy/utils/conf.py#L15
* https://github.com/scrapy/scrapy/blob/85604e1078b927c2a875040e29c58b4594c8d386/scrapy/middleware.py#L29 (note that create_instance is used there).

TBH, I don't have strong reasons to do this change, besides consistency.

scrapy_poet/injection.py
```diff
@@ -0,0 +1,323 @@
...
+                f"It must implement '__call__' method"
+            )
+
+
+def load_provider_classes(settings: Settings) -> List[PageObjectInputProvider]:
+    providers = settings.getlist('SCRAPY_POET_PROVIDER_CLASSES') or PROVIDERS
+    logger.info(f"Loading providers {providers}")
+    result = []
+    for cls in settings.getlist('SCRAPY_POET_PROVIDER_CLASSES') or PROVIDERS:
+        if not callable(cls):
```
kmike, line 180: FTR, this check is not needed since scrapy 2.4

example/example/autoextract.py
```diff
@@ -19,16 +19,13 @@ class AutoextractProductResponse:
...
+    provided_classes = {AutoextractProductResponse}
 
     @inlineCallbacks
-    def __call__(self):
-        data = (yield get_autoextract_product(self.response.url))
-        res = AutoextractProductResponse(data=data)
+    def __call__(self, to_provide, request: Request):
+        data = (yield get_autoextract_product(request.url))
+        res = {AutoextractProductResponse: AutoextractProductResponse(data=data)}
         raise returnValue(res)
```
kmike, line 29: returnValue shouldn't be needed here, I recall that was only needed for Python2 compatibility

```suggestion
        return {AutoextractProductResponse: AutoextractProductResponse(data=data)}
```

scrapy_poet/injection.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def load_providers(self):
+        self.providers = [
+            cls(self.crawler)
+            for cls in load_provider_classes(self.spider.settings)
+        ]
+        check_all_providers_are_callable(self.providers)
+        # Caching whether each provider requires the scrapy response
+        self.is_provider_requiring_scrapy_response = {
+            id(provider): is_provider_requiring_scrapy_response(provider)
```
kmike, line 44: Why is `id` call needed here?

kmike: If the idea is to avoid making a reference, https://docs.python.org/3/library/weakref.html#weakref.WeakKeyDictionary can be better.

scrapy_poet/injection.py
```diff
@@ -0,0 +1,323 @@
...
+    The attribute ``provided_classes`` from each provided is used.
+    This attribute can be a ``set`` or a ``Callable``. All sets are
+    joined together for efficiency.
+    """
+    sets_of_types: Set[Callable] = set()  # caching all sets found
+    individual_is_callable: List[Callable[[Callable], bool]] = [sets_of_types.__contains__]
+    for provider in providers:
+        provided_classes = provider.provided_classes
+
+        if isinstance(provided_classes, Set):
```
kmike, line 203: does it work with frozenset?

scrapy_poet/page_input_providers.py
```diff
@@ -1,102 +1,127 @@
...
+    - :class:`~scrapy.settings.Settings`
+    - :class:`~scrapy.statscollectors.StatsCollector`
 
-    @abc.abstractmethod
-    def __call__(self):
-        """This method is responsible for building Page Input dependencies."""
+    Finally, ``__call__`` function can execute asynchronous code. Just
+    either prepend the declaration with ``async`` or annotate it with
+    ``@inlineCallbacks`` depending on how Scrapy ``TWISTED_REACTOR``
+    is configured.
```
kmike, line 55: `async def` support shouldn't require setting asyncio reactor, see https://docs.scrapy.org/en/latest/topics/coroutines.html

kmike: If we declare `async def` support for providers, it'd be nice to have a test for it. Maybe modify one of the inlineCallbacks + returnValue test, or copy it? Or maybe there is such test already, and I missed it :)

tests/test_providers.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __init__(self, crawler: Crawler):
+        assert isinstance(crawler, Crawler)
+        super().__init__(crawler)
+
+    def __call__(self, to_provide, response: scrapy.http.Response, spider: scrapy.Spider):
+        assert isinstance(spider, scrapy.Spider)
+        ret: Dict[Type, Any] = {}
+        if Price in to_provide:
+            ret[Price] = response.css(".price::text").get()
```
kmike, line 56: hm, what's the use case of allowing to return a value of a non-matching type? Here for Price a str is returned. 

A related question: why do users need to write `return {cls: instance_of_cls}`, instead of `return [instance_of_cls]` - is it because of some use cases for returning not exactly the same type? What are these use cases? Sorry if I misssed that in the previous discussions.

scrapy_poet/injection.py
kmike: I'm asking because it is quiite natural to define Provider.povided_classes as frozenset - it is not supposed to be mutable.

scrapy_poet/injection.py
ivanprado: Keeping references is not a problem (the provider's instances are kept internally within the injector). The main reason is avoiding expensive operations if the provider implements an overridden `__eq__` (or even if it was wrongly implemented). For this particular case, it is known that there will be a single instance of each provider and it can be identified by the id well. Why would WeakKeyDictionary be better? 

In other words, for example, I want to avoid using the expensive comparator that `attr.s` could create for a provider if somebody is using it.

scrapy_poet/injection.py
ivanprado: Good point. Implemented in https://github.com/scrapinghub/scrapy-poet/pull/41/commits/d0432e3da5885de37c21131e763540bb879ff1e3

scrapy_poet/page_input_providers.py
ivanprado: TIL. I added the test case and clarified a little bit the documentation here: https://github.com/scrapinghub/scrapy-poet/pull/41/commits/adef92ed36287e57fb205dec8f140c8f0d0a0985

scrapy_poet/injection.py
kmike: For me `id` is always suspicious because Python can reuse ids between different object, and actually does so by default in many cases. So, an object is created, you take an id from it, then the object is deallocated (e.g. nobody refers to it anymore), a completely different object is created, and it may get the same id. You then do a lookup and get a wrong value, misteriously.

This is not a theoretical issue, as it happens by default for some objects (I think tuples? not sure) - Python is free to use preallocated pools, and reuse pointers of objects which no longer have references, to avoid extra memory allocation. 

I don't think this happens here, but the code just doesn't feel "correct" - it may lead to very hard-to-debug bugs. You take id(proider), which doesn't create a new reference to provider, so nothing prevents provider to be gone, and id to be reused. If you use provider itself as a key, this can't happen. It also can't happen with weakref.

ivanprado: I see your point. Sure that Python is reusing memory, at the very minimum as C does: If you call `malloc` after `free` you could get the same memory address. But - correct me if I'm wrong - this won't happen for a living object: once the object is alive, it will retain the same address. In fact, the default `__eq__` implementation for `object` just uses `id()` (again correct me if I'm wrong). 

>  You take id(proider), which doesn't create a new reference to provider, so nothing prevents provider to be gone

Well, it is impossible for the provider to be gone. Only if somebody externally modifies the internal list of providers in the `Injector`... but note that in this case, this will also fail with a `WeakKeyDictionary`: a newly introduced provider won't be present in the dictionary and the `Injector` will just fail. The current implementation doesn't allow for run-time updating of the providers by design. We could do it, but It doesn't seem required. 

So well, as conclusion, I think using `id()` is an optimization and in this case is not risky (I'm just forcing using `is` vs `==` for comparison), but it is also Ok to me to remove the `id` and just use the instance. Using `WeakKeyDictionary` doesn't provide any benefit in this case IMHO. Implemented in: https://github.com/scrapinghub/scrapy-poet/pull/41/commits/4fed87ae81f9e58e6aaa64b69bc28afce11ab352

kmike: Yeah, that's fair that for living objects id is fine. I guess my issue is that with `id(provider)` code becomes non-local: if you use provider as a key, it guarantees that the object will be alive (as a mapping now has a reference to the object); with id it is not guaranteed, so you you have to read the whole program to know if using `id(provider)` is valid or not here. 

It is like a very light https://en.wikipedia.org/wiki/COMEFROM :) in a sense global code structure needs to be checked to know what happens locally.

scrapy_poet/injection.py
ivanprado: Working on changing the settings, but one question, what do you think about the `from_crawler`? I can stick to the standard of using the `from_crawler` or just keeping as is now and receive always the `crawler` as `__init__` argument. I'm inclined to the latter because I see the `from_crawler` as boilerplate, but this would break somehow the standard. @kmike what do you think?

ivanprado: Also another question regarding `SCRAPY_POET_PROVIDERS_BASE`. This works well for Scrapy as you can include the default in [default_settings.py](https://github.com/scrapy/scrapy/blob/5e9cc3298be4ca1146d68098899365b334706db8/scrapy/settings/default_settings.py), but I don't think we have a good way to access to that from `scrapy-poet`. So I'll have a local default for this case with the default provider.

ivanprado: Implemented in https://github.com/scrapinghub/scrapy-poet/pull/41/commits/b176600581570bb9a6a5606b3d5cb7b206e88fa9

Finally decided not to follow `from_crawler` standard

scrapy_poet/injection.py
ivanprado: Removed

scrapy_poet/injection.py
ivanprado: TIL COMEFROM :smile:

tests/test_providers.py
ivanprado: I tried to trace back this decision. I found this reference: https://github.com/scrapinghub/scrapy-poet/pull/40#discussion_r491875341 that stated that this is still an open question. So let me philosophize about it. 

We have only two possibilities IMO:
1) Keep the current structure (but include a check that makes sure that the returned object is a subclass of the expected type for security). 
2) Force to always return instances of the expected type. In this case the return `dict` could be replaced by a `list`. 

Option 2. clearly makes the API cleaner, but disallow supporting instances that are a subclass of the expected type. So the question we have to answer is: **is the functionality of allowing returning subclasses of the expected type useful**. I'll focus on listing some examples where it could be useful. This is the full list of cases:

```py
[]
```

I can think of cases where subclassing the provider to modify its behaviour makes sense (e.g. subclassing the default `ReponseDataProvider` to apply ad-block rules to the HTML), but the returning instance would have the same type: it is enough with changing its data. 

But I don't see a clear case where the returning instance requires a change in its business logic, especially given that these objects are usually just data carriers. So I'm going to go for option 2 and simplify the API. @kmike what do you think?

kmike: This sounds good @ivanprado. I also can't think of a use case. 

Even if we find a use case in future, we'd be able to adjust API in a backwards compatible way, but we won't need to pay the complexity price in the meantime.

ivanprado: Providers returning a list of instances implemente here: https://github.com/scrapinghub/scrapy-poet/pull/41/commits/24b9627ff4ff0749c9f24bbf5ab6bc6ff37b88af

ivanprado:
@kmike I addressed all your comments. Ready for a re-review.

docs/advanced.rst
```diff
@@ -33,26 +33,39 @@ that need it, like the ``ItemWebPage``.
...
     class ResponseDataProvider(PageObjectInputProvider):
         """This class provides ``web_poet.page_inputs.ResponseData`` instances."""
+        provided_classes = {ResponseData}
 
-        def __init__(self, response: Response):
-            """This class receives a Scrapy ``Response`` as a dependency."""
-            self.response = response
+        def __call__(self, to_provide: Set[Callable], response: Response):
+            """Builds a ``ResponseData`` instance using a Scrapy ``Response``"""
+            return [ResponseData(url=response.url,html=response.text)]
```
kmike, line 42:
```suggestion
            """Build a ``ResponseData`` instance using a Scrapy ``Response``"""
            return [ResponseData(url=response.url, html=response.text)]
```

example/example/autoextract.py
```diff
@@ -19,17 +19,13 @@ class AutoextractProductResponse:
...
+    provided_classes = {AutoextractProductResponse}
 
     @inlineCallbacks
-    def __call__(self):
-        data = (yield get_autoextract_product(self.response.url))
-        res = AutoextractProductResponse(data=data)
-        raise returnValue(res)
+    def __call__(self, to_provide, request: Request):
+        data = (yield get_autoextract_product(request.url))
+        return {AutoextractProductResponse: AutoextractProductResponse(data=data)}
```
kmike, line 28:
```suggestion
        return [AutoextractProductResponse(data=data)]
```

scrapy_poet/__init__.py
```diff
@@ -1,2 +1,4 @@
-from .middleware import InjectionMiddleware
-from .utils import callback_for
+from .middleware import InjectionMiddleware, DEFAULT_PROVIDERS
```
kmike, line 1: Do we need to expose DEFAULT_PROVIDERS on a top level? What's the use case?

scrapy_poet/injection.py
```diff
@@ -0,0 +1,316 @@
...
+        instances = yield from self.build_instances_from_providers(
+            request, response, plan
+        )
+        # All the remaining dependencies are internal so they can be built just
+        # following the andi plan.
+        for cls, kwargs_spec in plan.dependencies:
+            if cls not in instances.keys():
+                instances[cls] = cls(**kwargs_spec.kwargs(instances))
+
+        raise returnValue(instances)
```
kmike, line 123:
```suggestion
        return instances
```

scrapy_poet/injection.py
```diff
@@ -0,0 +1,316 @@
...
+            objs_by_type: Dict[Callable, Any] = {type(obj): obj for obj in objs}
+            extra_classes = objs_by_type.keys() - provided_classes
+            if extra_classes:
+                raise UndeclaredProvidedTypeError(
+                    f"{provider} has returned {extra_classes} but they're not "
+                    f"listed as provided classes {provider.provided_classes}"
+                )
+            instances.update(objs_by_type)
+
+        raise returnValue(instances)
```
kmike, line 157:
```suggestion
        return instances
```

scrapy_poet/injection.py
```diff
@@ -0,0 +1,316 @@
...
+    def build_callback_dependencies(self, request: Request, response: Response):
+        """
+        Scan the configured callback for this request looking for the
+        dependencies and build the corresponding instances. Return a kwargs
+        dictionary with the built instances.
+        """
+        plan = self.build_plan(request)
+        provider_instances = yield from self.build_instances(request, response, plan)
+        callback_kwargs = plan.final_kwargs(provider_instances)
+        raise returnValue(callback_kwargs)
```
kmike, line 169:
```suggestion
        return plan.final_kwargs(provider_instances)
```

scrapy_poet/__init__.py
ivanprado: Maybe is too much? The unique use case is to be able to know which are the default providers. So maybe better to remove from the top level?

tests/test_middleware.py
```diff
@@ -106,46 +120,104 @@ def test_optional_and_unions(settings):
...
-    def __init__(self, response: scrapy.http.Response):
-        self.response = response
+    provided_classes = {ProvidedWithDeferred}
 
     @inlineCallbacks
-    def __call__(self):
+    def __call__(self, to_provide, response: scrapy.http.Response):
         five = yield deferToThread(lambda: 5)
-        raise returnValue(ProvidedAsyncTest(f"Provided {five}!", None))
+        raise returnValue([ProvidedWithDeferred(f"Provided {five}!", None)])
```
kmike, line 140:
```suggestion
        return [ProvidedWithDeferred(f"Provided {five}!", None)]
```

tests/test_response_required_logic.py
```diff
@@ -32,47 +32,50 @@ class FakeProductResponse:
...
             'product': {
-                'url': self.request.url,
+                'url': request.url,
                 'name': 'Sample',
             },
         }
-        return DummyProductResponse(data=data)
+        return {
+            DummyProductResponse: DummyProductResponse(data=data)
+        }
```
kmike, line 48:
```suggestion
        return [DummyProductResponse(data=data)]
```

Why weren't tests failing here?

tests/test_response_required_logic.py
```diff
@@ -32,47 +32,50 @@ class FakeProductResponse:
...
         data = {
             'product': {
                 'url': 'http://example.com/sample',
                 'name': 'Sample',
             },
         }
-        return DummyProductResponse(data=data)
+        return {
+            DummyProductResponse: DummyProductResponse(data=data)
+        }
```
kmike, line 64:
```suggestion
        return [DummyProductResponse(data=data)]
```

scrapy_poet/__init__.py
kmike: Yeah, users always can import those directly. Top-level shortcuts are there to simplify common tasks & suggest what's important in the API. 

I think users may be able to see default providers in logs, at spider startup - do you know if this happens?

kmike (approved):
Looks great, thanks @ivanprado for picking it up, refactoring & addressing all the comments! 
And thanks @victor-torres for the original implementation!

tests/test_response_required_logic.py
ivanprado: Because in this test the injection part is not tested, so this is not invoked. But It is not nice to have some code that is never called, so I'll add something to the tests. Sorry for missing all these changes and thank you for sporting them!

docs/advanced.rst
ivanprado: Always forgetting the imperative :upside_down_face:

scrapy_poet/injection.py
ivanprado: TIL

scrapy_poet/__init__.py
ivanprado: Removed here: https://github.com/scrapinghub/scrapy-poet/pull/41/commits/7ac3911a35db0572309d198bc5cbec17f61ff1af and improving logging here: https://github.com/scrapinghub/scrapy-poet/pull/41/commits/02f7b2b92a9ddea107e071815d20e58774a4ace1

ejulio (commented):
hey @ivanprado good job here :clap: 

I still have some concerns regarding the API.
Why do we have a priority instead of single registry entry?
```
SCRAPY_POET_PROVIDERS = {
    ToBeProvided: Provider
}
```

Also, to avoid repeating the default providers, we could provide a base value for this setting.
However, it needs to be somewhat extensible, so I'm fine with the idea of a registry function that we need to call explicitly.
```
scrapy_poet.register_defaults(SCRAPY_POET_PROVIDERS)
SCRAPY_POET_PROVIDERS,update({
   ToBeProvided: MyCustomProvider
})
```
This way we have control over the context and can decide on what is injected.

example/example/settings.py
```diff
@@ -6,83 +6,19 @@
...
 
 BOT_NAME = 'example'
 
 SPIDER_MODULES = ['example.spiders']
 NEWSPIDER_MODULE = 'example.spiders'
 
-
-# Crawl responsibly by identifying yourself (and your website) on the user-agent
-#USER_AGENT = 'example (+http://www.yourdomain.com)'
+SCRAPY_POET_PROVIDERS = {AutoextractProductProvider: 500}
```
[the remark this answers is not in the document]
ivanprado: The reason is here: https://github.com/scrapinghub/scrapy-poet/pull/41#discussion_r525548386 . Basically the idea is to be coherent with the Scrapy way. 

Still, priorities can be useful. So it is not that priorities are not required here and we are just including it artificially. An example where priorities are important: A default provider is providing all AutoExtract page inputs (products, articles, etc), but for your project, you want to replace/alter/modify the generated page input for products without altering the rest. By creating your provider and placing it with more priority, you will get the desired result.

The number of 500 is to put something between 0 and 999. This way there is room to put providers with higher or lower priority.

scrapy_poet/injection.py
```diff
@@ -0,0 +1,316 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Injector:
+    """
+    Keep all the logic required to do dependency injection in Scrapy callbacks.
+    Initializes the providers from the spider settings at initialization.
+    """
+    def __init__(self, crawler: Crawler, default_providers: Mapping = None):
```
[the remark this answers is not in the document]
ivanprado: I remember reading somewhere that `Optional[<type>]` is automatically inferred in these cases, but better to be explicit. Done here: https://github.com/scrapinghub/scrapy-poet/pull/41/commits/88b363d4b913d12bad91c6ac799f39db58531e30

scrapy_poet/injection.py
```diff
@@ -0,0 +1,316 @@
...
+    def discover_callback_providers(self, callback: Callable
+                                    ) -> Set[PageObjectInputProvider]:
+        """Discover the providers that are required to fulfil the callback dependencies"""
+        plan = andi.plan(
+            callback,
+            is_injectable=is_injectable,
+            externally_provided=self.is_class_provided_by_any_provider,
+        )
+        result = set()
+        for cls, _ in plan:
```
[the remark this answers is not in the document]
ivanprado: `provided_classes` can be a function, so it is not possible to just build a `dict` for speed up the check. This will be useful for implementing generic providers, for example, ones for `AutoExtract`.

```py
def provided_classes(cls):
  return issubclass(cls, AutoExtractPageInput)
```

I don't get your use of the word `priorities` here.

ivanprado:
Thank you @ejulio for having a look.

> Why do we have a priority instead of single registry entry?

Not sure to get your idea here. It seems the registry you propose is key=<provided_class>, value=<provider_class>. This has a major problem IMHO: What would it happen if you assign the wrong provided_class to a provider? It would just fail. I mean, each provider has a specific set of classes that can provide so there is no flexibility there. So it would only server for disabling some classes, but he already can decide not to use a class in its code. Or it can serve to prioritize different providers based on the provided class itself (if we consider that the dictionary is keeping the ordering, which is true for the supported Python version). But this seems a little bit overkilling. It is not required so fine-grained level for priorities. Also, this system doesn't allow to use a function to declare which are the provided classes, which can be very useful to write generic providers (e.g. a single provider for all AutoExtract page inputs). 

> Also, to avoid repeating the default providers, we could provide a base value for this setting.

There is a mechanism already in place for that. `DEFAULT_PROVIDERS` contains a list of providers are always included and if at some point you want to disable any of them, you can do by declaring it and assigning probability `None`. This is the relevant part of the code: https://github.com/scrapinghub/scrapy-poet/blob/new-provider-interface/scrapy_poet/middleware.py#L30

example/example/settings.py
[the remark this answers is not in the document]
ejulio: Yeah, but a list is sorted, so the ordering in the list solves the priority thing.
But I agree on keeping it coherent with scrapy

scrapy_poet/injection.py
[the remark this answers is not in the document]
ejulio: As we have priorities at hand, we could create a `dict` with the highest priority provider, so there wouldn't be a need to iterate over things.

scrapy_poet/__init__.py
kmike: Scrapy logs settings at startup, I was hoping this covers logging of the providers

ejulio:
We can keep the same interface with `Providers` listing the provided classes. But a registry function would map the provided class to the provider.
The priorities still apply here, the last one to be registered in the valid one, no need to keep the other ones around.

scrapy_poet/__init__.py
ivanprado: It doesn't seem to be the case.

example/example/settings.py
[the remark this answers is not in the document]
ivanprado: Yep, indeed my first implementation was a list. Finally, we changed it to the Scrapy way.

scrapy_poet/injection.py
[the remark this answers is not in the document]
ivanprado: Still, I don't get it. Could you please elaborate? The providers are already sorted by priority. But note that a provider could decide not to provide one of its classes listed, so we need the list of all of them, even if it has a lower priority. This is a rare feature, but it is as is implemented. Allows some providers to ignore providing some objects, for example, because of the URL. This allows creating a provider that overrides the default only for a domain for example, which can be useful in some situations.

ivanprado:
> We can keep the same interface with Providers listing the provided classes. But a registry function would map the provided class to the provider.
The priorities still apply here, the last one to be registered in the valid one, no need to keep the other ones around.

It seems more complex than the current approach to me. @ejulio what would be the advantages of that?

ivanprado:
Merging this PR. Thank you very much @victor-torres for the great foundation code for this PR :clap: . And also thank you for the reviewers @ejulio @kmike and @sortafreel :beers:

PR #43, opened 2021-02-09
Overrides support

ivanprado:
* Add support for overriding Page Objects and other Injectables/page inputs. 
* The default implementation allows configuring rules per domain.

This depends on a new release of `andi`


docs/overrides.rst
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+The last step is configuring ``scrapy-poet`` so that these Page Objects
+are used for Book Page URLs. This can be done by configuring
+``SCRAPY_POET_OVERRIDES`` in the Scrapy ``settings.py`` configuration file.
+
+.. code-block:: python
+
+    SCRAPY_POET_OVERRIDES = {
+        "bookpage.com": {
```
kmike, line 85: Would it make sense to have path-based configuration here as well? I.e. "use a different PO for this domain if path matches a regexp"?

docs/overrides.rst
```diff
@@ -0,0 +1,167 @@
...
+
+Let's reuse the very same ``BooksSpider`` spider but now to extract books from a different
+page: `bookpage.com/reviews <https://bookpage.com/reviews>`_.
+
+We cannot reuse the Books to Scrape Page Objects in this case, so we have
+to implement new ones:
+
+.. code-block:: python
+
+    class BPBookListPage(WebPage):
```
ejulio, line 64: I'd expect it to be a subclass of `BookListPage`.

Thinking that the crawler is generic, it should work with a common base interface and extensions should inherit from it

docs/overrides.rst
```diff
@@ -0,0 +1,167 @@
...
+
+The last step is configuring ``scrapy-poet`` so that these Page Objects
+are used for Book Page URLs. This can be done by configuring
+``SCRAPY_POET_OVERRIDES`` in the Scrapy ``settings.py`` configuration file.
+
+.. code-block:: python
+
+    SCRAPY_POET_OVERRIDES = {
+        "bookpage.com": {
+            BookListPage: BPBookListPage,
```
ejulio, line 86: If the subclass above makes sense, this mapping is not required, just need to list the new POs

docs/overrides.rst
```diff
@@ -0,0 +1,167 @@
...
+            BookPage: BPBookPage
+        }
+    }
+
+``SCRAPY_POET_OVERRIDES`` contain rules that overrides the Page Objects
+used for a particular domain. In this particular case, Page Objects
+``BPBookListPage`` and ``BPBookPage`` will be used instead of
+``BookListPage`` and ``BPBookPage`` for any request whose domain is
```
ejulio, line 94: `BPBookPage` -> `BookPage`

docs/overrides.rst
ejulio: Maybe an arbitrary function that gets a response and returns `True/False` indicating that the given overrides should be applied

scrapy_poet/utils.py
```diff
@@ -0,0 +1,15 @@
+from tldextract import tldextract
+
+
+def get_domain(url):
```
ejulio, line 4: It seems this will be called many times while crawling a website.
Maybe am alternative would be to use a parse cache
https://github.com/scrapy/scrapy/blob/e42d82526a1a519ff1305b189acde064a40fbd8d/scrapy/utils/httpobj.py#L10

docs/overrides.rst
ivanprado: That's one of the first things I thought, and my conclusion was that:
* Likely you can control that in the PO itself and it would be even cleaner. If the path is that, then use this extraction logic, etc. 
* If it would be really required a more elaborate ruling (e.g. using `request.meta` to apply overrides or no, or using a regex in the `path` to decide), you can always use a custom `SCRAPY_POET_OVERRIDES_REGISTRY`. 

Adding a regexp path to the config makes it less straightforward so I don't think it is worth it given that there are alternatives. What do you reckon?

tests/test_injection.py
```diff
@@ -250,6 +256,71 @@ def callback(response: DummyResponse,
...
+        item["currency"] = "$"
+        return item
+
+
+class TestInjectorOverrides:
+
+    @pytest.mark.parametrize("override_expected", [True, False])
+    @inlineCallbacks
+    def test_overrides(self, providers, override_expected):
+        overrides = {("other-example.com", "example.com")[override_expected]: {
```
ejulio, line 297: I think it is better to extract the common code in other function and have a clear setup and expectations, instead of having to build the `if` paths based on the input

ejulio (commented):
Great stuff @ivanprado :tada: 
I'm not yet sold on the overrides structure.
I'll put some thought on it and get back to you

docs/overrides.rst
ivanprado: Inheriting from `BookListPage` have the problem that you would be inheriting implementation details that might not want to. But I understand your point, so consider then the example where you have a base PO abstract class and then you have a concrete implementation for each case. You can see this implementation in the example `books_04_overrides_02.py` here: https://github.com/scrapinghub/scrapy-poet/pull/43/files#diff-855ff677d82fe0f56107cdaeb283941ff477930235ce9da6e9c16fbe5758fdf8

It also makes a lot of sense to me. Nothing prevents the user to follow this approach. Maybe the question is if we should also follow this approach in the documentation.

docs/overrides.rst
ivanprado: Autodetecting from the inheritance has its caveats. Every single class in hierarchy up to `Injectable` is a candidate to be the overridden class. How to select it? Using just the parent of a type? I think this is not enough, because you might use inheritance because of other reasons not related with overrides (e.g. creating a base class that contains some functionality but that still is abstract and needs requires create subclasses with the final implementation). 

Multiple inheritance could also be another problem. Imagine you define the following base classes: `Menu`, `Breadcrubms`, `Footer`. You might implement a PO `FooSitePO` that inherits for all of them, as it is providing all of them in one go. And you might be interested in overriding `Breadcrumbs` using `FooSitePo` but without being overriding `Footer` and `Menu`. With the current implementation, wich is explicit, this is not a problem.

Anyway, I wouldn't say I'm 100% sure we shouldn't go the detection from the inheritance way. Let's see if @kmike has some opinion here.

docs/overrides.rst
ivanprado: Thank you! Fixed here: https://github.com/scrapinghub/scrapy-poet/pull/43/commits/eab5e92e0bb50e1da4302c938ad7f5cac058ef6e

scrapy_poet/utils.py
ivanprado: Yes, I would say this would even be a thing to be moved to Scrapy itself. It is pretty useful to be able to extract the domain. Here is only called twice per request, so I don't think it is worth it.

tests/test_injection.py
ivanprado: I've tried to make it more clear here https://github.com/scrapinghub/scrapy-poet/pull/43/commits/222e950447159c2463a5aedae542519b2c419e7c

ivanprado:
@ejulio thank you for the review!. I answer your questions. I didn't get that one:

> Maybe an arbitrary function that gets a response and returns True/False indicating that the given overrides should be applied

What do you mean?

docs/overrides.rst
kmike: I think that'd be good to follow this approach in the tutorial, because having site-specific code in a generic spider makes the example less clear. What do you think about the following structure?

1. Present basic example, with a concrete, website-specific BookPages used in a spider - no changes needed  here, this is already in the tutorial.
2. Show how to refactor it to make the spider generic: empty default implementation, overrides for both websites.

ivanprado: @kmike it makes sense to me. It will make tutorial longer, but also useful as will explain which is the way to organize multi-site spiders.

ivanprado: @ejulio @kmike I've updated the documentation as @kmike suggested here https://github.com/scrapinghub/scrapy-poet/pull/43/commits/26277a86d4a43a4591e8d38ea6e5beaadbea9311

docs/overrides.rst
ejulio: This one is resolved for me, not sure about @kmike

docs/overrides.rst
ejulio: Resolved for me.
Not sure about you @kmike

ejulio:
@ivanprado 

> Maybe an arbitrary function that gets a response and returns True/False indicating that the given overrides should be applied

It's related to the routing strategy.
Instead of using the domain/regex, you could define a function that returns True/False to perform the routing.
However, it seems it is not required.

ejulio (approved):

ivanprado:
Thank you @ejulio for the review and the approval :beers: 

> Instead of using the domain/regex, you could define a function that returns True/False to perform the routing.

@ejulio is a good point. This is something to consider in the future. We might have pages where one PageObject or other should be used based on the content itself (e.g. because they use technology for comments that are used on many sites). In this case, something like what you propose would be required. In any case, this should be considered separately. 

@kmike only your review is remaining then. I've been changing the PR to polish the docu but I have no plans to change the PR anymore.

docs/index.rst
```diff
@@ -12,6 +12,11 @@ way, which separates extraction and crawling logic from the I/O,
 and from the Scrapy implementation details as well.
 It makes the code more testable and reusable.
 
+Beside, it
```
ejulio, line 15: I think this shouldn't be a new paragraph, as it's linked to the idea of the previous one.

Also, but I'm not sure about this one.
_Beside, it opens .._ -> _Furthermore, it opens .._

docs/intro/tutorial.rst
```diff
@@ -238,6 +238,182 @@ At the end of our job, the spider should look like this:
...
+    class BookListPage(WebPage):
+
+        def book_urls(self):
+            return []
+
+
+    class BookPage(ItemWebPage):
+
+        def to_item(self):
+            return {}
```
kmike, line 325: This is a reasonable implementation, but if I'd need to write this cod efor non-educational purposes I'd probably use abc for BookListPage, and remove to_item method from BookPage (as it is already defined, and it is an abstract method in the base class). 

That's a nitpick, and doing as per ^^ suggestion could make code less clear. Feel free to decide what to do @ivanprado :)

docs/intro/tutorial.rst
```diff
@@ -238,6 +238,182 @@ At the end of our job, the spider should look like this:
...
+.. code-block:: python
+
+    SCRAPY_POET_OVERRIDES = {
+        "toscrape.com": {
+            BookListPage: BTSBookListPage,
+            BookPage: BTSBookPage
+        }
+    }
+
+The spider is back to live!
```
kmike, line 358:
```suggestion
The spider is back to life!
```

docs/overrides.rst
```diff
@@ -0,0 +1,78 @@
...
+.. code-block:: python
+
+    class BookPage(ItemWebPage):
+        def to_item(self):
+            ...
+
+Imagine this Page Object is provided by an external library, so you cannot
+directly modify it. But you want to extract an additional attribute (e.g. ``ISBN``) that
+was not extracted by the original Page Object. No problem, you can just override it
+using the following Page Object:
```
kmike, line 28: This particular example can be solved in a more "standard" way with inheritance: inherit from BookPage, call super() in to_item. What do you think about adding a note about this, or providing a different example (not sure which though)?

docs/overrides.rst
```diff
@@ -0,0 +1,78 @@
...
+
+The overrides registry is responsible for informing whether there exists an
+override for a particular type for a given response. The default overrides
+registry keeps a map of overrides for each domain and read this configuration
+from settings ``SCRAPY_POET_OVERRIDES`` as has been seen in the :ref:`intro-tutorial`
+example.
+
+But the registry implementation can be changed at convenience. A different
+registry implementation can be configured using the property
+``SCRAPY_POET_OVERRIDES_REGISTRY`` in ``settings.py``. The new registry
```
kmike, line 73:
```suggestion
registry implementation can be configured using the 
``SCRAPY_POET_OVERRIDES_REGISTRY`` option in ``settings.py``. The new registry
```

scrapy_poet/injection.py
```diff
@@ -31,9 +33,14 @@ class Injector:
     Keep all the logic required to do dependency injection in Scrapy callbacks.
     Initializes the providers from the spider settings at initialization.
     """
-    def __init__(self, crawler: Crawler, default_providers: Optional[Mapping] = None):
+    def __init__(self,
+                 crawler: Crawler,
+                 *,
```
kmike, line 38: 👍

scrapy_poet/injection.py
```diff
@@ -106,6 +109,7 @@ def build_plan(self, request: Request) -> andi.Plan:
             callback,
             is_injectable=is_injectable,
             externally_provided=self.is_class_provided_by_any_provider,
+            overrides=self.overrides_registry.overrides_for(request).get
```
kmike, line 112: What do you think about changing overrides_for interface to return functions? This is a low-level API, so maybe we should be leaning towards power / completeness vs ease of use here.

docs/overrides.rst
kmike: Custom registry definitely solves that.

kmike (approved):
Great job @ivanprado. 

Implementation is very clean - code is clear, organized and tested well, a good split between andi and scrapy-poet, Scrapy conventions are followed. 

You put a lot of effort to documentation, from the start - it would have been easy to skim over this step and do a bare minimum, but here we got a great tutorial from beginning, as well as some lower-level docs. Because there is a lot of docs, they received some of comments from me and @ejulio, but the docs were very good from the start.

It was also a great idea to introduce SCRAPY_POET_OVERRIDES_REGISTRY, it solves a lot of potential issues.

docs/index.rst
ivanprado: Spot on! Changed here https://github.com/scrapinghub/scrapy-poet/pull/43/commits/948885f1fdd891bd24b329259435308af82d3875

docs/intro/tutorial.rst
ivanprado: If the methods were abstract, the spider would raise non implemented errors for URLs not belonging to the domain. I thought it would be better to have a default implementation that just does nothing in such a case. Yeah, but I'm unsure about that. Some users might prefer to see a failure because it could be a failure in the crawling logic. Let's keep it as is for clarity reasons.

However, I've noticed that is better to change the default `to_item()` to return `None`. Otherwise empty items would be returned if the callback is used for a non overridden domain. Done here: https://github.com/scrapinghub/scrapy-poet/pull/43/commits/cbb7e5cc60b30d8cbd79e1edc7144acb4efe1818

scrapy_poet/injection.py
ivanprado: I reckon it would be worth even if it is done only for the fact that the return annotation becomes `Callable[[Callable], Callable]` :rofl: 

Now seriously, the problem I see is that the interface is then very convoluted. See this example of a simple custom registry in such a case:

```python
            class MyRegistry(OverridesRegistryBase):

                def overrides_for(self, request):
                    def overrides_fn(request, cls):
                        if get_domain(request.url) == 'toscrape.com':
                            return BSTBookItem if cls is BookItem else return None
                        return None

                    return partial(overrides_fn, request)
```

The flexibility advantage that you could get makes sense if you have families of things to overrides and you don't want to be explicit about their mapping. For example:

```python
                def overrides_for(self, request):
                    def overrides_fn(request, cls):
                        if issubclass(cls, BookPage):
                            return xxx

                    return partial(overrides_fn, request)
```

But I cannot imagine a real example where this makes sense at all. So my vote here is not to over-engineer it and leave it at is. But I have a open mind :smile:, if you have an example where this could be useful, please, tell me.

docs/overrides.rst
ivanprado: Good point. I've changed the example. Now you cannot use inheritance, because the original PO doesn't have any HTML at all. Changed here:

https://github.com/scrapinghub/scrapy-poet/pull/43/commits/b1bb4c406fcd4af077f0a82c39bc81b1d3478e40

ivanprado: Great point! I've changed the example, and now inheritance is not enough. The change is here: https://github.com/scrapinghub/scrapy-poet/pull/43/commits/b1bb4c406fcd4af077f0a82c39bc81b1d3478e40

ivanprado:
Thank you @ejulio for the re-review and @kmike for the review. I'm happy to see an unexpectedly low number of corrections in my English. It seems the money I'm expending on English classes is paying! :laughing: 

I did some changes and answered some questions, so even if the PR is already approved, I left it some time open just in case you have something to say.

ivanprado:
Merging this. Thank you @ejulio and @kmike . Next: Release

Issue #47, opened 2021-05-30
Improve logging message for DummyResponse


kmike:
Currently in case of DummyResponse a message like this appears in the logs:

> 2021-05-30 00:16:51 [category] DEBUG: Skipping download of <GET https://www.example.com/products/eacs-i-09harx-n3-20061030>

I find it confusing because of thee following:

1. It uses spider name as a logger name, which makes it hard to figure out which component does this. It also makes it hard to filter out or tune these messes if needeed (though this is not very important)
2. It reads as if the spider is not going to download the URL, while in fact it still could, but in the provider. I've been using it with scrapy-autoextract, where downloading is handled via AutoExtract, while in the log it reads like the request is filtered out for some reason. It seems that in most practical cases when DummyResponse is used, the URL is going to be downloaded anyways, but in some different way.

I think it'd be good to use a module-specific logger (`logger = logging.getLogger(__name__)`), and change the message, but I haven't came up with a good messsage yet :)

PR #49, opened 2021-05-30
Improve logging message for DummyResponse. 

kmike:
Fixes GH-47.

Before:

> 2021-05-30 00:16:51 [category] DEBUG: Skipping download of <GET https://www.example.com/products/eacs-i-09harx-n3-20061030>

After:

> 2021-05-31 01:42:27 [scrapy_poet.middleware] DEBUG: Using DummyResponse instead of downloading <GET https://www.example.com/products/eacs-i-09harx-n3-20061030>

### DIFF
--- a/scrapy_poet/middleware.py
+++ b/scrapy_poet/middleware.py
@@ -2,6 +2,7 @@
 responsible for injecting Page Input dependencies before the request callbacks
 are executed.
 """
+import logging
 
 from scrapy import Spider
 from scrapy.crawler import Crawler
@@ -13,6 +14,9 @@ from . import api
 from .overrides import PerDomainOverridesRegistry
 from .page_input_providers import ResponseDataProvider
 from .injection import Injector
+
+
+logger = logging.getLogger(__name__)
 
 
 DEFAULT_PROVIDERS = {
@@ -58,7 +62,7 @@ class InjectionMiddleware:
         if self.injector.is_scrapy_response_required(request):
             return
 
-        spider.logger.debug(f'Skipping download of {request}')
+        logger.debug(f"Using DummyResponse instead of downloading {request}")
         return api.DummyResponse(url=request.url, request=request)
 
     @inlineCallbacks


ivanprado:
@kmike Big +1 to change this log message. It is really really confusing. 

> Using DummyResponse instead of downloading <GET https://www.example.com/products/eacs-i-09harx-n3-20061030>

Still, I wonder if we can improve this message even more. If you are used to `DummyResponse` you might know what is going on, but if you are using an existing spider the message will still be confusing. Idea:

`DummyResponse in use. Scrapy downloader won't fetch <GET https://www.example.com/products/eacs-i-09harx-n3-20061030>`

scrapy_poet/middleware.py
```diff
@@ -58,7 +62,7 @@ def process_request(self, request: Request, spider: Spider):
         if self.injector.is_scrapy_response_required(request):
             return
 
-        spider.logger.debug(f'Skipping download of {request}')
+        logger.debug(f"Using DummyResponse instead of downloading {request}")
```
ivanprado, line 65: As suggested, in case you liked it:
```suggestion
        logger.debug(f"DummyResponse in use. Scrapy downloader won't fetch {request}")
```

kmike:
> DummyResponse in use. Scrapy downloader won't fetch <GET https://www.example.com/products/eacs-i-09harx-n3-20061030>

That's why coming up with a good message is hard here :) The issue with this variant is that a provider is free to use Scrapy downloader to fetch this web page, so the message might be untrue.

kmike:
> 2021-05-31 01:42:27 [scrapy_poet.middleware] DEBUG: Using DummyResponse instead of downloading <GET https://www.example.com/products/eacs-i-09harx-n3-20061030>

I was hoping this reads as "scrapy_poet middleware uses DummyResponse instead of downloading a request directly", which doesn't tell anything about providers or other components, or about the end result (which may differ). So, a lower-level message, but hopefully more accurate. It is still not accurate though, as middlewares don'd download stuff themselves :)

ivanprado:
> The issue with this variant is that a provider is free to use Scrapy downloader to fetch this web page, so the message might be untrue

@kmike Note that at this point in the code we are absolutely sure that no provider has requested `ResponseData`. You mentioned that a provider could directly use the downloader to download the page... but that would be somehow breaking the contract: If a provider want's the request from the downloader, he should just declare that he needs `ResponseData`. 

Also to add more context, these logs events appear in log lines just before a "Crawled (200)" log lines emitted by `scrapy.core.engine`
```
2021-05-31 12:16:29 [XXX] DEBUG: Skipping download of <GET http://books.toscrape.com>
2021-05-31 12:16:39 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://books.toscrape.com> (referer: None)
```

Therefore you get a message saying that the downloading of the page was skipped, and right after, you get a message saying that the page was finally downloaded :confused: . Forget about my last proposal. This is my new one:

```
2021-05-31 12:16:29 [XXX] DEBUG: Skipping Scrapy download of <GET http://books.toscrape.com>. Faking "Crawled (200)" response using DummyResponse
2021-05-31 12:16:39 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://books.toscrape.com> (referer: None)
```

The advantage of the new proposal is that is accurate, as it is describing what is really happening. It restricts the Skipping to "Scrapy download" which opens to door for providers to download the page in a different way. And include "Crawled (200)" that allows to link this log line with the one from `scrapy.core.engine`.

Maybe the line is too long now, though. Anyway, I'm also fine if we keep your current proposal. There is no perfect message here :-)

ivanprado (approved):

kmike:
As discussed on a call with @ivanprado, I'm merging it as-is. It seems there is no a perfect message, but what's here is better than what's in master.

kmike:
Thanks @ivanprado for the review and discussion!

PR #53, opened 2021-08-11
New overrides registry that allows more complex configuration rules for overrides

ivanprado:
The existing overrides registry allowed to configure overrides per domain. But this is not enough in some cases where we need to configure overrides for subdomains or subsections of a site. This PR introduces the class `HierarchicalOverridesRegistry` that allows configuring these rules in a simple and efficient way. Check the provided documentation for more information.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,13 @@
 Changelog
 =========
 
+TBR
+------------------
+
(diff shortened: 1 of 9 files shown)


ivanprado:
Close in favour of https://github.com/scrapinghub/scrapy-poet/pull/56

PR #54, opened 2021-09-13
[WIP] Hierarchichal overrides using regex

ivanprado:
**Do not merge!**

An alternative to https://github.com/scrapinghub/scrapy-poet/pull/53

Existing rules as defined in the original requests are working. But now custom regexes can be used as well when required. The way they can be defined is by tuples containing the domain and a regex over the full URL. 

For example, consider the following configuration:
```py
OVERRIDES_1 = { ... }
OVERRIDES_2 = { ... }
{
"example.com/en_gb": OVERRIDES_1
("example.com", r"http://example.com/.*?product_id=[0-9]*.*): OVERRIDES_2
}
``` 
All URLs with `product_id` argument in the URL will go through `OVERRIDES_2`. Other URLs belonging to the subpath `en_gb` will go through `OVERRIDES_1`. 

This brings the best of the two words: using simple `domain_or_more` rules can still be used, and they will cover most of the cases. If something more powerful is required, then you can use regex. The fact that we are partitioning by `domain` makes also efficient enough this new approach. 

Note that in case of collusion of both rules  (for example for url http://example.com/en_gb/product?product_id=23 in the former example) those defined as regex has priority. And the priority over different regexes is defined by the order in which the regexes are declared. 

The priorities over the `domain_or_more` rules is defined according to the hierarchy (tld, domain, subdomain, path, etc), so the order in which they are defined is not affected. 

There is a possibility: enrich a bit more the `domain_or_more` rules by allowing globbing. For example, why not allow to declare rules like example.com/*product_id=*?. Implementation-wise it is not very hard to do. @kmike what do you think?

Note that we are here allowing registering default POs by using the empty string. Theoretically, you can also register regex for the empty domain and then they will be applied for any URL. But I don't think this is something we should promote, as it is not a good practice from the point of view of performance. 

TODO:

- [ ] Testing regex priorities 
- [ ] Documentation
- [ ] Separated registries per page object type
- [ ] globbing?
- [ ] Removing the old HierarchicalRegistry

### DIFF
--- a/scrapy_poet/overrides.py
+++ b/scrapy_poet/overrides.py
@@ -151,6 +151,11 @@ class HierarchicalOverridesRegistry(OverridesRegistryBase):
 
 @attr.s(auto_attribs=True, order=False)
 class RegexOverridesRecord:
+    """
+    Keep a reverse ordering on hurl. This is required to prioritize the more
+    especific rules over the less especific ones using the hierarchy determined
+    by the hierarchical url.
+    """
     hurl: str = attr.ib(eq=False)
     regex: str
     overrides: Mapping[Callable, Callable] = attr.ib(eq=False)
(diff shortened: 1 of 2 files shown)


ivanprado:
Close in favour or https://github.com/scrapinghub/scrapy-poet/pull/56

PR #55, opened 2021-12-02
SCRAPY_POET_CACHE mechanism

ivanprado:
New providers interface

Error caching can be enabled or disabled.

Compression can be enabled or disabled.

Avoiding having key conflicts between provides was done by adding prefix to each key. It would be more efficient to have a separate table per provider in the sqlite database. I leave this for the future.

Todo:

- [x] FIx tox py36. It seems to get stuck
- [ ] Improve test coverage
- [ ] Documentation

**Possible documentation schema**:

It can be a new section "Cache" in the `ADVANCED` section. 

1.  What is the cache
2. A simple example enabling the cache
3. Cache configuration
3.1 Simple mode (setting it to TRUE)
3.2 Custom cache file (setting a custom file for the cache)
3.3 GZIP compression enable/disable (enabled by default)
3.4. Cache errors. Why. Disabled by default.

The "Providers" section should be updated. It should include:

* How to create providers that can be cached: they must have a name and implement the methods `fingerprint`, `serialize`, `deserialize`
* Explain that cache is important if we want the `scrapy override` command to work
* Explain how these methods behave and how should they be implemented
* Show examples

### DIFF
--- a/scrapy_poet/cache.py
+++ b/scrapy_poet/cache.py
@@ -21,6 +21,10 @@ class _Cache(abc.ABC):
 
 
 class SqlitedictCache(_Cache):
+    """Stores dependencies from Providers in a persistent local storage using
+    https://github.com/RaRe-Technologies/sqlitedict.
+    """
+
     def __init__(self, path: str, *, compressed=True):
         self.path = path
         self.compressed = compressed
--- a/scrapy_poet/cache.py
+++ b/scrapy_poet/cache.py
@@ -2,14 +2,14 @@ import abc
 import gzip
 import pickle
 import sqlite3
+from typing import Any
 
 import sqlitedict
 
 
 class _Cache(abc.ABC):
-
     @abc.abstractmethod
-    def __getitem__(self, fingerprint: str):
+    def __getitem__(self, fingerprint: str) -> Any:
         pass
 
     @abc.abstractmethod
@@ -21,24 +21,26 @@ class _Cache(abc.ABC):
 
 
 class SqlitedictCache(_Cache):
-
-    def __init__(self, path, *, compressed=True):
+    def __init__(self, path: str, *, compressed=True):
+        self.path = path
         self.compressed = compressed
         tablename = 'responses_gzip' if compressed else 'responses'
-        self.db = sqlitedict.SqliteDict(path,
-                                        tablename=tablename,
-                                        autocommit=True,
-                                        encode=self.encode,
-                                        decode=self.decode)
+        self.db = sqlitedict.SqliteDict(
+            path,
+            tablename=tablename,
+            autocommit=True,
+            encode=self.encode,
+            decode=self.decode,
+        )
 
-    def encode(self, obj):
+    def encode(self, obj: Any) -> memoryview:
         # based on sqlitedict.encode
         data = pickle.dumps(obj, pickle.HIGHEST_PROTOCOL)
         if self.compressed:
             data = gzip.compress(data, compresslevel=3)
         return sqlite3.Binary(data)
 
-    def decode(self, obj):
+    def decode(self, obj: Any) -> Any:
         # based on sqlitedict.decode
         data = bytes(obj)
         if self.compressed:
@@ -47,16 +49,21 @@ class SqlitedictCache(_Cache):
             data = gzip.decompress(data)
         return pickle.loads(data)
 
-    def __str__(self):
-        return f"SqlitedictCache <{self.db.filename} | " \
-               f"compressed: {self.compressed} | " \
-               f"{len(self.db)} records>"
+    def __str__(self) -> str:
+        return (
+            f"SqlitedictCache <{self.db.filename} | "
+            f"compressed: {self.compressed} | "
+            f"{len(self.db)} records>"
+        )
 
-    def __getitem__(self, fingerprint: str):
+    def __repr__(self) -> str:
+        return f"SqlitedictCache({self.path!r}, compressed={self.compressed})"
+
+    def __getitem__(self, fingerprint: str) -> Any:
         return self.db[fingerprint]
 
-    def __setitem__(self, fingerprint: str, value) -> None:
+    def __setitem__(self, fingerprint: str, value: Any) -> None:
         self.db[fingerprint] = value
 
-    def close(self):
+    def close(self) -> None:
         self.db.close()
--- a/scrapy_poet/cache.py
+++ b/scrapy_poet/cache.py
@@ -1,0 +1,62 @@
+import abc
+import gzip
+import pickle
+import sqlite3
+
+import sqlitedict
+
+
+class _Cache(abc.ABC):
+
+    @abc.abstractmethod
+    def __getitem__(self, fingerprint: str):
+        pass
+
+    @abc.abstractmethod
+    def __setitem__(self, fingerprint: str, value) -> None:
+        pass
+
+    def close(self):
+        pass
+
+
+class SqlitedictCache(_Cache):
+
+    def __init__(self, path, *, compressed=True):
+        self.compressed = compressed
+        tablename = 'responses_gzip' if compressed else 'responses'
+        self.db = sqlitedict.SqliteDict(path,
+                                        tablename=tablename,
+                                        autocommit=True,
+                                        encode=self.encode,
+                                        decode=self.decode)
+
+    def encode(self, obj):
+        # based on sqlitedict.encode
+        data = pickle.dumps(obj, pickle.HIGHEST_PROTOCOL)
+        if self.compressed:
+            data = gzip.compress(data, compresslevel=3)
+        return sqlite3.Binary(data)
+
+    def decode(self, obj):
+        # based on sqlitedict.decode
+        data = bytes(obj)
+        if self.compressed:
+            # gzip is slightly less efficient than raw zlib, but it does
+            # e.g. crc checks out of box
+            data = gzip.decompress(data)
+        return pickle.loads(data)
+
+    def __str__(self):
+        return f"SqlitedictCache <{self.db.filename} | " \
+               f"compressed: {self.compressed} | " \
+               f"{len(self.db)} records>"
+
+    def __getitem__(self, fingerprint: str):
+        return self.db[fingerprint]
+
+    def __setitem__(self, fingerprint: str, value) -> None:
+        self.db[fingerprint] = value
+
+    def close(self):
+        self.db.close()
--- a/scrapy_poet/injection.py
+++ b/scrapy_poet/injection.py
@@ -66,6 +66,10 @@ class Injector:
         self.is_class_provided_by_any_provider = \
             is_class_provided_by_any_provider_fn(self.providers)
 
+    def close(self) -> None:
+        if self.cache:
+            self.cache.close()
+
     def init_cache(self):
         self.cache = None
         cache_filename = self.spider.settings.get('SCRAPY_POET_CACHE')
@@ -361,7 +365,7 @@ def get_response_for_testing(callback: Callable) -> Response:
         <html>
             <body>
                 <div class="breadcrumbs">
-                    <a href="/food">Food</a> / 
+                    <a href="/food">Food</a> /
                     <a href="/food/sweets">Sweets</a>
                 </div>
                 <h1 class="name">Chocolate</h1>
--- a/scrapy_poet/injection.py
+++ b/scrapy_poet/injection.py
@@ -1,8 +1,10 @@
 import inspect
 import logging
+import os
 import pprint
 from typing import Dict, Callable, Any, List, Set, Mapping, Optional
 
+from .utils import get_scrapy_data_path
 from twisted.internet.defer import inlineCallbacks
 
 import andi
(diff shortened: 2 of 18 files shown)


BurnzZ (commented):
Hi @ivanprado, added up some small comments here on the first pass. I still need to review the internals for the injection mechanism on the providers to have a better input on the PR review.

Overall, well done for writing a solid code here! 💯

tests/test_injection.py
```diff
@@ -364,3 +368,94 @@ class WrongProvider(PageObjectInputProvider):
...
+    injector = get_injector_for_testing(providers, settings)
+
+    response = get_response_for_testing(callback)
+    plan = injector.build_plan(response.request)
+    instances = yield from injector.build_instances_from_providers(
+        response.request, response, plan)
+
+    assert instances[str] == "foo"
+    assert instances[int] == 3
+    assert instances[float] == 3.0
```
BurnzZ, line 453: We could group these assertions with the same code block as above and call a single function.

tests/test_injection.py
```diff
@@ -364,3 +368,94 @@ class WrongProvider(PageObjectInputProvider):
...
+
+    response.request = Request.replace(response.request, url="http://willfail.page")
+    with pytest.raises(ValueError):
```
BurnzZ, line 434: It's not directly clear to me why such operation below would fail. Perhaps some comments in the code would help future readers as well.

sortafreel (approved):
GJ Ivan :+1:

scrapy_poet/injection.py
```diff
@@ -61,6 +66,17 @@ def load_providers(self, default_providers: Optional[Mapping] = None):
         self.is_class_provided_by_any_provider = \
             is_class_provided_by_any_provider_fn(self.providers)
 
+    def init_cache(self):
+        self.cache = None
+        cache_filename = self.spider.settings.get('SCRAPY_POET_CACHE')
+        if cache_filename and isinstance(cache_filename, bool):
```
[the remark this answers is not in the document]
ivanprado: If the value is `True` (a boolean), then the cache location is automatic: it will be under `.scrapy` folder. If the value is a folder (that is, of type `str`) then the cache will be saved in the corresponding folder. If false or empty, then the cache will be disabled.

scrapy_poet/cache.py
```diff
@@ -0,0 +1,62 @@
...
+    def __setitem__(self, fingerprint: str, value) -> None:
+        pass
+
+    def close(self):
+        pass
+
+
+class SqlitedictCache(_Cache):
+
+    def __init__(self, path, *, compressed=True):
```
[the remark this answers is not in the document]
BurnzZ: @sortafreel This syntax forces the 2nd parameter to always use its **keyword**-value like this: `SqlitedictCache("path/to/somewhere", compressed=True)`.

Otherwise, calling it without the keyword args to the right the `*` specifier like `SqlitedictCache("path/to/somewhere", True)`, results in `TypeError: SqlitedictCache() takes 1 positional argument but 2 were given`.

Reference: https://www.python.org/dev/peps/pep-0570/#specification

scrapy_poet/utils.py
```diff
@@ -13,3 +16,15 @@ def get_domain(url):
     'example.co.uk'
     """
     return ".".join(tldextract.extract(url)[-2:])
+
+
+def get_scrapy_data_path(createdir=True):
```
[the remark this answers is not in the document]
BurnzZ: Updated in https://github.com/scrapinghub/scrapy-poet/pull/55/commits/431f4763f71a1895288539c5801298e5ad6773b7

scrapy_poet/cache.py
```diff
@@ -0,0 +1,62 @@
...
+        if self.compressed:
+            # gzip is slightly less efficient than raw zlib, but it does
+            # e.g. crc checks out of box
+            data = gzip.decompress(data)
+        return pickle.loads(data)
+
+    def __str__(self):
+        return f"SqlitedictCache <{self.db.filename} | " \
+               f"compressed: {self.compressed} | " \
+               f"{len(self.db)} records>"
```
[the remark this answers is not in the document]
BurnzZ: Handled in https://github.com/scrapinghub/scrapy-poet/pull/55/commits/431f4763f71a1895288539c5801298e5ad6773b7

scrapy_poet/cache.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __getitem__(self, fingerprint: str):
```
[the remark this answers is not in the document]
BurnzZ: Handled in https://github.com/scrapinghub/scrapy-poet/pull/55/commits/431f4763f71a1895288539c5801298e5ad6773b7

scrapy_poet/injection.py
```diff
@@ -142,14 +158,44 @@ def build_instances_from_providers(
...
+                    is_injectable=is_injectable,
+                    externally_provided=scrapy_provided_dependencies,
+                    full_final_kwargs=False,
+                ).final_kwargs(scrapy_provided_dependencies)
+                try:
+
+                    # Invoke the provider to get the data
+                    objs = yield maybeDeferred_coro(provider, set(provided_classes), **kwargs)
+
+                except BaseException as e:
```
kmike, line 192: Why is BaseException caught here, not just Exception? It'd include system-exiting exception; I'm not sure we should be catching those.

scrapy_poet/page_input_providers.py
```diff
@@ -101,6 +106,26 @@ def __init__(self, crawler: Crawler):
         """Initializes the provider. Invoked only at spider start up."""
         pass
 
+    def fingerprint(self, to_provide: Set[Callable], request: Request) -> str:
```
kmike, line 109: I wonder if those 3 methods can be moved to an abstract mixing. When you implement a provider which supports caching, you inherit from this mixin, and get all your methods checked (that all 3 needs to be implemented).

scrapy_poet/page_input_providers.py
```diff
@@ -117,7 +142,17 @@ class ResponseDataProvider(PageObjectInputProvider):
...
 
     provided_classes = {ResponseData}
+    name = "response_data"
 
     def __call__(self, to_provide: Set[Callable], response: Response):
         """Builds a ``ResponseData`` instance using a Scrapy ``Response``"""
         return [ResponseData(url=response.url, html=response.text)]
+
+    def fingerprint(self, to_provide: Set[Callable], request: Request) -> str:
+        return request_fingerprint(request)
```
kmike, line 152: When would a developer prefer built-in Scrapy's cache, and when should ResponseDataProvider's cache be used?

kmike: Also, about fingerprints. The approach here makes total sense, as this is what built-in Scrapy cache is using - request_fingerprint. It also has a performance advantage, as these fingerprints are used in many places, and they're cached. 

However, these fingerprints are stored in sqlite columns now. So, when deciding on a fingerprint for AutoExtract provider, I opted in json-encoded request data (https://github.com/scrapinghub/scrapy-autoextract/blob/85f6318a9212f53300ad9569248600e25e98017e/scrapy_autoextract/cache.py#L72). It is much less efficient, and takes disk space, but it's much easier to inspect the cache - e.g. find and delete cache records for a certain domain, or a certain URL.

scrapy_poet/injection.py
```diff
@@ -61,6 +66,17 @@ def load_providers(self, default_providers: Optional[Mapping] = None):
...
             is_class_provided_by_any_provider_fn(self.providers)
 
+    def init_cache(self):
+        self.cache = None
+        cache_filename = self.spider.settings.get('SCRAPY_POET_CACHE')
+        if cache_filename and isinstance(cache_filename, bool):
+            cache_filename = os.path.join(get_scrapy_data_path(createdir=True), "scrapy-poet-cache.sqlite3")
+        if cache_filename:
+            compressed = self.spider.settings.getbool('SCRAPY_POET_CACHE_GZIP', True)
+            self.caching_errors = self.spider.settings.getbool('SCRAPY_POET_CACHE_ALSO_ERRORS', False)
```
kmike, line 76: What do you think about calling the option just `SCRAPY_POET_CACHE_ERRORS`?

scrapy_poet/cache.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class _Cache(abc.ABC):
+
+    @abc.abstractmethod
+    def __getitem__(self, fingerprint: str):
+        pass
+
+    @abc.abstractmethod
+    def __setitem__(self, fingerprint: str, value) -> None:
```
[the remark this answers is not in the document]
BurnzZ: Handled in https://github.com/scrapinghub/scrapy-poet/pull/55/commits/431f4763f71a1895288539c5801298e5ad6773b7

tests/test_injection.py
BurnzZ: addressed in https://github.com/scrapinghub/scrapy-poet/pull/55/commits/027e884f2a800eb54b19cb1911aba5c1503fe4e1

tests/test_injection.py
BurnzZ: addressed in https://github.com/scrapinghub/scrapy-poet/pull/55/commits/027e884f2a800eb54b19cb1911aba5c1503fe4e1

scrapy_poet/injection.py
```diff
@@ -61,6 +66,17 @@ def load_providers(self, default_providers: Optional[Mapping] = None):
         self.is_class_provided_by_any_provider = \
             is_class_provided_by_any_provider_fn(self.providers)
 
+    def init_cache(self):
+        self.cache = None
```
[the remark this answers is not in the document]
BurnzZ: addressed in https://github.com/scrapinghub/scrapy-poet/pull/55/commits/65d8d75446712cf4b007e9f22e0470a269e0d7a8

scrapy_poet/injection.py
BurnzZ: +1 on this. Updated on https://github.com/scrapinghub/scrapy-poet/pull/55/commits/df29c938bf888b13339c52003b2eaa5e16a8a16d

scrapy_poet/injection.py
BurnzZ: Agreed. Updated on https://github.com/scrapinghub/scrapy-poet/pull/55/commits/1fd5f891d56d92c3070ccc02178b4f86adac74fa

scrapy_poet/page_input_providers.py
BurnzZ: > When would a developer prefer built-in Scrapy's cache, and when should ResponseDataProvider's cache be used?

I shall be adding this in the docs in a succeeding commit.

> It is much less efficient, and takes disk space, but it's much easier to inspect the cache - e.g. find and delete cache records for a certain domain, or a certain URL.

Those trade-offs makes sense to me. I also agree about prioritizing debuggability first and foremost. Should performance deterioration be observed on certain scenarios, the user could always override the `fingerprint()` method.

I've made changes in-line with this on https://github.com/scrapinghub/scrapy-poet/pull/55/commits/734bd5ed0abe45ba6b15b671b81d6a7d57acd216. I've still preserved Scrapy's Request Fingerprint so that the user can easily use it as a reference later on if needed.

scrapy_poet/page_input_providers.py
BurnzZ: Good call. Refactored this on https://github.com/scrapinghub/scrapy-poet/pull/55/commits/15c5a7fe4fd968c2cb103ffb9820b79791dc9119.

Made it into an abstract base class to force the user to implement them when subclassing. Also fixed a bug where if `SCRAPY_POET_CACHE` is being used, all providers should support the cache. This should not be the case as each provider can optionally support it.

scrapy_poet/injection.py
```diff
@@ -142,14 +162,44 @@ def build_instances_from_providers(
...
-            objs = yield maybeDeferred_coro(provider, set(provided_classes),
-                                               **kwargs)
+            objs, fingerprint = None, None
+            cache_hit = False
+            if self.cache and provider.has_cache_support:
+                if not provider.name:
+                    raise NotImplementedError(f"The provider {type(provider)} must have a `name` defined if"
+                                              f" you want to use the cache. It must be unique across the providers.")
+                # Return the data if it is already in the cache
+                fingerprint = f"{provider.name}_{provider.fingerprint(set(provided_classes), request)}"
```
kmike, line 172: (very minor) What do you think about extracting a method which computes these fingerprints? Use case: get a cache key programmatically, if you know the provider.

kmike: Also, an idea, probably for the future: store cache for different providers in different tables (sqlitedict supports this via "tablename" argument, when SqliteDict is created). The main advantage would be that fingerprints would remain json. sqlite may even be able to query inside, via json1 extension. But that's all not needed now :)

kmike:
Thanks @ivanprado, @BurnzZ and @sortafreel!

scrapy_poet/injection.py
BurnzZ: Got it. Makes sense. I've created a draft PR here before these ideas get buried down. 😄 https://github.com/scrapinghub/scrapy-poet/pull/59

PR #59, opened 2021-12-22
WIP: small improvements

BurnzZ:
Stemming from this PR review: https://github.com/scrapinghub/scrapy-poet/pull/55#pullrequestreview-836638135

------

> (very minor) What do you think about extracting a method which computes these fingerprints? Use case: get a cache key programmatically, if you know the provider.

> Also, an idea, probably for the future: store cache for different providers in different tables (sqlitedict supports this via "tablename" argument, when SqliteDict is created). The main advantage would be that fingerprints would remain json. sqlite may even be able to query inside, via json1 extension.

### DIFF
--- a/scrapy_poet/injection.py
+++ b/scrapy_poet/injection.py
@@ -2,7 +2,7 @@ import inspect
 import logging
 import os
 import pprint
-from typing import Dict, Callable, Any, List, Set, Mapping, Optional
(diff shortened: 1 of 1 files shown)


BurnzZ:
Closing as we'll handle this in a larger PR.

PR #60, opened 2022-02-06
docs: update docs urls to use https

peter-gy:
Accessing http://scrapy-poet.readthedocs.io is not possible, as redirection to https does not always happen automatically. This PR fixes the issue.


BurnzZ (approved):
Thanks @peter-gy ! 🎉

PR #56, opened 2021-12-08
url-matcher integration with scrapy-poet

ivanprado:
**WARNING: this branch was created on top of https://github.com/scrapinghub/scrapy-poet/pull/55. It should be merged only after the former one has been merged**

New registry that uses the `url-matcher` patterns to configure the overrides. It is a backwards-incompatible change. See the updated documentation to get more context.

A new release of web-poet is required because this depends on ~https://github.com/scrapinghub/web-poet/pull/16~ https://github.com/scrapinghub/web-poet/pull/27

TODO: 
- [ ] Update the web-poet deps both in `tox.ini` and `setup-py` once a new version is released
- [x] Update the docs to link **web-poet** contents like `OverrideRule`, `PageObjectRegistry`, etc. using `:ref:` directive. But it requires https://github.com/scrapinghub/web-poet/pull/27 to be merged first.

PR #62, opened 2022-02-08
integration for web-poet's support on additional requests and Meta

### LINKED ISSUE #65: `callback_for` does not produce a coroutine.
Looks like `callback_for` doesn't produce parse method that is a coroutine. So it's not possible to use `callback_for` in case `to_item` method is a coroutine (declared with async/await syntax).

Example:
```python
import asyncio
import inspect

from scrapy_poet import callback_for
from web_poet.pages import ItemWebPage


class SimplePage(ItemWebPage):
    def to_item(self):
        return {}

cb = callback_for(SimplePage)
print(f"Is callback a coroutine: {inspect.iscoroutinefunction(cb)}")
print(f"Is to_item a coroutine: {inspect.iscoroutinefunction(SimplePage.to_item)}")

class SimplePageAIO(ItemWebPage):
    async def to_item(self):
        await asyncio.sleep(0.1)
        return {}

cb_aio = callback_for(SimplePageAIO)
print(f"Is callback a coroutine: {inspect.iscoroutinefunction(cb_aio)}")
print(f"Is to_item a coroutine: {inspect.iscoroutinefunction(SimplePageAIO.to_item)}")
```
This produces:
```
Is callback a coroutine: False
Is to_item a coroutine: False 
Is callback a coroutine: False
Is to_item a coroutine: True
```
But it should be like:
```
Is callback a coroutine: False
Is to_item a coroutine: False 
Is callback a coroutine: True
Is to_item a coroutine: True
```
Tested on `dev-1` branch.

BurnzZ:
Requires the following to be merged in `web-poet`:
- https://github.com/scrapinghub/web-poet/pull/22
- https://github.com/scrapinghub/web-poet/pull/33

This is built on top of 
- https://github.com/scrapinghub/scrapy-poet/pull/67

### Progress

- [x] Testing across a wide-variety of scenarios/cases
     - contextvars and Twisted
     - scrapy-poet and scrapy integration
- [x] Docs
    - [x] update doc links using intersphinx-mapping when `web-poet` is released.
- [x] update error handling after https://github.com/scrapinghub/web-poet/pull/38 is merged
    - [x] ensure that Scrapy doesn't redirect on `HEAD` requests ([reference](https://github.com/scrapinghub/web-poet/pull/38#discussion_r867297335))
- [x] Tests
    - [x] Add test to verify that additional requests work both with an asyncio and a non-asyncio Twisted reactor.
- [x] Changelog
- [x] Use dont_filter for additional requests
- [x] Document that spider middlewares are not followed for additional requests, while downloader middlewares are.
- Once there is a web-poet release with the new page input classes, like `HttpResponse`:
    - [x] Update minimum dependencies `web-poet>=0.2.0` and `scrapy>=2.6.0` in **setup.py** and **tox.ini**

Fixes #65.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ TBR
       ``OverridesRegistry`` which provides a wide variety of features
       for better URL matching.
     * This resuls in a newer format in the ``SCRAPY_POET_OVERRIDES`` setting.
+* Removal of this deprecated module: ``scrapy.utils.reqser``
 
 * add ``async`` support for ``callback_for``.
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,8 @@ TBR
       for better URL matching.
     * This resuls in a newer format in the ``SCRAPY_POET_OVERRIDES`` setting.
 
+* add ``async`` support for ``callback_for``.
+
 
 0.3.0 (2022-01-28)
 ------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+TBR
+---
+
+* Use the new ``web_poet.HttpResponse`` which replaces ``web_poet.ResponseData``.
+
 
 0.3.0 (2022-01-28)
 ------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,7 @@ Changelog
 =========
 
 
-0.3.0 (2021-01-28)
+0.3.0 (2022-01-28)
 ------------------
 
 * Cache mechanism using ``SCRAPY_POET_CACHE``
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,7 +9,8 @@ TBR
 * Support for the new features in ``web_poet>=0.2.0`` for supporting additional 
   requests inside Page Objects:
 
-    * Created new providers for ``web_poet.Meta`` and ``web_poet.HttpClient``.
+    * Created new providers for ``web_poet.PageParams`` and
+      ``web_poet.HttpClient``.
     * The minimum Scrapy version is now ``2.6.0``.
 * We have these **backward incompatible** changes since the
   ``web_poet.OverrideRule`` follow a different structure:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,13 @@ TBR
     * Created new providers for ``web_poet.Meta`` and ``web_poet.HttpClient``.
     * Using the said additional requests needs ``async/await`` support in
       ``asyncio``. This raises the minimum scrapy requirement to ``scrapy>=2.6.0``.
+* We have these **backward incompatible** changes since the
+  ``web_poet.OverrideRule`` follow a different structure:
+
+    * Deprecated ``PerDomainOverridesRegistry`` in lieu of the newer
+      ``OverridesRegistry`` which provides a wide variety of features
+      for better URL matching.
+    * This resuls in a newer format in the ``SCRAPY_POET_OVERRIDES`` setting.
 
 
 0.3.0 (2022-01-28)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,12 @@ TBR
 ---
 
 * Use the new ``web_poet.HttpResponse`` which replaces ``web_poet.ResponseData``.
+* Support for the new features in ``web_poet>=0.2.0`` for supporting additional 
+  requests inside Page Objects:
+
+    * Created new providers for ``web_poet.Meta`` and ``web_poet.HttpClient``.
+    * Using the said additional requests needs ``async/await`` support in
+      ``asyncio``. This raises the minimum scrapy requirement to ``scrapy>=2.6.0``.
 
 
 0.3.0 (2022-01-28)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,8 +10,7 @@ TBR
   requests inside Page Objects:
 
     * Created new providers for ``web_poet.Meta`` and ``web_poet.HttpClient``.
-    * Using the said additional requests needs ``async/await`` support in
-      ``asyncio``. This raises the minimum scrapy requirement to ``scrapy>=2.6.0``.
+    * The minimum Scrapy version is now ``2.6.0``.
 * We have these **backward incompatible** changes since the
   ``web_poet.OverrideRule`` follow a different structure:
 
--- a/docs/intro/advanced-tutorial.rst
+++ b/docs/intro/advanced-tutorial.rst
@@ -131,7 +131,7 @@ This basically acts as a switch to update the behavior of the Page Object:
 
 Passing the ``enable_extracting_all_images`` meta value from the spider into
 the Page Object can be achieved by using **Scrapy's** ``Request.meta`` attribute.
-Specifically, any ``dict`` value inside the ``po_args`` parameter inside
+Specifically, any ``dict`` value inside the ``po_meta`` parameter inside
 **Scrapy's** ``Request.meta`` will be passed into ``web_poet.Meta``.
 
 Let's see it in action:
@@ -159,7 +159,7 @@ Let's see it in action:
                 yield scrapy.Request(
                     url=url,
                     callback=self.parse,
-                    meta={"po_args": {"enable_extracting_all_images": True}}
+                    meta={"po_meta": {"enable_extracting_all_images": True}}
                 )
 
         async def parse(self, response, page: ProductPage):
--- a/docs/intro/advanced-tutorial.rst
+++ b/docs/intro/advanced-tutorial.rst
@@ -54,11 +54,10 @@ Suppose we have the following Page Object:
             }
 
             # Simulates clicking on a button that says "View All Images"
-            response: web_poet.ResponseData = await self.http_client.get(
+            response: web_poet.HttpResponse = await self.http_client.get(
                 f"https://api.example.com/v2/images?id={item['product_id']}"
             )
-            page = web_poet.WebPage(response)
-            item["images"] = page.css(".product-images img::attr(src)").getall()
+            item["images"] = response.css(".product-images img::attr(src)").getall()
             return item
 
 
@@ -122,11 +121,10 @@ This basically acts as a switch to update the behavior of the Page Object:
 
             # Simulates clicking on a button that says "View All Images"
             if self.meta.get("enable_extracting_all_images")
-                response: web_poet.ResponseData = await self.http_client.get(
+                response: web_poet.HttpResponse = await self.http_client.get(
                     f"https://api.example.com/v2/images?id={item['product_id']}"
                 )
-                page = web_poet.WebPage(response)
-                item["images"] = page.css(".product-images img::attr(src)").getall()
+                item["images"] = response.css(".product-images img::attr(src)").getall()
 
             return item
 
--- a/docs/intro/advanced-tutorial.rst
+++ b/docs/intro/advanced-tutorial.rst
@@ -143,7 +143,7 @@ Let's see it in action:
     import scrapy
 
 
-    def ProductSpider(scrapy.Spider):
+    class ProductSpider(scrapy.Spider):
 
         custom_settings = {
             "DOWNLOADER_MIDDLEWARES": {
--- a/docs/intro/advanced-tutorial.rst
+++ b/docs/intro/advanced-tutorial.rst
@@ -8,12 +8,12 @@ This section intends to go over the supported features in **web-poet** by
 **scrapy-poet**:
 
     * ``web_poet.HttpClient``
-    * ``web_poet.Meta``
+    * ``web_poet.PageParams``
 
 These are mainly achieved by **scrapy-poet** implementing **providers** for them:
 
     * :class:`scrapy_poet.page_input_providers.HttpClientProvider`
-    * :class:`scrapy_poet.page_input_providers.MetaProvider`
+    * :class:`scrapy_poet.page_input_providers.PageParamsProvider`
 
 .. _`intro-additional-requests`:
 
@@ -94,11 +94,11 @@ since the ``to_item()`` method of the Page Object we're using is an ``async``
 method as well.
 
 
-Meta
-====
+Page params
+===========
 
-Using ``web_poet.Meta`` allows the Scrapy spider to pass any arbitrary information
-into the Page Object.
+Using ``web_poet.PageParams`` allows the Scrapy spider to pass any arbitrary
+information into the Page Object.
 
 Suppose we update the earlier Page Object to control the additional request.
 This basically acts as a switch to update the behavior of the Page Object:
@@ -112,7 +112,7 @@ This basically acts as a switch to update the behavior of the Page Object:
     @attr.define
     class ProductPage(web_poet.ItemWebPage):
         http_client: web_poet.HttpClient
-        meta: web_poet.Meta
+        page_params: web_poet.PageParams
 
         async def to_item(self):
             item = {
@@ -122,7 +122,7 @@ This basically acts as a switch to update the behavior of the Page Object:
             }
 
             # Simulates clicking on a button that says "View All Images"
-            if self.meta.get("enable_extracting_all_images")
+            if self.page_params.get("enable_extracting_all_images")
                 response: web_poet.HttpResponse = await self.http_client.get(
                     f"https://api.example.com/v2/images?id={item['product_id']}"
                 )
@@ -130,10 +130,11 @@ This basically acts as a switch to update the behavior of the Page Object:
 
             return item
 
-Passing the ``enable_extracting_all_images`` meta value from the spider into
-the Page Object can be achieved by using **Scrapy's** ``Request.meta`` attribute.
-Specifically, any ``dict`` value inside the ``po_meta`` parameter inside
-**Scrapy's** ``Request.meta`` will be passed into ``web_poet.Meta``.
+Passing the ``enable_extracting_all_images`` page parameter from the spider
+into the Page Object can be achieved by using **Scrapy's** ``Request.meta``
+attribute. Specifically, any ``dict`` value inside the ``page_params``
+parameter inside **Scrapy's** ``Request.meta`` will be passed into
+``web_poet.PageParams``.
 
 Let's see it in action:
 
@@ -160,7 +161,7 @@ Let's see it in action:
                 yield scrapy.Request(
                     url=url,
                     callback=self.parse,
-                    meta={"po_meta": {"enable_extracting_all_images": True}}
+                    meta={"page_params": {"enable_extracting_all_images": True}}
                 )
 
         async def parse(self, response, page: ProductPage):
--- a/docs/intro/advanced-tutorial.rst
+++ b/docs/intro/advanced-tutorial.rst
@@ -93,9 +93,6 @@ Note that we needed to update the ``parse()`` method to be an ``async`` method,
 since the ``to_item()`` method of the Page Object we're using is an ``async``
 method as well.
 
-This is also the primary reason why **scrapy-poet** requires ``scrapy>=2.6.0``
-since it's the minimum version that has full :mod:`asyncio` support.
-
 
 Meta
 ====
--- a/docs/intro/advanced-tutorial.rst
+++ b/docs/intro/advanced-tutorial.rst
@@ -26,14 +26,18 @@ the spider. It would work as-is because of the readily available
 out of the box.
 
 This supplies the Page Object with the necessary ``web_poet.HttpClient`` instance.
-Take note the HTTP Downloader implementation that **scrapy-poet** provides to
-``web_poet.HttpClient`` would be the **Scrapy Downloader**.
 
-.. tip::
+The HTTP client implementation that **scrapy-poet** provides to
+``web_poet.HttpClient`` handles requests as follows:
 
-    This means that the additional requests inside a Page Object will have access
-    to the **Downloader Middlewares** that the Spider is using.
+-   Requests go through downloader middlewares, but they do not go through
+    spider middlewares or through the scheduler.
 
+-   Duplicate requests are not filtered out.
+
+-   In line with the web-poet specification for additional requests,
+    ``Request.meta['dont_redirect']`` is set to ``True`` for requests with the
+    ``HEAD`` HTTP method.
 
 Suppose we have the following Page Object:
 
--- a/docs/intro/advanced-tutorial.rst
+++ b/docs/intro/advanced-tutorial.rst
@@ -73,7 +73,7 @@ It can be directly used inside the spider as:
     import scrapy
 
 
-    def ProductSpider(scrapy.Spider):
+    class ProductSpider(scrapy.Spider):
 
         custom_settings = {
             "DOWNLOADER_MIDDLEWARES": {
--- a/docs/intro/advanced-tutorial.rst
+++ b/docs/intro/advanced-tutorial.rst
@@ -1,0 +1,167 @@
+.. _`intro-advanced-tutorial`:
+
+=================
+Advanced Tutorial
+=================
+
+This section intends to go over the supported features in **web-poet** by
+**scrapy-poet**:
+
+    * ``web_poet.HttpClient``
+    * ``web_poet.Meta``
+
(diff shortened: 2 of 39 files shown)


tox.ini
```diff
@@ -7,7 +7,7 @@ deps =
     pytest-cov
     scrapy >= 2.1.0
     pytest-twisted
-    web-poet
+    web-poet @ git+https://git@github.com/scrapinghub/web-poet@additional-requests#egg=web-poet
```
BurnzZ, line 10: **TODO**: Reminder to revert this change before merging.

scrapy_poet/middleware.py
```diff
@@ -48,11 +51,16 @@ def __init__(self, crawler: Crawler) -> None:
...
         o = cls(crawler)
         crawler.signals.connect(o.spider_closed, signal=signals.spider_closed)
+        crawler.signals.connect(o.spider_opened, signal=signals.spider_opened)
         return o
 
     def spider_closed(self, spider: Spider) -> None:
         self.injector.close()
 
+    def spider_opened(self, spider):
+        scrapy_downloader_var.set(spider.crawler.engine.download)
```
kmike, line 61: How does it work with multiple spiders in the same process? It seems this might be a problem. I'm not sure about the solution yet.

BurnzZ: I've tested this out but couldn't find any problem as running multiple spiders in the same process works okay.

@kmike I've tagged you on a separate PR which shows the simple test I've made. Perhaps there's some problem or behavior change that I've missed out.

kmike: answered there, thanks @BurnzZ!

BurnzZ: Hi @kmike , cheers again for raising. 

After some deep investigation on this matter, it would seem that **Twisted** only has a limited support for **ContextVars** at the moment. It currently breaks some of our expectations in using **ContextVars** inside Scrapy. _(see section below for full explanation)_

With that, I've updated the approach inside `scrapy-poet` to use **Dependency Injection** instead. This creates a consistent non-leaking value across different multiple CrawlerProcesses. Here's the commit for this: https://github.com/scrapinghub/scrapy-poet/pull/62/commits/e3b96dce17b199ebe3804aa68f20a1e6ee3b96d0

### Investigation Notes
---------------------

It would seem that having this order of multiple spiders in a single process:

```python
process = CrawlerProcess(settings)
process.crawl(SpiderAAA)
process.crawl(SpiderBBB)
process.crawl(SpiderCCC)
process.start()
```

would result in:

- ✅ **SpiderAAA** using the downloader from **SpiderAAA** itself
- ❌ **SpiderBBB** using the downloader from **SpiderAAA**
- ❌ **SpiderCCC** using the downloader from **SpiderAAA**

In other words, when the first spider sets the downloader, it would be used throughout in the succeeding spiders, regardless if the succeeding spiders attempt to update the ContextVars directly. I've tried it with more spiders and the behavior is simply the same.

We can override the `ContextVar` value inside the spider though. However, after some testing, it would seem that Twisted leaks the context outside of each crawler process. So if we update the `ContextVar` value inside **SpiderBBB**, it would leak as:

- ✅ **SpiderAAA** using the downloader from **SpiderAAA** itself
- ✅ **SpiderBBB** using the downloader from **SpiderBBB**
- ❌ **SpiderCCC** using the downloader from **SpiderBBB**


Interestingly, the behavior above only happens when `twisted>=21.2.0` since it introduced this change:
> twisted.internet.defer.inlineCallbacks and ensureDeferred will now associate a contextvars.Context with the coroutines they run, meaning that ContextVar objects will maintain their value within the same coroutine, similarly to asyncio Tasks. This functionality requires Python 3.7+, or the contextvars PyPI backport to be installed for Python 3.5-3.6. (#9719, #9826)

Here's the code comment in Twisted that tests the ContextVars in InlineCallbacks ([reference](https://github.com/twisted/twisted/pull/1192/files#diff-d51be359d7730f86734b450ecbf6dc8f54090e84e7efcf5616399fb5cffc96a7R3114-R3118)).

> When an inlineCallbacks function is called, the context is taken from
> when it was first called. When it resumes, the same context is applied.

A twisted version below `21.2.0` would result in the the following as expected ([Reference](https://github.com/twisted/twisted/releases/tag/twisted-21.2.0)):

- ✅ **SpiderAAA** using the downloader from **SpiderAAA** itself
- ✅ **SpiderBBB** using the downloader from **SpiderBBB**
- ✅ **SpiderCCC** using the downloader from **SpiderCCC**

However, the expected correct behavior above only happens during normal spider operation. When the time comes where an Additional Request is to be executed in the downloader, the ContextVar raises a `LookupError` since the context is not preserved. ❌

After digging further, it would seem that Twisted doesn't support all of the ContextVar features completely:

- [Support contextvars in Deferred](https://twistedmatrix.com/trac/ticket/9807)
- [Support contextvars in DelayedCall](https://twistedmatrix.com/trac/ticket/9824)
- [Contextvars support does not support `.reset`](https://twistedmatrix.com/trac/ticket/10301)

CHANGELOG.rst
```diff
@@ -2,6 +2,17 @@
 Changelog
 =========
 
+TBR
+---
+
+* Support for the new features in ``web_poet>=0.2.0`` for supporting additional
```
BurnzZ, line 8: Reminder to confirm this before merging since the anticipated `web_poet` version release might change.

docs/intro/advanced-tutorial.rst
```diff
@@ -0,0 +1,165 @@
...
+                    f"https://api.example.com/v2/images?id={item['product_id']}"
+                )
+                item["images"] = response.css(".product-images img::attr(src)").getall()
+
+            return item
+
+Passing the ``enable_extracting_all_images`` meta value from the spider into
+the Page Object can be achieved by using **Scrapy's** ``Request.meta`` attribute.
+Specifically, any ``dict`` value inside the ``po_args`` parameter inside
+**Scrapy's** ``Request.meta`` will be passed into ``web_poet.Meta``.
```
kmike, line 134: I think the fact that @BurnzZ you felt a need to call a key for web_poet Meta "``po_args``" means we should think about web_poet.Meta name again. Ideally, names should match.

BurnzZ: Hmmm, yes, a good point.

One option is to rename the `Meta` class as `PoArgs` to semantically match `request.meta["po_args"]`.

Maybe other names could be `PageArgs`, `PageParams`, and `PoParams`.

Gallaecio: This is something to change in web-poet itself, if we change it at all, right?

We could also use `request.meta["po_meta"]`. I don’t see a big issue with it, I think the repeated `meta` is acceptable in this scenario. We could use `request.meta["po_meta"]`, but we would need to be sure that we will never need additional page object data handled through Scrapy request metadata.

BurnzZ: > This is something to change in web-poet itself, if we change it at all, right?

Yes, [this one](https://github.com/scrapinghub/scrapy-poet/pull/62/files#diff-a38b072945ddeae120d2462383b70ddc0f40967b53c849c49d6a3cbdf1917f53R217-R225) in particular, which is at its core.

> but we would need to be sure that we will never need additional page object data handled through Scrapy request metadata

I think it's safe to say that it will be needed, for cases [like this](https://github.com/scrapinghub/scrapy-poet/pull/62/files#diff-261312b8086ac4bda934c1858d1479ee18c82ba2c22615f1f0ccc798a0dc30b4R162).

Gallaecio: > We could use `request.meta["po_meta"]`, but we would need to be sure that we will never need additional page object data handled through Scrapy request metadata.

I made a typo here :facepalm: . I meant `"po"`. We could use `"po"` if we are sure po-related stuff in `request.meta` will always be limited to `po.Meta` data. Which I would not bet on, even if it seems so now.

scrapy_poet/backend.py
```diff
@@ -0,0 +1,45 @@
...
+            )
+
+        deferred = backend(request)
+        deferred_or_future = maybe_deferred_to_future(deferred)
+        try:
+            response = await deferred_or_future
+        except scrapy.exceptions.IgnoreRequest:
+            raise HttpError(f"Additional request ignored: {request}")
+        except Exception:
+            raise HttpRequestError(f"Additional request failed: {request}")
```
Gallaecio, line 41: I am not sure how to best approach unexpected exceptions. Here I am assuming they are requests errors (twisted exceptions, openssl ones, etc.), but they could as well come from downloader middlewares.

By the way, I thought I would need to handle https://github.com/scrapy/scrapy/blob/master/scrapy/spidermiddlewares/httperror.py#L13 here and map it to `HttpResponseError`, but it turns out it is a spider middleware, not a downloader middleware, so it is not executed for additional requests, and the default status code handling by `HttpClient` works.

BurnzZ: Yeah, this is indeed quite tricky. For example, the retry middleware handles a [bunch of different request errors](https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/retry.py#L131-L134) from twisted but [returns None](https://github.com/scrapy/scrapy/blob/master/scrapy/downloadermiddlewares/retry.py#L124) when max retries are reached.

I wonder if it would be safe to assume that if there was no response (e.g. max retries reached) then we can raise `HttpRequestError`.

scrapy_poet/backend.py
```diff
@@ -0,0 +1,48 @@
...
+        if request.method == "HEAD":
+            request.meta["dont_redirect"] = True
+
+        deferred = backend(request)
+        deferred_or_future = maybe_deferred_to_future(deferred)
+        try:
+            response = await deferred_or_future
+        except scrapy.exceptions.IgnoreRequest:
+            raise HttpError(f"Additional request ignored: {request}")
+        except Exception:
```
kmike, line 43: I wonder if we should use `raise from` when re-raising exceptions, so that the original Scrapy's exception is available as `ex.__cause__`.

I'm a bit confused about it overall, because `ex.__context__` is also set explicitly, but it seems the semantics are a bit different.

BurnzZ:
Hi @Gallaecio , thanks for wrapping this up and bringing it to the finish line! 🙌 LGTM! 🚀

scrapy_poet/backend.py
```diff
@@ -0,0 +1,49 @@
...
+def create_scrapy_backend(backend):
+    async def scrapy_backend(request: HttpRequest):
+        if not isinstance(request, HttpRequest):
+            raise TypeError(
+                f"The request should be 'web_poet.HttpRequest' but received "
+                f"one of type: '{type(request)}'."
+            )
+
+        try:
+            request = scrapy.Request(**attr.asdict(request), dont_filter=True)
```
kmike, line 27: I think it'd be good to create an explicit function to convert from web_poet.HttpRequest to Scrapy's Request, and add tests for it - check that headers work, etc. 
``attr.asdict(request)`` looks a bit risky; it'd also break after we introduce URL classes.

scrapy_poet/backend.py
```diff
@@ -0,0 +1,49 @@
...
+        except TypeError:
+            raise RequestBackendError(
+                f"The given Request isn't compatible with `scrapy.Request`. "
+                f"Ensure that it doesn't contain extra fields: {request}"
```
kmike, line 31: I wonder how can users act on this suggestion

scrapy_poet/backend.py
```diff
@@ -0,0 +1,49 @@
...
+            )
+
+        if request.method == "HEAD":
+            request.meta["dont_redirect"] = True
+
+        deferred = backend(request)
+        deferred_or_future = maybe_deferred_to_future(deferred)
+        try:
+            response = await deferred_or_future
+        except scrapy.exceptions.IgnoreRequest as e:
```
kmike, line 41: Could you please add a comment, explaining when this error can happen?

scrapy_poet/utils.py
```diff
@@ -14,3 +16,15 @@ def get_scrapy_data_path(createdir: bool = True, default_dir: str = ".scrapy") -
     if createdir:
         os.makedirs(path, exist_ok=True)
     return path
+
+
+def scrapy_response_to_http_response(response: Response):
+    """Convenience method to convert a ``scrapy.http.Response`` into a
+    ``web_poet.HttpResponse``.
+    """
+    return HttpResponse(
```
kmike, line 25: This is probably a rare edge case, but it seems the conversion won't work properly for TextResponses where encoding is set explicitly, in `__init__` method. We're not passing this encoding, which means HttpResponse would try to guess the encoding again, and the result could be different.

scrapy_poet/utils.py
```diff
@@ -14,3 +16,15 @@ def get_scrapy_data_path(createdir: bool = True, default_dir: str = ".scrapy") -
     if createdir:
         os.makedirs(path, exist_ok=True)
     return path
+
+
+def scrapy_response_to_http_response(response: Response):
```
kmike, line 21:
```suggestion
def scrapy_response_to_http_response(response: Response) -> HttpResponse:
```

scrapy_poet/backend.py
Gallaecio: Should we expose a way (e.g. a setting) for users to provide their custom implementation of web poet request to Scrapy request conversion?

Or should we maybe reconsider how we handle this? For example, ignore extra fields, possibly with a warning.

kmike: I'm not sure when could the user get this message tbh. web_poet.HttpRequest is a well-defined class, scrapy.Request is a well-defined class, why would there be more than one way to convert between those?

scrapy_poet/page_input_providers.py
```diff
@@ -198,3 +200,26 @@ def deserialize(self, data: Any) -> Sequence[Any]:
...
+
+class MetaProvider(PageObjectInputProvider):
+    """This class provides ``web_poet.page_inputs.Meta`` instances."""
+    provided_classes = {Meta}
+
+    def __call__(self, to_provide: Set[Callable], request: Request):
+        """Creates a ``web_poet.requests.Meta`` instance based on the data found
+        from the ``meta["po_meta"]`` field of a ``scrapy.http.Response`` instance.
+        """
+        return [Meta(**request.meta.get("po_meta", {}))]
```
kmike, line 225: Would this work? I wonder if some of the keys may not be valid argument names.

```suggestion
        return [Meta(request.meta.get("po_meta", {}))]
```

tests/test_backend.py
```diff
@@ -0,0 +1,190 @@
...
+    ) as mock_dtf:
+        mock_dtf.return_value = fake_http_response
+        mock_downloader = mock.MagicMock(return_value=AsyncMock)
+        scrapy_backend = create_scrapy_backend(mock_downloader)
+
+        await scrapy_backend(req)
+
+        args, kwargs = mock_downloader.call_args
+        scrapy_request = args[0]
+        assert scrapy_request.dont_filter is True
```
kmike, line 151: I wonder how hard would it be to have integration-style tests for such features, similar to https://github.com/scrapy-plugins/scrapy-splash/blob/f5273b3a0ef15a9f6809d305a1dbe8f3efd12c5f/tests/test_integration.py#L113. My worry is that Scrapy has many components, which interact in various ways, so it's possible we miss something if we rely on mock-style tests. What do you think, would it worth it?

scrapy_poet/page_input_providers.py
Gallaecio: TIL

```python
>>> dict(1=2)
  File "<stdin>", line 1
SyntaxError: keyword can't be an expression
>>> dict(**{"1": "2"})
{'1': '2'}
```

So I guess we could leave the `**` there, but it is probably better to remove it unless there is a good reason to keep it.

scrapy_poet/backend.py
Gallaecio: I guess we can assume there will be no HttpRequest subclasses requiring special conversion logic. We can always revisit in the future if someone comes up with a valid use case.

docs/intro/advanced-tutorial.rst
kmike: hey! Before we close it off, let's do a final check on the argument/dependency name. It affects both web-poet and scrapy-poet.

I think two main options are:

1. keep web_poet.Meta. This is similar to Scrapy's request.meta, so it can be familiar for developers who worked with Scrapy (?).
2. rename it to something like web_poet.Arguments, web_poet.PageParams, etc. 

After giving it some time to settle, I think I slighly (+0.5) prefer the second option, because of these reasons:

1. In most examples we provide, this argument ("po_meta") is used to control Page behavior. It's some additional per-request parameters, arguments we pass to a page object, which is specific to this page object. So, something like "arguments" or "params" make it more explicit.
2. In general, "meta" or "metadata" is some information about other information - about request in our case. Common examples of metadata are "date created", "author", etc. It's commonly used to aggregate information later, or to provide some additional information about a data point. I'm not sure how common is it to use "meta" or "metadata" to describe arguments which control the behavior; we don't call function parameters "metadata". In Scrapy this happens, but even there meta is used to provide after-the-fact information about the request as well (e.g. download_latency). Also, some uses of Scrapy's meta are deprecated in favor of cb_kwargs.
3. Naming it "meta" could lead to confusion with Scrapy's request.meta. First, user might expect that by depending on web_poet.Meta you'd get access to request.meta, which is not the case. Second, request.meta is mutable - for components like middlewares it can be used both as an input (to control the behavior), and as an output (to provide some information about what happened), while web_poet.Meta is not intended to be mutable, it's only there to be read. So, even for Scrapy users not having "Meta" in the name could allow to reduce the possible confusion.

----

Unrealted to the above, I think we shouldn't use "po" shortcut. We moved away from it (e.g. scrapy-poet used to be called scrapy-po), because it doesn't always sound great in English, some people were saying it like "poo" :)

BurnzZ: You've raised great points @kmike.

I'm inclined to rename `web_poet.Meta` into `web_poet.PageParams`. From the naming semantics, it's quite clear that it contains parameters for a page. In addition, perhaps we could also rename `request.meta["po_args"]` into `request.meta["page_params"]` for symmetry.

tests/test_backend.py
```diff
@@ -0,0 +1,422 @@
...
+    reason=(
+        "Currently, we do not make a distinction between exceptions raised "
+        "from the downloader or from a downloader middleware, except for "
+        "IgnoreRequest. In the future, we might want to inspect the stack to "
+        "determine the source of an exception and raise HttpError instead of "
+        "HttpRequestError when the exception originates in a downloader "
+        "middleware."
+    ),
+    strict=True,
+)
```
Gallaecio, line 371: Is it OK to leave this out of the scope of this pull request, or should I look into addressing this here?

Gallaecio:
Today I (Re-)Learn duplicate filtering happens on the scheduler, so we do not need to use `dont_filter=True` for additional requests because they already bypass the scheduler. Integration tests have been worth it for sure.

docs/intro/advanced-tutorial.rst
```diff
@@ -0,0 +1,168 @@
...
+``web_poet.PageParams``.
+
+Let's see it in action:
+
+.. code-block:: python
+
+    import scrapy
+
+
+    def ProductSpider(scrapy.Spider):
```
kmike, line 146:
```suggestion
    class ProductSpider(scrapy.Spider):
```

scrapy_poet/backend.py
```diff
@@ -0,0 +1,52 @@
...
+
+from scrapy_poet.utils import (
+    http_request_to_scrapy_request,
+    scrapy_response_to_http_response,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def create_scrapy_backend(backend):
```
kmike, line 19: Would it make sense to call the argument as something like `download_func`, not `backend`?

tests/test_backend.py
kmike: I think that's ok.

tests/test_backend.py
```diff
@@ -0,0 +1,441 @@
...
+                yield Request(server.root_url)
+
+            async def parse(self, response, page: ItemPage):
+                item = await page.to_item()
+                items.append(item)
+
+        crawler = make_crawler(TestSpider, {})
+        yield crawler.crawl()
+
+    assert items == [{'foo': 'bar'}]
```
kmike, line 441: it seems this test doesn't check that 2 requests in to_item are made, it could pass if the second (or both) of the requests were no-op, without raising an exception

docs/intro/advanced-tutorial.rst
```diff
@@ -0,0 +1,168 @@
...
+        def start_requests(self):
+            for url in start_urls:
+                yield scrapy.Request(
+                    url=url,
+                    callback=self.parse,
+                    meta={"page_params": {"enable_extracting_all_images": True}}
+                )
+
+        async def parse(self, response, page: ProductPage):
+            return await page.to_item()
```
Gallaecio, line 168: Shouldn’t this be more like:

```python
                    meta={"page_params": {ProductPage: {"enable_extracting_all_images": True}}}
```

So that we can have page-specific params?

```python
                    meta={
                        "page_params": {
                            ProductPage: {"enable_extracting_all_images": True},
                            BrowserHtml: {"geo": "FR"}
                        }
                    }
                )

        async def parse(self, response, html: browserHtml, page: ProductPage):
```

BurnzZ: Yes, this interface makes sense to me. 👍

Gallaecio: After an off-GitHub discussion with @BurnzZ and @kmike, we decided against this because at the moment there is no use case where the same param should have a different value for different page objects for the same request. If that ever changes, maybe we can revisit.

kmike (approved):
👍 +1 to merge it after making sure it works with the latest web-poet (https://github.com/scrapinghub/web-poet/pull/42 is merged).

Gallaecio:
Closing and reopening to trigger tests with the latest web-poet master.

As for merging, we still need to wait for a web-poet release (0.2?), then update requirements here, re-run tests, and then merge.

kmike:
@Gallaecio could you please check https://github.com/scrapinghub/web-poet/pull/51 before making scrapy-poet release with these changes?

Gallaecio:
https://github.com/scrapinghub/scrapy-poet/pull/62/commits/3398dc77f96bfdbffaa104612547252a730f7a51 is not as trivial as the commit message may suggest:
- Added the right minimum versions of dependencies to `setup.py` (lowest versions for which tests pass)
- Had tox test those minimum versions (I have just realized that I need to update GitHub Actions accordingly in a follow-up commit)
- Made Scrapy a dependency in line with https://github.com/scrapy-plugins/scrapy-zyte-api/pull/29
- Removed tldextract as a dependency, not needed since https://github.com/scrapinghub/scrapy-poet/pull/56 (although it is still available, as it is a dependency of Scrapy)
- mypy now runs with the minimum version of Python and dependencies.

tox.ini
```diff
@@ -1,22 +1,45 @@
...
+basepython = python3.7
+commands =
+    {[testenv:asyncio]commands}
+deps =
+    {[testenv:min]deps}
+
 [testenv:mypy]
+basepython = python3.7
 deps =
+    {[testenv:min]deps}
```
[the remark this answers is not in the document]
Gallaecio: My concern was that we could be using typing hints incompatible with the latest Python, but come to think of it, thouse should cause SyntaxError on the regular Python 3.7 tests, so maybe that is fine.

setup.py
```diff
@@ -12,12 +12,12 @@
...
-        'url-matcher',
-        'web-poet @ git+https://git@github.com/scrapinghub/web-poet@master#egg=web-poet',
-        'tldextract',
-        'sqlitedict',
+        'attrs >= 21.3.0',
+        'parsel >= 1.5.0',
+        'scrapy >= 2.6.0',
+        'sqlitedict >= 1.5.0',
+        'url-matcher >= 0.2.0',
+        'web-poet >= 0.2.0',
```
BurnzZ, line 20: As per https://github.com/scrapinghub/web-poet/pull/51#issuecomment-1154937042, we should probably make this into `0.3.0`, as well as updating the CHANGELOG and tox.ini.

BurnzZ (commented):
Got one minor change left and I believe we're good to go. 🚀

setup.py
Gallaecio: Yes, I went through the same path. It was when I went to apply the contextvar name change to the patch when I realized we do not need 0.3.0 here.

kmike:
Thanks @BurnzZ and @Gallaecio, that's a big milestone!

PR #66, opened 2022-03-16
async support for `callback_for`

BurnzZ:
Fixes https://github.com/scrapinghub/scrapy-poet/issues/65. This PR's branch is built on top of https://github.com/scrapinghub/scrapy-poet/pull/62 following `async/await` support on Page Objects.

**Progress:**

- [x] Tests
- [x] Update doc tutorial mentioning `callback_for`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,7 @@ TBR
     * Created new providers for ``web_poet.Meta`` and ``web_poet.HttpClient``.
     * Using the said additional requests needs ``async/await`` support in
       ``asyncio``. This raises the minimum scrapy requirement to ``scrapy>=2.6.0``.
+* add ``async`` support for ``callback_for``.
 
 
 0.3.0 (2022-01-28)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,13 @@ Changelog
 TBR
 ---
 
+* Support for the new features in ``web_poet>=0.2.0`` for supporting additional 
+  requests inside Page Objects:
+
+    * Created new providers for ``web_poet.Meta`` and ``web_poet.HttpClient``.
+    * Using the said additional requests needs ``async/await`` support in
+      ``asyncio``. This raises the minimum scrapy requirement to ``scrapy>=2.6.0``.
+
 * Use the new ``web_poet.HttpResponse`` which replaces ``web_poet.ResponseData``.
 * Support for the new features in ``web_poet>=0.2.0`` for supporting additional 
   requests inside Page Objects:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,13 +5,6 @@ Changelog
 TBR
 ---
 
-* Support for the new features in ``web_poet>=0.2.0`` for supporting additional 
-  requests inside Page Objects:
-
-    * Created new providers for ``web_poet.Meta`` and ``web_poet.HttpClient``.
-    * Using the said additional requests needs ``async/await`` support in
-      ``asyncio``. This raises the minimum scrapy requirement to ``scrapy>=2.6.0``.
-
 * Use the new ``web_poet.HttpResponse`` which replaces ``web_poet.ResponseData``.
 * Support for the new features in ``web_poet>=0.2.0`` for supporting additional 
   requests inside Page Objects:
@@ -19,6 +12,7 @@ TBR
     * Created new providers for ``web_poet.Meta`` and ``web_poet.HttpClient``.
     * Using the said additional requests needs ``async/await`` support in
       ``asyncio``. This raises the minimum scrapy requirement to ``scrapy>=2.6.0``.
+
 * add ``async`` support for ``callback_for``.
 
 
--- a/docs/intro/advanced-tutorial.rst
+++ b/docs/intro/advanced-tutorial.rst
@@ -15,6 +15,7 @@ These are mainly achieved by **scrapy-poet** implementing **providers** for them
     * :class:`scrapy_poet.page_input_providers.HttpClientProvider`
     * :class:`scrapy_poet.page_input_providers.MetaProvider`
 
+.. _`intro-additional-requests`:
 
 Additional Requests
 ===================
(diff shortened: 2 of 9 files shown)


scrapy_poet/api.py
```diff
@@ -90,5 +99,12 @@ def parse(self, response):
     def parse(*args, page: page_cls, **kwargs):  # type: ignore
         yield page.to_item()  # type: ignore
 
-    setattr(parse, _CALLBACK_FOR_MARKER, True)
-    return parse
+    async def async_parse(*args, page: page_cls, **kwargs):  # type: ignore
+        yield await page.to_item()  # type: ignore
+
+    if is_async:
```
proway2, line 105: @BurnzZ thanks! Tested, works well.
I'm wondering if it's possible to refactor it this way:
```python
is_async = inspect.iscoroutinefunction(page_cls.to_item)
```
this way it'd be possible to drop `is_async` from function's signature.

BurnzZ: That's a good point. 👍 I think we can use `inspect.isasyncgenfunction` as well:

```python
import asyncio
import inspect

async def to_item_1():
    yield await asyncio.sleep(1)

async def to_item_2():
    return await asyncio.sleep(1)

inspect.iscoroutinefunction(to_item_1) # False
inspect.iscoroutinefunction(to_item_2) # True

inspect.isasyncgenfunction(to_item_1) # True
inspect.isasyncgenfunction(to_item_2) # False
```

proway2: Yes, may be `iscoroutinefunction` is not suitable here, my point was about if we can automatically determine whether `to_item` should be awaited or not.

BurnzZ: Hi @proway2 , it would seem that `iscoroutinefunction` is indeed enough for the conditional check. :) Thanks for the suggestion again! I've updated the code with this.

docs/intro/basic-tutorial.rst
```diff
@@ -198,6 +198,30 @@ returning the result of the ``to_item`` method call. We could use
     ``response.follow_all(links, callback_for(BookPage))``, without creating
     an attribute, but currently it won't work with Scrapy disk queues.
 
+.. tip::
+
+    :func:`~.callback_for` also supports `async generators` by checking if the
+    ``to_item()`` method is a coroutine. So having the following:
```
Gallaecio, line 204: Maybe we can remove this, as an implementation detail?

```suggestion
    :func:`~.callback_for` also supports `async generators`. So having the
    following:
```

scrapy_poet/api.py
```diff
@@ -90,5 +110,12 @@ def parse(self, response):
...
-    return parse
+    async def async_parse(*args, page: page_cls, **kwargs):  # type: ignore
+        yield await page.to_item()  # type: ignore
+
+    if iscoroutinefunction(page_cls.to_item):
+        setattr(async_parse, _CALLBACK_FOR_MARKER, True)
+        return async_parse
+    else:
+        setattr(parse, _CALLBACK_FOR_MARKER, True)
+        return parse
```
Gallaecio, line 121: :lipstick: Maybe we should define the methods within the `if/else` that uses them, and even replace `if/else` with an early return approach?

CHANGELOG.rst
```diff
@@ -5,13 +5,21 @@ Changelog
...
+      ``asyncio``. This raises the minimum scrapy requirement to ``scrapy>=2.6.0``.
+
 * Use the new ``web_poet.HttpResponse`` which replaces ``web_poet.ResponseData``.
 * Support for the new features in ``web_poet>=0.2.0`` for supporting additional 
   requests inside Page Objects:
 
     * Created new providers for ``web_poet.Meta`` and ``web_poet.HttpClient``.
     * Using the said additional requests needs ``async/await`` support in
       ``asyncio``. This raises the minimum scrapy requirement to ``scrapy>=2.6.0``.
+* add ``async`` support for ``callback_for``.
```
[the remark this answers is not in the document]
BurnzZ: Good catch! Fixed this in https://github.com/scrapinghub/scrapy-poet/pull/66/commits/3983ed6f1d8315c9f593f27bffe5f79a3dbb4cf7.

docs/intro/advanced-tutorial.rst
```diff
@@ -15,6 +15,7 @@ These are mainly achieved by **scrapy-poet** implementing **providers** for them
     * :class:`scrapy_poet.page_input_providers.HttpClientProvider`
     * :class:`scrapy_poet.page_input_providers.MetaProvider`
 
+.. _`intro-additional-requests`:
```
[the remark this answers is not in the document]
BurnzZ: I've added this to the checklist in https://github.com/scrapinghub/scrapy-poet/pull/68 so that we can cover all bases prior to a release. Updating the docs now would affect the other PRs awaiting to be merged.

scrapy_poet/api.py
BurnzZ: I'm not quite sure about this, as having `parse` and `async_parse` in an earlier section, we'd be able to provide some better structured notes about it. For example, this [comment block](https://github.com/scrapinghub/scrapy-poet/pull/66/files#diff-dc2cce7daeaceab39155f854a698ced1d50a1bb16f5cff8af127d8659aeed50aR104-R109).

Regarding the early return approach, that's a good idea. Updated that in https://github.com/scrapinghub/scrapy-poet/pull/66/commits/f1126fb5c1862718e053714ae980672d0420ce8a

docs/intro/basic-tutorial.rst
```diff
@@ -198,6 +198,30 @@ returning the result of the ``to_item`` method call. We could use
...
+    would result in:
+
+    .. code-block:: python
+
+        async def parse_book(self, response: DummyResponse, page: BookPage):
+            yield await page.to_item()
+
+    This is useful when the Page Objects uses additional requests which relies
+    heavily on ``async/await`` format. More info on this in this tutorial section:
+    :ref:`intro-additional-requests`.
```
Gallaecio, line 223:
```suggestion
    This is useful when the Page Objects uses additional requests, which rely
    heavily on ``async/await`` syntax. More info on this in this tutorial 
    section: :ref:`intro-additional-requests`.
```

Gallaecio (approved):

gatufo (approved):
LGTM!

kmike:
Thanks @BurnzZ!

PR #67, opened 2022-03-28
Use the new HttpResponse that replaces ResponseData in web_poet

BurnzZ:
Reference PR from `web_poet`: https://github.com/scrapinghub/web-poet/pull/30.

This also uses the small enhancement in https://github.com/scrapinghub/web-poet/pull/33.

### Checklist before release:

- [ ] Remove references repo in `setup.py` and `tox.ini` that prevents breaking the CI on this PR

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -40,8 +40,14 @@ that need it, like the ``web_poet.ItemWebPage``.
 
         def __call__(self, to_provide: Set[Callable], response: Response):
             """Build a ``web_poet.HttpResponse`` instance using a Scrapy ``Response``"""
-            return [web_poet.HttpResponse(url=response.url, body=response.body)]
-
+            return [
+                HttpResponse(
+                    url=response.url,
+                    body=response.body,
+                    status=response.status,
+                    headers=web_poet.HttpResponseHeaders.from_bytes(response.headers),
+                )
+            ]
 
 You can implement your own providers in order to extend or override current
 ``scrapy-poet`` behavior. All providers should inherit from this base class:
@@ -76,7 +82,14 @@ would lead to the following code:
 
         def __call__(self, to_provide: Set[Callable], response: Response):
             """Build a ``web_poet.HttpResponse`` instance using a Scrapy ``Response``"""
-            return [web_poet.HttpResponse(url=response.url, body=response.body)]
+            return [
+                HttpResponse(
+                    url=response.url,
+                    body=response.body,
+                    status=response.status,
+                    headers=web_poet.HttpResponseHeaders.from_bytes(response.headers),
+                )
+            ]
 
         def fingerprint(self, to_provide: Set[Callable], request: Request) -> str:
             """Returns a fingerprint to identify the specific request."""
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -20,33 +20,27 @@ Creating providers
 ==================
 
 Providers are responsible for building dependencies needed by Injectable
-objects. A good example would be the ``ResponseDataProvider``,
-which builds and provides a ``ResponseData`` instance for Injectables
-that need it, like the ``ItemWebPage``.
+objects. A good example would be the ``HttpResponseProvider``,
+which builds and provides a ``web_poet.HttpResponse`` instance for Injectables
+that need it, like the ``web_poet.ItemWebPage``.
 
 .. code-block:: python
 
     import attr
     from typing import Set, Callable
 
+    import web_poet
     from scrapy_poet.page_input_providers import PageObjectInputProvider
     from scrapy import Response
 
 
-    @attr.define
-    class ResponseData:
-        """Represents a response containing its URL and HTML content."""
-        url: str
-        html: str
-
-
-    class ResponseDataProvider(PageObjectInputProvider):
-        """This class provides ``web_poet.page_inputs.ResponseData`` instances."""
-        provided_classes = {ResponseData}
+    class HttpResponseProvider(PageObjectInputProvider):
+        """This class provides ``web_poet.HttpResponse`` instances."""
+        provided_classes = {web_poet.HttpResponse}
 
         def __call__(self, to_provide: Set[Callable], response: Response):
-            """Build a ``ResponseData`` instance using a Scrapy ``Response``"""
-            return [ResponseData(url=response.url, html=response.text)]
+            """Build a ``web_poet.HttpResponse`` instance using a Scrapy ``Response``"""
+            return [web_poet.HttpResponse(url=response.url, body=response.body)]
 
 
 You can implement your own providers in order to extend or override current
@@ -61,7 +55,7 @@ Cache Suppport in Providers
 ===========================
 
 ``scrapy-poet`` also supports caching of the provided dependencies from the
-providers. For example, :class:`~.ResponseDataProvider` supports this right off
+providers. For example, :class:`~.HttpResponseProvider` supports this right off
 the bat. It's able to do this by inheriting the :class:`~.CacheDataProviderMixin`
 and implementing all of its ``abstractmethods``.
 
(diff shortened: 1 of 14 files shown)


docs/providers.rst
```diff
@@ -20,33 +20,27 @@ Creating providers
...
-        provided_classes = {ResponseData}
+    class HttpResponseProvider(PageObjectInputProvider):
+        """This class provides ``web_poet.HttpResponse`` instances."""
+        provided_classes = {web_poet.HttpResponse}
 
         def __call__(self, to_provide: Set[Callable], response: Response):
-            """Build a ``ResponseData`` instance using a Scrapy ``Response``"""
-            return [ResponseData(url=response.url, html=response.text)]
+            """Build a ``web_poet.HttpResponse`` instance using a Scrapy ``Response``"""
+            return [web_poet.HttpResponse(url=response.url, body=response.body)]
```
kmike, line 43: we can use status and headers here

scrapy_poet/page_input_providers.py
```diff
@@ -178,7 +178,14 @@ def fingerprint(self, to_provide: Set[Callable], request: Request) -> str:
...
 
     def deserialize(self, data: Any) -> Sequence[Any]:
-        return [ResponseData(**response_data) for response_data in data]
+        return [HttpResponse(**response_data) for response_data in data]
+
+    @staticmethod
+    def _to_dict(data: Any) -> dict:
+        return attr.asdict(
+            data,
+            filter=lambda attrib, _: not attrib.name.startswith("_")
```
kmike, line 190: Is it filtering out the encoding argument? I'm not sure it'd work properly. For example, if headers suggest cp1252 encoding, but user passes utf-8 explicitly, encoding would be set to utf-8. It seems this won't survive the serialize/deserialize roundtrip.

tests/test_middleware.py
```diff
@@ -198,7 +198,7 @@ def __call__(self, to_provide):
         # we're returning a class that's not listed in self.provided_classes
         return {
             ExtraClassData: ExtraClassData("this should be returned"),
-            ResponseData: ResponseData("example.com", "this shouldn't"),
+            HttpResponse: HttpResponse("example.com", "this shouldn't"),
```
kmike, line 201: body type should be bytes, it raises an exception if str is passed: https://github.com/scrapinghub/web-poet/blob/256a0c3af97d9d91a3e2150b461d2d723b375f32/tests/test_page_inputs.py#L73

docs/providers.rst
BurnzZ: Nice catch! done in https://github.com/scrapinghub/scrapy-poet/pull/67/commits/64263a9ab64e0d672eae10132f506ff911f6eacd

scrapy_poet/page_input_providers.py
```diff
@@ -154,15 +154,15 @@ def has_cache_support(self):
...
 
-    provided_classes = {ResponseData}
+    provided_classes = {HttpResponse}
     name = "response_data"
 
     def __call__(self, to_provide: Set[Callable], response: Response):
-        """Builds a ``ResponseData`` instance using a Scrapy ``Response``"""
-        return [ResponseData(url=response.url, html=response.text)]
+        """Builds a ``HttpResponse`` instance using a Scrapy ``Response``"""
+        return [HttpResponse(url=response.url, body=response.body)]
```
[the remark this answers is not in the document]
BurnzZ: Nice catch! done in https://github.com/scrapinghub/scrapy-poet/pull/67/commits/64263a9ab64e0d672eae10132f506ff911f6eacd

scrapy_poet/page_input_providers.py
BurnzZ: Good point. Updated this on https://github.com/scrapinghub/scrapy-poet/pull/67/commits/64263a9ab64e0d672eae10132f506ff911f6eacd

tests/test_middleware.py
BurnzZ: Good catch! Updated this on https://github.com/scrapinghub/scrapy-poet/pull/67/commits/64263a9ab64e0d672eae10132f506ff911f6eacd

docs/providers.rst
```diff
@@ -20,34 +20,34 @@ Creating providers
...
-        """Represents a response containing its URL and HTML content."""
-        url: str
-        html: str
-
-
-    class ResponseDataProvider(PageObjectInputProvider):
-        """This class provides ``web_poet.page_inputs.ResponseData`` instances."""
-        provided_classes = {ResponseData}
+    class HttpResponseProvider(PageObjectInputProvider):
+        """This class provides ``web_poet.HttpResponse`` instances."""
```
kmike, line 38: It could be good to also have a way to provide HttpResponseBody, HttpResponseHeaders, and similar classes.

We may even go a step further, and add HttpResponseUrl and HttpResponseStatus classes; in this case, a provider for HttpResponse won't be needed at all, as andi would be able to construct it from other providers.
We don't have to do it now :)

BurnzZ: Hi @kmike , to clarify, do you mean adding providers for `HttpResponseBody` and `HttpResponseHeaders` inside the docs as an example? or actually adding such providers in **providers.py**?

For the latter, I'm not quite sure what would be the use case of having individual providers for each subcomponent of `HttpResponse` as it's already encompassed within the class. Although I might be missing something here.

kmike: hey @BurnzZ! I was thinking about having providers in providers.py. The use case is to support POs which define HttpResponseBody dependencies, not HttpResponse. It doesn't make much sense here, but such POs may be written to use other sources of inputs (e.g. Zyte API), and in this case the distinction could be more important. For example, it's possible to save a bit of API traffic by not requesting HttpResponseHeaders if they're not needed.

BurnzZ: Gotcha! That makes sense! I've created a new issue in https://github.com/scrapinghub/scrapy-poet/issues/70 to track this idea down the line. :)

docs/providers.rst
```diff
@@ -20,34 +20,34 @@ Creating providers
...
+        provided_classes = {web_poet.HttpResponse}
 
         def __call__(self, to_provide: Set[Callable], response: Response):
-            """Build a ``ResponseData`` instance using a Scrapy ``Response``"""
-            return [ResponseData(url=response.url, html=response.text)]
-
+            """Build a ``web_poet.HttpResponse`` instance using a Scrapy ``Response``"""
+            return [
+                HttpResponse(
```
kmike, line 44:
```suggestion
                web_poet.HttpResponse(
```

docs/providers.rst
```diff
@@ -70,18 +70,26 @@ would lead to the following code:
...
+    class HttpResponseProvider(PageObjectInputProvider, CacheDataProviderMixin):
+        """This class provides ``web_poet.HttpResponse`` instances."""
+        provided_classes = {web_poet.HttpResponse}
 
         def __call__(self, to_provide: Set[Callable], response: Response):
-            """Build a ``ResponseData`` instance using a Scrapy ``Response``"""
-            return [ResponseData(url=response.url, html=response.text)]
+            """Build a ``web_poet.HttpResponse`` instance using a Scrapy ``Response``"""
+            return [
+                HttpResponse(
```
kmike, line 86:
```suggestion
                web_poet.HttpResponse(
```

scrapy_poet/page_input_providers.py
```diff
@@ -181,4 +188,13 @@ def serialize(self, result: Sequence[Any]) -> Any:
         return [attr.asdict(response_data) for response_data in result]
 
     def deserialize(self, data: Any) -> Sequence[Any]:
-        return [ResponseData(**response_data) for response_data in data]
+        return [
+            HttpResponse(
+                response_data["url"],
+                response_data["body"],
+                status=response_data["status"],
+                headers=HttpResponseHeaders.from_bytes_dict(response_data["headers"]),
```
kmike, line 196: Why is HttpResponseHeaders.from_bytes_dict needed here?

kmike (approved):
The PR looks good @BurnzZ!

scrapy_poet/page_input_providers.py
BurnzZ: @kmike It's because Scrapy headers look like this:

```python
scrapy_headers = {
    b"Content-Encoding": [b"gzip", b"br"],
    b"Content-Type": [b"text/html"],
    b"content-length": b"648",
}
```

Reference for the motivation: https://github.com/scrapinghub/web-poet/pull/33#issue-1183279271

BurnzZ:
Thanks for the review @kmike ! 🙏 I haven't ticked off the TODO-list on this PR regarding updating the `setup.py` and `tox.ini` deps. In anycase, the PR in https://github.com/scrapinghub/scrapy-poet/pull/62 is built on top of this one so we could look into it again.

scrapy_poet/page_input_providers.py
kmike: If I'm not wrong, here the result is not coming from Scrapy, it's a part of serialization / deserialization for cache.

BurnzZ: Ahhh I see what you mean. Nice catch @kmike ! 🙌 Fixed this in https://github.com/scrapinghub/scrapy-poet/pull/72

PR #68, opened 2022-03-29
add precommit-hook configuration for linters

BurnzZ:
### TODO before releasing a new version
- [ ] Run `tox -e linters` to fix all formatting issues and address any linting issues

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,6 @@
+[tool.black]
+line-length = 120
+
+[tool.isort]
+profile = "black"
+multi_line_output = 3
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,20 @@
+[flake8]
+ignore =
+    # Refers to the max-line length. Let's suppress the error and simply
+    # let black take care on how it wants to format the lines.
+    E501,
+
+    # Refers to "line break before binary operator".
+    # Similar to above, let black take care of the formatting.
+    W503,
+
+    # Refers to "nnecessary dict call - rewrite as a literal".
+    C408
+
+per-file-ignores =
+    # Ignore "imported but unused" errors in __init__ files, as those imports are there
+    # to expose submodule functions so they can be imported directly from that module
+    scrapy_poet/__init__.py:F401
+
+    # Ignore * imports in these files 
+    scrapy_poet/__init__.py:F403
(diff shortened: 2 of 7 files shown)


BurnzZ:
Hi @kmike @Gallaecio , I've updated this previously draft-PR so it can be included in the upcoming release. :)

.git-blame-ignore-revs
```diff
@@ -0,0 +1,5 @@
+# Contains commits to be ignored due to linting
+
+# https://github.com/scrapinghub/scrapy-poet/pull/68
+58c903617911b3209ad68bfefe3fa1a86be629f4
+e4a1cfb7d1e7e55de39039ee3925be073ae6d412
```
kmike, line 5: flake8-fixing commit is not purely stylistic, it contains functional changes; I'd not exclude it from git blame

example/example/spiders/books_05_1.py
```diff
@@ -12,30 +12,30 @@
 """
 
 import scrapy
-
+from example.autoextract import ProductPage
```
kmike, line 15: I wonder if we can make example imports relative, the way they are placed by black (?) look a bit weird otherwise. This applies to other files here as well.

pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[tool.black]
+line-length = 120
```
kmike, line 2: I know this is bikeshedding, but maybe we should stick to either 79 or 88; those two values seem to be the most common standards. 

https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length motivation seems reasonable to me.

Gallaecio: I have a personal preference of 79, but I gave up imposing it in shared repositories a long time ago, so +1 to any choice.

And I am not sure 79 and 88 are the most common, PyCharm is probably the most popular Python IDE, and it enforces 120 by default, which probably has an effect on users.

Gallaecio (approved):

Gallaecio (approved):

example/example/spiders/books_05_1.py
kmike: > Perhaps what you meant was that the linter (i.e. https://github.com/PyCQA/isort) moved this line from down below to up here?

sorry, that was exactly my concern 

> I think he meant from ...autoextract import ProductPage

yes, that's what I meant :) Sorry folks for an unclear message. It's not a big deal.

PR #71, opened 2022-05-08
fix example in providers.rst

kmike:


BurnzZ (approved):
Thanks for catching this @kmike !

PR #75, opened 2022-06-13
Fix doc examples on callback_for()

BurnzZ:
Built on top of https://github.com/scrapinghub/scrapy-poet/pull/62.

I just noticed that the examples were incorrect or at least was not clear enough.


kmike:
Thanks @BurnzZ!

PR #76, opened 2022-06-15
Add provider for RequestUrl

Gallaecio:
We might want to add 1 more test once we merge https://github.com/scrapinghub/scrapy-poet/pull/62, featuring a page object that uses additional requests, to make sure those work even if the original callback gets a dummy request (they should, but to me it seems worth testing nonetheless).

### DIFF
--- a/scrapy_poet/middleware.py
+++ b/scrapy_poet/middleware.py
@@ -3,7 +3,7 @@ responsible for injecting Page Input dependencies before the request callbacks
 are executed.
 """
 import logging
-from typing import Optional, Type, TypeVar
+from typing import Generator, Optional, Type, TypeVar
 
 from scrapy import Spider, signals
 from scrapy.crawler import Crawler
@@ -56,7 +56,7 @@ class InjectionMiddleware:
         self.injector.close()
 
     @inlineCallbacks
-    def _inject_cb_kwargs(self, request: Request, response: Optional[Response] = None):
+    def _inject_cb_kwargs(self, request: Request, response: Optional[Response] = None) -> Generator[None, None, None]:
         # Find out the dependencies
         final_kwargs = yield from self.injector.build_callback_dependencies(
             request,
@@ -70,7 +70,7 @@ class InjectionMiddleware:
             # TODO: check if all arguments are fulfilled somehow?
 
     @inlineCallbacks
-    def process_request(self, request: Request, spider: Spider) -> Optional[DummyResponse]:
+    def process_request(self, request: Request, spider: Spider) -> Generator[None, None, Optional[DummyResponse]]:
         """This method checks if the request is really needed and if its
         download could be skipped by trying to infer if a ``Response``
         is going to be used by the callback or a Page Input.
@@ -92,7 +92,7 @@ class InjectionMiddleware:
 
     @inlineCallbacks
     def process_response(self, request: Request, response: Response,
-                         spider: Spider) -> Response:
+                         spider: Spider) -> Generator[None, None, Response]:
         """This method fills ``request.cb_kwargs`` with instances for
         the required Page Objects found in the callback signature.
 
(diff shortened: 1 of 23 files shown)


scrapy_poet/middleware.py
```diff
@@ -70,7 +86,7 @@ def process_request(self, request: Request, spider: Spider) -> Optional[DummyRes
         """
         if self.injector.is_scrapy_response_required(request):
             return None
-
+        yield from self._inject_cb_kwargs(request)
```
kmike, line 89: Obviously it was added here for a reason, but wouldn't process_response be called anyways if DummyResponse is returned here?

Gallaecio: Hmm… That makes sense. I think I added that code while trying to figure out why things were not working, and once I figured out the actual reason (did not update `DEFAULT_PROVIDERS`) I did not realize this code was not necessary anymore.

kmike: I'm asking because I'm not sure about the consequences of having it in process_request, maybe we'd get some dependencies too soon. E.g. there is a middleware which drops the request later, but we already called all the providers, which did e.g. external API calls.

tests/test_middleware.py
```diff
@@ -327,6 +340,18 @@ def test_skip_downloads(settings):
...
+@inlineCallbacks
+def test_skip_download_request_url(settings):
+    item, url, crawler = yield crawl_single_item(
+        RequestUrlSpider, ProductHtml, settings)
+    assert isinstance(item['response'], Response) is True
+    assert isinstance(item['response'], DummyResponse) is True
+    assert isinstance(item['url'], RequestUrl)
+    assert str(item['url']) == url
+    assert crawler.stats.get_stats().get('downloader/request_count', 0) == 0
+    assert crawler.stats.get_stats().get('downloader/response_count', 0) == 1
```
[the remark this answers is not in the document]
Gallaecio: I also found it weird, although I am not sure if it should be considered an issue, and how to best approach it.

scrapy_poet/middleware.py
```diff
@@ -72,11 +73,12 @@ def process_request(self, request: Request, spider: Spider) -> Optional[DummyRes
             return None
 
         logger.debug(f"Using DummyResponse instead of downloading {request}")
+        self.crawler.stats.inc_value("downloader/request_count/skipped")
```
kmike, line 76: Is it a general Scrapy issue, that if a response is returned in process_request, a request is not counted? If so, would it make more sense to have a stat like this in Scrapy itself? I'm worried that we're reusing Scrapy's builtin namespace for stats here.

Gallaecio: In Scrapy these stats are set from https://github.com/scrapy/scrapy/blob/4bdaa54af4470582845206a9fa21779f80a4b123/scrapy/downloadermiddlewares/stats.py#L22-L50

So I see no easy way to address this upstream. I am not even sure that it needs addressing, as [I mentioned](https://github.com/scrapinghub/scrapy-poet/pull/76#discussion_r897961022).

However, if the issue is the builtin namespace, and that makes sense, shall we simply switch the namespace to something like `scrapy_poet/skipped_request_count` or `scrapy_poet/request_count/skipped`?

kmike: I think that's fine to use another namespace; I'm not sure about the use case, but and extra stat value won't hurt. 

Something like `scrapy_poet/dummy_response_count`? `scrapy_poet/request_count/skipped` is also ok, but it reads as scrapy-poet skips some requests, while we're increasing it where scrapy-poet is actually handling this request. 

If naming is an issue, we may also skip the stats increase for now, and handle it separately. But anyways, I'm fine with merging it, as soon as Scrapy's namespace is not used :)

kmike (approved):
👍

PR #77, opened 2022-06-17
changelog for 0.4.0 release

kmike:
It seems that's the time to do a next scrapy-poet release.


CHANGELOG.rst
```diff
@@ -2,27 +2,42 @@
...
     * Deprecated ``PerDomainOverridesRegistry`` in lieu of the newer
       ``OverridesRegistry`` which provides a wide variety of features
       for better URL matching.
     * This resuls in a newer format in the ``SCRAPY_POET_OVERRIDES`` setting.
-* Removal of this deprecated module: ``scrapy.utils.reqser``
-
-* add ``async`` support for ``callback_for``.
-
+* Removal the use of ``scrapy.utils.reqser`` deprecated module.
+* The minimum supported Scrapy version is now ``2.6.0``.
```
BurnzZ, line 40: I think it's worth pooling this sentence together with the paragraph above that mentions `web-poet >= 0.2.0`. The motivation for this is that the reader could easily see which dependencies he/she needs to update when using `scrapy-poet == 0.4.0`.

CHANGELOG.rst
```diff
@@ -2,27 +2,42 @@
...
+
+This release is backwards incompatible, following backwards-incompatible
+changes in web-poet 0.2.0.
+
+The main new feature is support for ``web-poet >= 0.2.0``, including
+support for ``async def to_item`` methods, making additional requests
+in the ``to_item`` method, new Page Object dependencies, and the new way
+to configure overrides.
+
+Full list of changes:
```
BurnzZ, line 16: What do you think about splitting the list of changes which pertain to `web-poet` and `scrapy-poet`? So it could look something like the following.

```
Changes in line with ``web-poet==0.2.0``:

* ``web_poet.HttpResponse`` replaces ``web_poet.ResponseData`` as a dependency
  to use.
* Additional requests inside Page Objects: a provider for 
  ``web_poet.HttpClient``, as well as ``web_poet.HttpClient``
  backend implementation, which uses Scrapy downloader.
* Provider for ``web_poet.PageParams`` is added, which uses
  ``request.meta["page_params"]`` value.
* Provider for ``web_poet.RequestUrl`` is added, which allows to access the
  request URL in the page object without triggering the download.
* New ``scrapy_poet/dummy_response_count`` value appears in Scrapy stats;
  it is the number of times ``DummyResponse`` is used instead of downloading
  the response as usual.

Internal changes for ``scrapy-poet``:

* Removal the use of ``scrapy.utils.reqser`` deprecated module.
* ``callback_for`` now supports Page Objects which define ``async def to_item``
  method.
* We have these **backward incompatible** changes since the
  ``web_poet.OverrideRule`` follow a different structure:

    * Deprecated ``PerDomainOverridesRegistry`` in lieu of the newer
      ``OverridesRegistry`` which provides a wide variety of features
      for better URL matching.
    * This resuls in a newer format in the ``SCRAPY_POET_OVERRIDES`` setting.
```

It could be easier to read imho.

CHANGELOG.rst
kmike: Added a section at the bottom, about dependency updates.

CHANGELOG.rst
kmike: A good suggestion, could you please check the updated version?

Gallaecio (approved):
:rocket:

PR #78, opened 2022-07-22
Support retries

Gallaecio:
Before we merge, we need:
- A release of web-poet with Retry support.
- To update `setup.py`, and undo the change in `tox.ini` introducing a web-poet installation from Git.

This implementation maintains support for web-poet 0.2.0. However, it probably makes sense to drop it instead and keep the code simpler.

I also wonder if it makes sense to deprecate `scrapy_poet.middleware` and move its content to `scrapy_poet.downloadermiddlewares`. Although if we drop support for web-poet 0.2.0, merging `scrapy_poet.spidermiddlewares` into `scrapy_poet.middleware` is also a valid option.

### DIFF
--- a/docs/intro/basic-tutorial.rst
+++ b/docs/intro/basic-tutorial.rst
@@ -113,12 +113,15 @@ extract a property from the ``to_item`` method:
 Configuring the project
 =======================
 
-To use ``scrapy-poet``, enable its downloader middleware in ``settings.py``:
+To use ``scrapy-poet``, enable its middlewares in ``settings.py``:
 
 .. code-block:: python
 
     DOWNLOADER_MIDDLEWARES = {
         'scrapy_poet.InjectionMiddleware': 543,
+    }
+    SPIDER_MIDDLEWARES = {
+        "scrapy_poet.RetrySpiderMiddleware": 275,
     }
 
 
--- a/scrapy_poet/spidermiddlewares.py
+++ b/scrapy_poet/spidermiddlewares.py
@@ -1,0 +1,32 @@
+try:
+    from web_poet.exceptions import Retry
+except ImportError:
+    pass
+else:
+    from typing import List, Optional
+
+    from scrapy import Spider
+    from scrapy.downloadermiddlewares.retry import get_retry_request
+    from scrapy.http import Request, Response
+
+
+    class RetrySpiderMiddleware:
+        """Captures :exc:`web_poet.exceptions.Retry` exceptions from spider
+        callbacks, and retries the source request."""
+
+        def process_spider_exception(
+            self,
+            response: Response,
+            exception: BaseException,
+            spider: Spider,
+        ) -> Optional[List[Request]]:
+            if not isinstance(exception, Retry):
+                return None
+            new_request_or_none = get_retry_request(
+                response.request,
+                spider=spider,
+                reason='page_object_retry',
+            )
+            if not new_request_or_none:
+                return []
+            return [new_request_or_none]
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,7 @@ setup(
         "parsel >= 1.5.0",
         "scrapy >= 2.6.0",
         "sqlitedict >= 1.5.0",
+        "twisted >= 18.9.0",
         "url-matcher >= 0.2.0",
         "web-poet >= 0.2.0",
     ],
--- a/tests/test_retries.py
+++ b/tests/test_retries.py
@@ -1,0 +1,174 @@
+from pytest import importorskip
+
+Retry = importorskip("web_poet.exceptions.Retry")
+
+from collections import deque
+
+from pytest_twisted import inlineCallbacks
+from scrapy import Spider
+from web_poet.pages import ItemWebPage
+
+from tests.utils import EchoResource, make_crawler, MockServer
+
+
+@inlineCallbacks
+def test_retry_once():
+    retries = deque([True, False])
+    items = []
+
+    with MockServer(EchoResource) as server:
+
+        class ItemPage(ItemWebPage):
+
+            def to_item(self):
+                if retries.popleft():
+                    raise Retry
+                return {"foo": "bar"}
+
+        class TestSpider(Spider):
+            name = "test_spider"
+            start_urls = [server.root_url]
+
+            custom_settings = {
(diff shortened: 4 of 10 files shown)


scrapy_poet/spidermiddlewares.py
```diff
@@ -0,0 +1,31 @@
...
+            if not isinstance(exception, Retry):
+                return None
+            new_request_or_none = get_retry_request(
+                response.request,
+                spider=spider,
+                reason="page_object_retry",
+            )
+            if not new_request_or_none:
+                return []
+            return [new_request_or_none]
```
Gallaecio, line 31: Funny that we are getting this feature here before upstream Scrapy gets it.

I thought about eventually relying on the Scrapy implementation, e.g. having a `RETRY_EXCEPTIONS` Scrapy setting and extending it with `web_poet.exceptions.Retry` as part of the web-poet configuration, instead of installing a middleware. However, the handling of the retry `reason` I think makes that complicated.

So I think the way to go will be to eventually have projects with both, a `RetrySpiderMiddleware` enabled by Scrapy by default that captures a future Scrapy `Retry` exception that supports a `reason` parameter, and this other `RetrySpiderMiddleware` specifically for web-poet support.

setup.py
```diff
@@ -16,6 +16,7 @@
         "parsel >= 1.5.0",
         "scrapy >= 2.6.0",
         "sqlitedict >= 1.5.0",
+        "twisted >= 18.9.0",
```
Gallaecio, line 19: I realized we are using it directly in scrapy_poet.middlewares, so I think it best to include it here. I used the minimum version from Scrapy, since Scrapy is a dependency as well and using an older version will not be possible, but we only use inlineCallbacks and Deferred, so if it were not for Scrapy we could probably support much older versions.

tests/test_retries.py
```diff
@@ -0,0 +1,175 @@
...
+    retries = deque([True, True, True])
+    items = []
+
+    with MockServer(EchoResource) as server:
+
+        class ItemPage(ItemWebPage):
+            def to_item(self):
+                if retries.popleft():
+                    raise Retry
+                return {"foo": "bar"}
```
kmike, line 110: What do you think about testing a case where to_item always raises Retry error? We must ensure that scrapy-poet doesn't keep retrying indefinitely. Or is the idea to make sure tests don't hang if this is not working?

kmike:
> This implementation maintains support for web-poet 0.2.0. However, it probably makes sense to drop it instead and keep the code simpler.

+1 to do it.

> I also wonder if it makes sense to deprecate scrapy_poet.middleware and move its content to scrapy_poet.downloadermiddlewares. Although if we drop support for web-poet 0.2.0, merging scrapy_poet.spidermiddlewares into scrapy_poet.middleware is also a valid option.

Separating download and spider middlewares sounds good to me.

tests/test_retries.py
Gallaecio: I believe the test that covers giving up effectively does that, even though the code is similar to the other tests. But I can modify the test to do it for real.

scrapy_poet/spidermiddlewares.py
BurnzZ: I'm thinking if `web_poet.exceptions.Retry` should accept `**kwargs` params. This is quite similar to https://github.com/psf/requests/blob/main/requests/exceptions.py#L17. In this way, it doesn't blow up with `TypeError: Retry() takes no keyword arguments`.

In any case, I'm good with having 2 separate middlewares at the moment, just in case the retry behaviors _might_ diverge in the future. 👍

BurnzZ (approved):
Looking good @Gallaecio ! +1 to merge after the tests are fixed.

Gallaecio:
The failing tests (docs) should be fixed once we release web-poet with Retry support.

tests/test_retries.py
```diff
@@ -0,0 +1,199 @@
...
+            start_urls = [server.root_url]
+
+            custom_settings = {
+                "DOWNLOADER_MIDDLEWARES": {
+                    "scrapy_poet.InjectionMiddleware": 543,
+                },
+                "SPIDER_MIDDLEWARES": {
+                    "scrapy_poet.RetrySpiderMiddleware": 275,
+                },
+            }
```
kmike, line 187: What do you think about having a base spider, with name and custom_settings, to reduce a bit code duplication in tests?

tox.ini
```diff
@@ -27,7 +28,7 @@ deps =
     scrapy==2.6.0
     sqlitedict==1.5.0
     url-matcher==0.2.0
-    web-poet==0.2.0
```
kmike, line 30: web-poet 0.4.0 is released

docs/intro/basic-tutorial.rst
```diff
@@ -113,13 +113,16 @@ extract a property from the ``to_item`` method:
...
-To use ``scrapy-poet``, enable its downloader middleware in ``settings.py``:
+To use ``scrapy-poet``, enable its middlewares in ``settings.py``:
 
 .. code-block:: python
 
     DOWNLOADER_MIDDLEWARES = {
         'scrapy_poet.InjectionMiddleware': 543,
     }
+    SPIDER_MIDDLEWARES = {
+        "scrapy_poet.RetrySpiderMiddleware": 275,
```
kmike, line 124: It's a bit unclean that InjectionMiddleware is not named InjectionDownloaderMiddleware, or that RetrySpiderMiddleware is not named RetryMiddleware. But that's probably fine. What do you think?

kmike (approved):
👍

docs/intro/basic-tutorial.rst
Gallaecio: The confusion with the built-in middleware is not that much of a concern to me, I think the different modules make it clear enough.

I do not like that there is no way besides documentation to tell if the injection middleware is a downloader middleware or a spider middleware. In Scrapy built-in middlewares you can at least tell from the module path, but we do not have that here.

But yes, the name I chose breaks consistency, and it is probably OK to name it `RetryMiddleware`. I will do that.

Gallaecio: Oh, I had forgotten that after these changes we _are_ separating them into different files. I am feeling much better with the name change :smile:

kmike:
Thanks @Gallaecio!

PR #82, opened 2022-08-08
update docs/docstrings to consistently use double quotes

BurnzZ:
Very minor.

I noticed this when writing a talk for POs wherein copying code from the official docs renders inconsistent quotes in the code. I've chosen double quotes on this PR since we use Black, making the docs consistent with the code and tests.


kmike (approved):
Thanks @BurnzZ!

Gallaecio (approved):

PR #87, opened 2022-09-06
create ResponseUrlProvider

### LINKED ISSUE #83: Add ResponseUrl provider
A basic implementation should be pretty straightforward. 

But I wonder if current scrapy-poet providers API is powerful enough to support further iterations of ResponseUrl. 
For example, let's say there is a base class defined:

```py
import attrs
from zyte_common_items import Product
from web_poet import ItemPage, field, ResponseUrl

@attrs.define
class ProductPage(ItemPage[Product]):
    response_url: ResponseUrl

    @field
    def url(self):
        return self.response_url
```

Ideally, if HttpResponse is a dependency, its URL should be used. But if BrowserHtml is a dependency, we should use its URL. So, it'd be great if the following works without downloading the page twice (with and without browser) - only browser should be used:

```py
@attrs.define
class ExampleProductPage(ProductPage):
    html: BrowserHtml
```

Solving the case above is not a requirement to have a basic ResponseUrl provider, but if anyone has any thoughts about it, please share :)

BurnzZ:
Partially resolves #83.

The only thing missing here is ensuring that the providers are powerful enough such that having a dependency of `HttpResponse` and `BrowserHtml` would not trigger downloading the page twice (using the `ResponseUrl` from the `BrowserHtml` is preferred if available). However, `web-poet` needs to be updated to add a url component to `BrowserHtml`.

Checklist:
- [x] Basic implementation
- [x] Changelog

### DIFF
--- a/tests/test_middleware.py
+++ b/tests/test_middleware.py
@@ -379,11 +379,11 @@ def test_skip_download_response_url(settings):
 
 
 @attr.s(auto_attribs=True)
-class ResponseUrlPage(ItemPage):
-    url: ResponseUrl
+class ResponseUrlPage(WebPage):
+    response_url: ResponseUrl
 
     def to_item(self):
-        return {"url": self.url}
+        return {"response_url": self.response_url}
 
 
 class ResponseUrlPageSpider(scrapy.Spider):
@@ -401,8 +401,8 @@ def test_skip_download_response_url_page(settings):
     item, url, crawler = yield crawl_single_item(
         ResponseUrlPageSpider, ProductHtml, settings
     )
-    assert tuple(item.keys()) == ("url",)
-    assert str(item["url"]) == url
+    assert tuple(item.keys()) == ("response_url",)
+    assert str(item["response_url"]) == url
     # Even if the spider marked the response with DummyResponse, the response
     # is still needed since ResponseUrl depends on it.
     assert crawler.stats.get_stats().get("downloader/request_count", 0) == 1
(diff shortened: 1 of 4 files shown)


tests/test_middleware.py
```diff
@@ -349,6 +349,66 @@ def test_skip_download_request_url(settings):
...
+
+@inlineCallbacks
+def test_skip_download_response_url(settings):
+    item, url, crawler = yield crawl_single_item(
+        ResponseUrlSpider, ProductHtml, settings
+    )
+    assert isinstance(item["response"], Response) is True
+    # Even if the spider marked the response with DummyResponse, the response
+    # is still needed since ResponseUrl depends on it.
+    assert isinstance(item["response"], DummyResponse) is False
```
Gallaecio, line 373: I am unsure about this, setting `response: DummyResponse` in the signature and getting `response: Response` feels wrong, even if a response is actually acquired to provide `ResponseUrl`.

BurnzZ: I believe this should be fine since it's noted down in the docstrings of the middleware ([code ref](https://github.com/scrapinghub/scrapy-poet/blob/master/scrapy_poet/downloadermiddlewares.py#L72-L78)):
> This method checks if the request is really needed and if its
        download could be skipped by trying to infer if a ``Response``
        is going to be used by the callback or a Page Input.
        If the ``Response`` can be ignored, a ``utils.DummyResponse`` object is
        returned on its place. This ``DummyResponse`` is linked to the original
        ``Request`` instance.

Requesting for a `DummyResponse` isn't actually guaranteed but provided if downloading the request can be avoided.

Gallaecio: OK.

Since setting `response: DummyResponse` is meant to signify that you are not interested in the response object, and we want to eventually allow users to skip the parameter itself from the signature, I guess it does not matter what it contains.

Gallaecio (approved):

kmike:
Thanks @BurnzZ!

PR #88, opened 2022-10-06
Supporting `to_return` in web-poet rules

BurnzZ:
In line with the upcoming development in https://github.com/scrapinghub/web-poet/pull/84.

This is built on top of https://github.com/scrapinghub/scrapy-poet/pull/89 for now which moves away from the deprecated functionalities of **web-poet** as introduced in https://github.com/scrapinghub/web-poet/pull/84.

This also addresses #90 .

# TODO:

- [x] move away from deprecated **web-poet** functionalities
- [x] tests and adjustments for the new `to_return` parameter in web-poet's rules.
    - [x] `callback_for`
    - [x] add test cases when multiple rules for same PO is supported (https://github.com/scrapinghub/web-poet/pull/101)
- [x] deprecate passing tuples in `SCRAPY_POET_OVERRIDES` and the `Registry` with the advent of `to_return`.
- [x] pass mypy
- [x] update setup.py and tox.ini to point to the new web-poet version containing new registry features.

# NOTES:
- A drop in code coverage has been observed since I moved most of test utilities into `scrapy_poet/utilities/` ([reference](https://github.com/scrapinghub/scrapy-poet/pull/88#issuecomment-1323410808)) which did not have any tests to begin with. We can add tests to them in a separate PR.
- Some other things we could do, but they could be placed into another PR to make this cleaner.
  - consider renaming `OverridesRegistry` into `RulesRegistry`
  - consider renaming `SCRAPY_POET_OVERRIDES` into `SCRAPY_POET_RULES`
  - consider renaming `SCRAPY_POET_OVERRIDES_REGISTRY` into `SCRAPY_POET_RULES_REGISTRY`
  - **UPDATE**: PR created in https://github.com/scrapinghub/scrapy-poet/pull/103
- We could also move most of the URL matching functionality of the **scrapy-poet**'s registry into **web-poet**. Let's do this in another PR.
  - **UPDATE**: `web-poet` PR in https://github.com/scrapinghub/web-poet/pull/112
- This doesn't include the functionality of selecting which fields to populate when asking for items. This warrants its own API and would be discussed in a separate PR.
  - **UPDATE**: PR https://github.com/scrapinghub/scrapy-poet/pull/111
- Documentation and tutorial for selecting items in POs and callbacks could be placed in the other PR. We can even include it in the PR for selecting which fields to populate _(see bullet point above)._
  - **UPDATE**: PR created in https://github.com/scrapinghub/scrapy-poet/pull/103


scrapy_poet/overrides.py
```diff
@@ -89,26 +91,27 @@ def from_crawler(cls, crawler: Crawler) -> Crawler:
...
+        logger.debug(f"List of parsed ApplyRules:\n{self.rules}")
 
     def add_rule(self, rule: RuleFromUser) -> None:
         if isinstance(rule, (tuple, list)):
             if len(rule) != 3:
                 raise ValueError(
-                    f"Invalid overrides rule: {rule}. Rules as tuples must have "
+                    f"Invalid rule: {rule}. Rules as tuples must have "
                     f"3 elements: (1) the pattern, (2) the PO class used as a "
                     f"replacement and (3) the PO class to be replaced."
```
BurnzZ, line 106: I'm proposing to remove (or at least set for deprecation) the current feature of declaring the rules via triplets:

```python
"SCRAPY_POET_OVERRIDES": [
    ("toscrape.com", BTSBookListPage, BookListPage),
    ("toscrape.com", BTSBookPage, BookPage),
    ("bookpage.com", BPBookListPage, BookListPage),
    ("bookpage.com", BPBookPage, BookPage)
]
```

This could be a bit confusing to users with the introduction of `to_return` since either `instead_of` and `to_return` can now be `None`.

Approaches:
1. We could continue supporting the triplets with its current behavior but set a deprecation warning each time. If the user wants to use `to_return`, they can use `ApplyRule` to declare it.
2. Fully remove the triplets interface and for the user to declare `ApplyRule`.

I'm a bit inclined with approach 2 since I'm not a fan of the triplet interface. It's not clear which of the 2nd or 3rd element is the one assigned to `use` or `instead_of`.

kmike: I'm fine with doing (2), but before going this way it could make sense to make ApplyRule easier to create.
Currently the example would look like this, right?

```py
("toscrape.com", BTSBookListPage, BookListPage)  # was
ApplyRule(for_patterns=Patterns(["toscrape.com"]), use=BTSBookListPage, instead_of=BookListPage)  # now
```

Most of it is fine, but `for_patterns=Patterns(["toscrape.com"])` is quite a bit worse than `"toscrape.com"`.
I wonder if we should make a shortcut in ApplyRule, so that the following works:

```py
ApplyRule("toscrape.com", use=BTSBookListPage, instead_of=BookListPage)
```

BurnzZ: Good idea! Yes, an attrs converter would solve this nicely. :)

BurnzZ: Implemented in `web-poet`: https://github.com/scrapinghub/web-poet/pull/84/commits/33a48ac967176218dd2552c97e916a9d54279212. 🎉 

I'll mark this on the backlog for now as integrating the new `to_return` takes more priority.

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,231 @@
...
+
+        @handle_urls(URL, priority=600)
+        class SubclassProductPage(ParentProductPage):
+            ...
+
+    """
+    item = yield crawl_item(ParentProduct)
+    assert item == ParentProduct(name="subclass product's name")
+
+
```
BurnzZ, line 123: I wanted to discuss the situation with the three (3) test cases below which are currently failing, as they try to abide the idea from https://github.com/scrapinghub/web-poet/pull/84#discussion_r985853698:
> The idea is that `to_return` passed explicitly to the decorator should take priority over the item class defined by the page object.

We could handle this in **scrapy-poet** by replacing the original item returned by the Page Object with the item class specified in `to_return`. Some examples include:

a. `dict` to item class

```python
# using the rule should return StandaloneProduct
@handle_urls(URL, to_return=StandaloneProduct)
class StandaloneProductPage(ItemPage):
    ... # to_item() returns a dict
```

b. an item class into another item class (typically because of subclasses)

```python
@handle_urls(URL)
class ParentReplacedProductPage(ItemPage[ParentReplacedProduct]):
    ...

# using the rule should return StandaloneProdSubclassReplacedProductuct
@handle_urls(URL, to_return=SubclassReplacedProduct) 
class SubclassReplacedProductPage(ParentReplacedProductPage):
    ... # to_item() still returns ParentReplacedProduct
```

---

However, this comes with some changes to **web-poet**:

### 1. `@handle_urls` may need to be updated to also support `skip_nonitem_fields`

This ensures that the current API in **web-poet** (https://web-poet.readthedocs.io/en/stable/advanced/fields.html#changing-item-type) below:

```python
@attrs.define
class CustomItem:
    name: str

class CustomFooPage(FooPage, Returns[CustomItem], skip_nonitem_fields=True):
    pass
```

is compatible with *changing the item type* with **web-poet**'s `@handle_urls` as well:

```python
@attrs.define
class CustomItem:
    name: str

@handle_urls(URL, returns=CustomItem, skip_nonitem_fields=True)
class CustomFooPage(FooPage):
    pass
```

They do different things:

- changing the return item type directly from the Page Object.
- changing the return item type when using the rules for a given Page Object.

-----

There's also some slight implementation challenges as well:

### 2. Converting objects into another item class

The object to be converted could be instances of `dict`, `dataclasses`, `attrs`, etc. depending on how the developer creates them.

Converting `dict` is straighforward as `item_cls(**item_dict)`.

Converting `dataclasses` and `attrs` instances could use `dataclass.asdict` and `attrs.asdict` respectively. For example:

```python
@attrs.define
class TargetItem:
    x: str
    y: int

@attrs.define
class AttrsItem:
    x: str
    y: int
        
    def do_something(self):
        print("action")

attrs_item = AttrsItem(x="data", y=1)

TargetItem(**attrs.asdict(attrs_item))
# TargetItem(x='data', y=1)
```

The challenge here is when developers use their own classes which makes it hard to determine which are the actual attributes containing data vs helper methods. Item classes should make it easy to derive such things for easier item conversion.

```python
class CustomItem:
    def __init__(self, x: int, y: str):
        self._x = x
        self._y = y
        
    # ... some setters and getters here that act similarly with dataclasses/attrs
        
    @property
    def x(self):
        return self._x
    
    @property
    def y(self):
        return self._y
        
    def do_something(self):
        print("action")
    
custom_item = CustomItem(x="data", y=1)

dir(custom_item)
# [..., 'do_something', 'x', 'y']
```

Some approaches we could do:
- enforce a requirement for items return by Page Objects to be either instances of `dict`, `dataclass`, or `attrs`.
- introduce a parameter which accepts a `Callable` that handles the item conversion. The developer could use this to make their custom items work.

scrapy_poet/api.py
```diff
@@ -104,29 +105,27 @@ def parse(self, response):
...
-    if iscoroutinefunction(page_cls.to_item):
-        setattr(async_parse, _CALLBACK_FOR_MARKER, True)
-        return async_parse
+    if issubclass(page_or_item_cls, ItemPage):
+        # When the callback is used as an instance method of the spider, it expects
+        # to receive 'self' as its first argument. When used as a simple inline
+        # function, it expects to receive a response as its first argument.
+        #
+        # To avoid a TypeError, we need to receive a list of unnamed arguments and
+        # a dict of named arguments after our injectable.
```
kmike, line 114: this comment applies to both code paths; it seem to explain why *args and **kwargs are needed

scrapy_poet/api.py
```diff
@@ -29,8 +29,9 @@ def __init__(self, url: str, request=Optional[Request]):
         super().__init__(url=url, request=request)
 
 
-def callback_for(page_cls: Type[ItemPage]) -> Callable:
-    """Create a callback for an :class:`web_poet.pages.ItemPage` subclass.
+def callback_for(page_or_item_cls: Type[ItemPage]) -> Callable:
```
kmike, line 32: if items are accepted, we should update the type

scrapy_poet/injection.py
```diff
@@ -246,7 +246,37 @@ def build_callback_dependencies(self, request: Request, response: Response):
...
+        final_kwargs = plan.final_kwargs(provider_instances)
+        final_kwargs = yield self.convert_po_into_item(request, final_kwargs)
+        return final_kwargs
+
+    @inlineCallbacks
+    def convert_po_into_item(self, request: Request, final_kwargs: dict):
+        """Given a mapping of callback dependencies, call the ``to_item()``
+        method of a dependency to return the requested item.
+        """
+        callback_deps = andi.inspect(get_callback(request, self.spider))
```
kmike, line 258: hey @BurnzZ! That's an interesting implementation, I haven't thought about it. 
Would it work e.g. to provide Product as a dependency to a Page Object?
I was thining we're going to have a provider for item classes.

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,231 @@
...
+@inlineCallbacks
+def test_item_return_subclass() -> None:
+    """A Page Object should properly inherit the ``Return[ItemType]`` from its
+    parent.
+
+    For the example below, asking for ParentProduct would return ``ParentProduct``
+    from the ``ParentProductPage`` since the rules have equal priority but the
+    ``ParentProductPage`` has been decleared first.
```
kmike, line 98: I think we might need to issue a warning in this case, suggesting to set priority explicitly. What do you think?

tests/test_web_poet_rules.py
kmike: See https://github.com/scrapinghub/scrapy-poet/pull/88#discussion_r993492293.

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,231 @@
...
+    For example, in the code below, the returned item should be ``ReplacedProduct(...)``.
+
+    ..code-block::
+
+        @handle_urls(URL, to_return=ReplacedProduct)
+        class ReplacedProductPage(ItemPage[Product]):
+            ...
+    """
+    item = yield crawl_item(ReplacedProduct)
+    assert item == ReplacedProduct(name="replaced product's name")
```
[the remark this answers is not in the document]
BurnzZ: I see, I got the incorrect idea there. Thanks for clarifying @kmike !

tests/test_web_poet_rules.py
BurnzZ: That's a good idea, as it could remove some confusion in users of why their rules aren't working as expected.

I've been writing tests on this but encountered the problem wherein the tests are flaky since the `ApplyRules` are being loaded in random orders for each run, affecting the expected order of priority should all of them have the default `priority=500`.

I think this could also be a problem for users later on, since the warning message changes each time.

I'll solve this problem first in the registry to make the order of rules deterministic.

scrapy_poet/injection.py
BurnzZ: Hi @kmike , I've attempted to move this logic into a new provider.

For example, let's have the following setup:

```python
@attrs.define
class InjectedInfo(ItemFromPageObject):
    info: str

@handle_urls(URL)
class InjectedInfoPage(ItemPage[ItemFromPageObject]):
    @field
    def info(self) -> str:
        return "injected info"

@attrs.define
class Product:
    name: str
    injected_info: InjectedInfo

@handle_urls(URL)
@attrs.define
class ProductPage(ItemPage[Product]):
    injected_info_item: InjectedInfo

    @field
    def name(self) -> str:
        return "product name"

    @field
    def injected_info(self) -> InjectedInfo:
        return self.injected_info_item
```

Using the above example, we get `to_provide=InjectedInfo` inside the provider.

The next step involves trying to find the PageObject that is able to return the item (to be provided) for the given request URL. This logic is quite straightforward and we could get `InjectedInfoPage` without any problems.

However, the derived `InjectedInfoPage` Page Object is still a class and yet to be instantiated. We can't directly instantiate it within the provider (*and call* `to_item()`) since the Page Object class could have its own dependencies (e.g. `HttpResponse`, `HttpClient`, etc) that needs to be resolved first.

In particular, the `InjectedInfoPage` *could be fully built* on this portion of the **injector.py** module:

```python
def build_instances(self, request: Request, response: Response, plan: andi.Plan):
    # First we build the external dependencies using the providers
    instances = yield from self.build_instances_from_providers(
        request, response, plan
    )

    # All the remaining dependencies are internal so they can be built just
    # following the andi plan.
    for cls, kwargs_spec in plan.dependencies:
        if cls not in instances.keys():
            instances[cls] = cls(**kwargs_spec.kwargs(instances))
    return instances
```

However, `InjectedInfoPage` is not even in the `andi.Plan` yet in the code snippet above. In order to make providers work, I'm considering refactoring **injector.py** such that we're able to make it properly work for our new use case.

Before that I wanted to ask if you see another way of solving this? as I might be missing some key concept in **andi** that could solve this easily.

-----

Another thing, I think we need to define in **web-poet** a marker for items signifying that a Page Object returns them. For example, observe that the `InjectedInfo` inherits from `ItemFromPageObject` from the above examples.

This enables the provider to easily identify if it should handle the dependency or not. For example:

```python
class ItemProvider(PageObjectInputProvider):
    provided_classes = lambda type_: isclass(type_) and issubclass(type_, ItemFromPageObject)
    name = "item"

    # and so on ...
```

However, I'm not to keen on this approach since it creates a poor user experience. There's a need to keep in mind to properly subclass from `ItemFromPageObject`, otherwise, it wouldn't be injected to the PageObject.

An alternative way is by checking out all the rules from the registry to see if the `type_` is indeed declared as `to_return` in any of the rules. However, this could be slow as there could be thousands of rules and we need to check each one of them. The `OverridesRegistry` needs to be refactored to support quick lookups like this.

kmike: Hey!

1. I think you're right, and we'd need to do at least 2 iterations of injection to provide an item. I don't see an easy way around it.
2. I think we should use registry. Can it be called once though, why is the speed a issue? We get all rules with `to_return` which is not None, and that's our `provided_classes`.

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,680 @@
...
+        return self.injected_item
+
+
+@inlineCallbacks
+def test_page_object_with_item_dependency() -> None:
+    """Page Objects with dependencies like item classes should have them resolved
+    by the Page Object assigned in one of the rules' ``use`` parameter.
+    """
+
+    # FIXME: This first test case fails
```
BurnzZ, line 588: Note to self: this is yet to be solved.

tests/utils.py
```diff
@@ -13,6 +16,22 @@
 from tests.mockserver import MockServer
 
 
+def create_scrapy_settings(request):
+    """Default scrapy-poet settings"""
+    s = dict(
+        # collect scraped items to crawler.spider.collected_items
+        ITEM_PIPELINES={
+            "tests.utils.CollectorPipeline": 100,
```
wRAR, line 24: Note that you can use (since Scrapy 2.4, and we require 2.6) type objects instead of full names here.

wRAR:
Continuing the today's discussion: my current implementation of a Scrapy command that takes a PO and returns its dependencies and result (like `crawl_item_and_deps()` from this PR, but integrated into the scrapy.cmdline-provided crawler) uses the following things from this PR: `tests.test_web_poet_rules.spider_for()`, `tests.utils.create_scrapy_settings()` + `tests.test_web_poet_rules.rules_settings()`, `tests.utils.CollectorPipeline`, `tests.utils.InjectedDependenciesCollectorMiddleware`. Its implementation looks like this:

```python
        po = load_object(po_name)
        # here we could also support running a specific spider to apply its custom_settings etc
        spider_cls = spider_for(po)
        self.settings.setdict(additional_settings())
        crawler = Crawler(spider_cls, self.settings)
        self.crawler_process.crawl(crawler, url=url)
        self.crawler_process.start()
        items = crawler.spider.collected_items[0]
        deps = crawler.spider.collected_response_deps[0]
```

So if we keep this implementation, it would need at least `CollectorPipeline`, `InjectedDependenciesCollectorMiddleware` and `spider_for` somewhere in `scrapy_poet`, not in its tests.

scrapy_poet/injection.py
BurnzZ: Updated the implementation to create an `ItemProvider`.

scrapy_poet/injection.py
```diff
@@ -141,34 +142,133 @@ def build_plan(self, request: Request) -> andi.Plan:
...
+                prov_req,
+                is_injectable=is_injectable,
+                externally_provided=self.is_class_provided_by_any_provider,
+            )
+
+            try:
+                instances, provider_instances = yield from self.build_instances(
+                    request, response, sub_plan
+                )
+            except RecursionError:
```
BurnzZ, line 190: I explored an option by storing the past seen plans and checking it if a deadlock has occurred. However, that pollutes the code by passing a lot of list references to the recursive function calls. It also harder to cover and test a lot of edge cases which deadlocks occur on far away items (e.g. 10 nodes).

I tried adding a context manager to decrease the recursion limit via `sys.setrecursionlimit()` but it could be fragile since we need to balance the current recursion depth with the said limit.

Settling on this solution for now as it's simpler and cleaner. I think we can live with the tradeoffs of having a deeper callstack (max default of 1000) as proactively storing and comparing the past seen plans for deadlock detection also consumes memory and compute time.

CHANGELOG.rst
```diff
@@ -2,6 +2,80 @@
...
+      of ``scrapy.crawler.Crawler``.
+    * An item type is now supported by ``scrapy_poet.callback_for`` alongside
+      the usual page objects.
+    * ``scrapy_poet.overrides.OverridesRegistry`` has been overhauled:
+
+        * It is now subclassed from ``web_poet.RulesRegistry``.
+        * It also allows retrieval of rules based on the returned item class.
+        * ``OverridesRegistry`` (alongside ``SCRAPY_POET_OVERRIDES``) won't
+          accept tuples as rules anymore. Only ``web_poet.ApplyRule``
+          instances are allowed.
```
BurnzZ, line 70: This sublist for `OverridesRegistry` will be updated later on as we're considering:
- renaming it to something else
- moving its matching functionality to `web_poet.RulesRegistry`.

scrapy_poet/page_input_providers.py
```diff
@@ -97,25 +105,25 @@ def __call__(self, to_provide, response: Response):
...
         else:
             raise MalformedProvidedClassesError(
-                f"Unexpected type '{type_}' for 'provided_classes' attribute of"
-                f"'{cls}.'. Expected either 'set' or 'callable'"
+                f"Unexpected type '{type_}' for 'provided_classes' attribute of "
+                f"'{self}.'. Expected either 'set' or 'callable'"
             )
 
-    def __init__(self, crawler: Crawler):
+    # FIXME: Can't import the Injector as class annotation due to circular dep.
```
BurnzZ, line 123: We're only using the `.build_instances()` method of the `Injector` here. Perhaps a strong sign that we need to decouple the instance creation and injection functionalities of `Injector`.

scrapy_poet/utils/__init__.py
```diff
@@ -1,16 +1,61 @@
...
+    return HttpResponse(
+        url=response.url,
+        body=response.body,
+        status=response.status,
+        headers=HttpResponseHeaders.from_bytes_dict(response.headers),
+        **kwargs,
+    )
 
 
 class HtmlResource(Resource):
```
Gallaecio, line 61: I wonder if most of the content that was in tests/utils.py should be in its own submodule, e.g. `utils.testing`. For example, you would not expect the same from `utils.create_scrapy_settings` that from `utils.testing.create_scrapy_settings`.

example/example/spiders/books_04_overrides_02.py
```diff
@@ -63,19 +62,10 @@ class BooksSpider(scrapy.Spider):
...
-            ),
-            ApplyRule(
-                for_patterns=Patterns(["bookpage.com"]),
-                use=BPBookPage,
-                instead_of=BookPage,
-            ),
+            ApplyRule("toscrape.com", use=BTSBookListPage, instead_of=BookListPage),
+            ApplyRule("toscrape.com", use=BTSBookPage, instead_of=BookPage),
+            ApplyRule("bookpage.com", use=BPBookListPage, instead_of=BookListPage),
+            ApplyRule("bookpage.com", use=BPBookPage, instead_of=BookPage),
```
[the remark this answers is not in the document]
BurnzZ: TIL that this website exists! Good catch!

We can have this on a separate PR to reduce the cognitive workload needed on this already large PR. Created this issue for tracking: https://github.com/scrapinghub/scrapy-poet/issues/107

scrapy_poet/overrides.py
```diff
@@ -26,39 +24,22 @@ def overrides_for(self, request: Request) -> Mapping[Callable, Callable]:
...
-            ("books.toscrape.com", ISBNBookPage, BookPage),
-
-            # Option 2
-            ApplyRule(
-                for_patterns=Patterns(["books.toscrape.com"]),
-                use=MyBookListPage,
-                instead_of=BookListPage,
-            ),
+            ApplyRule("books.toscrape.com", use=ISBNBookPage, instead_of=BookPage),
+            ApplyRule("books.toscrape.com", use=MyBookListPage, instead_of=BookListPage),
```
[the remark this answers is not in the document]
BurnzZ: I might be missing something here, but what's the motivation for only keeping one of these lines?

CHANGELOG.rst
```diff
@@ -2,6 +2,89 @@
...
+        * It is now subclassed from ``web_poet.RulesRegistry`` which allows
+          outright access to its registry methods.
+        * It now allows retrieval of rules based on the returned item class.
+        * ``OverridesRegistry`` (alongside ``SCRAPY_POET_OVERRIDES``) won't
+          accept tuples as rules anymore. Only ``web_poet.ApplyRule``
+          instances are allowed.
+
+            * This is backward incompabible.
+
+    * New exception: ``scrapy_poet.injector_error.ProviderDependencyDeadlockError``.
```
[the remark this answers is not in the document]
BurnzZ: Good catch. I missed out on the description for this. Added a docstring for the exception itself.

I also think we need to cover it in the release notes, since users might want to be aware to catch or detect such exceptions.

scrapy_poet/utils/__init__.py
BurnzZ: Makes sense. Updated in https://github.com/scrapinghub/scrapy-poet/pull/88/commits/5d2f0f9d081780340467f9f9bdc430feb7df4272. cc @wRAR

example/example/spiders/books_04_overrides_02.py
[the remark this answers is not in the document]
Gallaecio: I’m OK with the URL changes being made separately.

But I think we might want to remove these specific lines (and a similar case somewhere else in the changes) in this PR. I believe they used to exist to showcase how you could use either a class or a tuple to pass rules, and now that you only need a class, rather than converting the tuple ones to classes, we should remove the tuple ones, simplifying the example in the process.

scrapy_poet/overrides.py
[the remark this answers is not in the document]
Gallaecio: https://github.com/scrapinghub/scrapy-poet/pull/88#discussion_r1054536569

scrapy_poet/overrides.py
```diff
@@ -89,35 +70,65 @@ class OverridesRegistry(OverridesRegistryBase):
...
+                r
+                for p in pattern_dupes
+                for r in self.search(for_patterns=p, to_return=rule.to_return)
+            ]
+            warn(
+                f"Similar URL patterns {pattern_dupes} were declared earlier "
+                f"which uses to_return={rule.to_return}. This earlier rule "
+                f"will be used when matching against URLs. Consider updating "
+                f"the priority of these rules: {rules}."
+            )
```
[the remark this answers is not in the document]
Gallaecio: So, if 2 rules are the same but have a different priority value, the warning is not triggered?

example/example/spiders/books_04_overrides_02.py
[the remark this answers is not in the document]
BurnzZ: > I believe they used to exist to showcase how you could use either a class or a tuple to pass rules, and now that you can only use a class, rather than converting the tuple ones to classes,

Yes, that's right.

However, they also showcase that a single PO can be overridden by one or more POs. An example would be from `BookListPage`:
```python 
ApplyRule("toscrape.com", use=BTSBookListPage, instead_of=BookListPage)
ApplyRule("bookpage.com", use=BPBookListPage, instead_of=BookListPage)
```

From all of these example files, only `books_04_overrides_02` has this. Albeit its docstring doesn't explicitly mention it.

scrapy_poet/overrides.py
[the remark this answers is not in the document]
BurnzZ: That's right. 👍 

```python
pattern_dupes = [
    pattern
    for pattern in matched.patterns.values()
    if pattern == rule.for_patterns
]
if pattern_dupes:
    # warning
```

In particular, the current logic that we have matches the `url_matcher.matcher.Patterns` instance which contains the **priority** value as well: `Patterns(include="example.com", exclude=(), priority=500)`.

In other words, the for 2 rules to be consider "dupes", their priority should be the same as well.

In the future, we should create some guidelines/standards for teams to define their priority. For example, a team or org could have the following priority levels:
- **5xx to 6xx** for generic POs
- **7xx to 8xx** for team specific rules
- **9xx to 1xxx** for hotfixes or temporary rules
- and so on ...

CHANGELOG.rst
```diff
@@ -5,7 +5,94 @@ Changelog
...
+
+    @attrs.define
+    class Product:
+        name: str
+        image: Image
+
+    @handle_urls("example.com")
+    @attrs.define
+    class ProductPage(WebPage[Product]):
+        # ✨ NEW: Notice that the PO can ask for items as dependencies.
```
kmike, line 35: let's avoid PO terminology for now

CHANGELOG.rst
```diff
@@ -5,7 +5,94 @@ Changelog
...
+        # ``.to_item()`` method of ``ProductImagePage``.
+        image: Image
+
+        @field
+        def name(self) -> str:
+            return self.css("h1.name ::text").get("")
+
+        @field
+        def image(self) -> Image:
+            return self.image
```
kmike, line 46: is this example working? Isn't self.image redefined to be a method?

kmike: Even if not, probbaly it makes sense to change it, to avoid confusion

CHANGELOG.rst
```diff
@@ -5,7 +5,94 @@ Changelog
...
+
+    class MySpider(scrapy.Spider):
+        name = "myspider"
+        start_urls = ["https://example.com/products/some-product"]
+
+        # ✨ NEW: Notice that we're directly using the item here and not the PO.
```
kmike, line 52: let's avoid PO terminology for now

CHANGELOG.rst
```diff
@@ -5,7 +5,94 @@ Changelog
...
+        def parse(self, response, item: Product):
```
kmike, line 53: we should use DummyResponse here

scrapy_poet/overrides.py
```diff
@@ -89,35 +70,64 @@ class OverridesRegistry(OverridesRegistryBase):
...
+        self.overrides_matcher: Dict[Type[ItemPage], URLMatcher] = defaultdict(
+            URLMatcher
         )
+        self.item_matcher: Dict[Optional[Type], URLMatcher] = defaultdict(URLMatcher)
+        for rule_id, rule in enumerate(self._rules):
+            self.add_rule(rule_id, rule)
+        logger.debug(f"List of parsed ApplyRules:\n{self._rules}")
+
+    def add_rule(self, rule_id: int, rule: ApplyRule) -> None:
+        # A common case when a PO subclasses another one with the same URL
```
kmike, line 84: let's avoid PO terminology for now

tests/test_injection.py
```diff
@@ -116,11 +116,13 @@ def test_constructor(self):
...
 
-        # Asserting that we are not leaking providers references
+        # Asserting that some providers could still have access to some parts
+        # of the injector since the injector is passed when instantiating the
+        # provider.
         weak_ref = weakref.ref(injector.providers[0])
         assert weak_ref()
         del injector
-        assert weak_ref() is None
+        assert weak_ref()
```
kmike, line 125: I'm not sure what's the purpose of this test. Previously, it was checking that objects are deallocated properly. Now it's checking that they are not? :)

example/example/spiders/books_04_overrides_02.py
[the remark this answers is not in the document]
BurnzZ: I still think that we should keep them, since `books_04_overrides_02.py` is the only code example file that showcases multiple overrides for a single PO. For example:

```python
ApplyRule("toscrape.com", use=BTSBookListPage, instead_of=BookListPage)
ApplyRule("bookpage.com", use=BPBookListPage, instead_of=BookListPage)
```

CHANGELOG.rst
BurnzZ: Good catch!

CHANGELOG.rst
BurnzZ: Ohh yeah, good catch!

tests/test_injection.py
BurnzZ: You're right, this test has lost its usefulness when we started to provide the `Injector` directly instead of the `Crawler`. I've removed this test altogether.

CHANGELOG.rst
```diff
@@ -5,7 +5,100 @@ Changelog
 TBR
 ---
 
-* Official support for Python 3.11
+This release enables scrapy-poet to fully support item types as dependencies in
```
Gallaecio, line 8: I think we should normalize “item type” / “item class” usage, i.e. choose one and stick to it (there is an “item class” reference a bit later in these changes to the changelog). I have a slight preference for the latter, but I am OK with either.

CHANGELOG.rst
```diff
@@ -5,7 +5,100 @@ Changelog
...
+    * ``scrapy_poet.overrides.OverridesRegistry`` has been overhauled:
+
+        * It is now subclassed from ``web_poet.RulesRegistry`` which allows
+          outright access to its registry methods.
+        * It now allows retrieval of rules based on the returned item class.
+        * ``OverridesRegistry`` (alongside ``SCRAPY_POET_OVERRIDES``) won't
+          accept tuples as rules anymore. Only ``web_poet.ApplyRule``
+          instances are allowed.
+
+            * This also results from these type aliases to be removed:
```
Gallaecio, line 88:
```suggestion
            * This also results on these type aliases being removed:
```
or 

```suggestion
            * As a result, the following type aliases have been removed:
```

CHANGELOG.rst
```diff
@@ -5,7 +5,100 @@ Changelog
...
+              ``scrapy_poet.overrides.RuleAsTuple`` and
+              ``scrapy_poet.overrides.RuleFromUser``
+            * These changes are backward incompabible.
```
Gallaecio, line 91:
```suggestion
            * These changes are backward incompatible.
```

CHANGELOG.rst
```diff
@@ -5,7 +5,100 @@ Changelog
...
+
+    * New exception: ``scrapy_poet.injector_error.ProviderDependencyDeadlockError``.
+      This is raised when it's not possible to create the dependencies due to
+      a deadlock in their sub-dependencies.
```
Gallaecio, line 95: I think we might want to be explicit in mentioning circular dependencies as the cause of the deadlock. For example:

```suggestion
      This is raised when it's not possible to create the dependencies due to
      a deadlock in their sub-dependencies, e.g. due to a circular dependency
      between page objects.
```

CHANGELOG.rst
BurnzZ: You're right. The same is true for **web-poet** and we'll need to update it as well.

My preference is with "item class" as it's more straightforward of the idea that user-defined classes are used. On the other hand, "item type" could portray the idea of type annotations (e.g. `Union[ProductItem, ReviewItem]`) that _might_ create some confusion.

@kmike @wRAR, what are your thoughts?

kmike: +1 to use "class". Reasoning: https://github.com/scrapinghub/web-poet/pull/110#discussion_r1047496239

BurnzZ: Updated to "item class" after our off-GH discussion.

Gallaecio (approved):
I have no more nitpicks, *amazing* work!

scrapy_poet/overrides.py
```diff
@@ -1,19 +1,17 @@
...
+from web_poet import ItemPage, RulesRegistry
 from web_poet.rules import ApplyRule
 
 logger = logging.getLogger(__name__)
 
-RuleAsTuple = Union[Tuple[str, Type[ItemPage], Type[ItemPage]], List]
-RuleFromUser = Union[RuleAsTuple, ApplyRule]
-
 
 class OverridesRegistryBase(ABC):
```
kmike, line 16: Do we still need this ABC? I think we should either remove it, or update to include all the methods which users need to implement in order for registry to work. Removing might be easier.

BurnzZ: I don't think we'll need it anymore.

I'm planning to remove it in a succeeding PR after I submit another PR which moves most of the registry functionalities to `web-poet`. The `web_poet.rules.Registry` should implement `overrides_for()` and `page_objects_for()` from the start.

I also think the new `registry.py` in https://github.com/scrapinghub/scrapy-poet/pull/103 would shrink a lot.

scrapy_poet/overrides.py
```diff
@@ -89,35 +70,64 @@ class OverridesRegistry(OverridesRegistryBase):
...
+            rules = [
+                r
+                for p in pattern_dupes
+                for r in self.search(for_patterns=p, to_return=rule.to_return)
+            ]
+            warn(
+                f"Similar URL patterns {pattern_dupes} were declared earlier "
+                f"that use to_return={rule.to_return}. The first, highest-priority "
+                f"rule added to SCRAPY_POET_REGISTRY will be used when matching "
+                f"against URLs. Consider updating the priority of these rules: "
```
kmike, line 103: If user uses priorities in handle_urls, would it make this warning go away?

BurnzZ: Yes, similar to https://github.com/scrapinghub/scrapy-poet/pull/88#discussion_r1055156099

scrapy_poet/overrides.py
```diff
@@ -89,35 +70,64 @@ class OverridesRegistry(OverridesRegistryBase):
...
-        self.rules.append(rule)
-        # FIXME: This key will change with the new rule.to_return
-        self.matcher[rule.instead_of].add_or_update(  # type: ignore
-            len(self.rules) - 1, rule.for_patterns
+    def __init__(self, rules: Optional[Iterable[ApplyRule]] = None) -> None:
+        super().__init__(rules=rules)
+        self.overrides_matcher: Dict[Type[ItemPage], URLMatcher] = defaultdict(
+            URLMatcher
         )
+        self.item_matcher: Dict[Optional[Type], URLMatcher] = defaultdict(URLMatcher)
```
kmike, line 78: I think we should rename "item_matcher" to "item_matchers", and "overrides_matcher" to "overrides_matchers". 

I was reading code later, and for a minute thought that self.overrides_matcher is an UrlMatcher instance.

scrapy_poet/overrides.py
```diff
@@ -89,35 +70,64 @@ class OverridesRegistry(OverridesRegistryBase):
...
+                rule_id, rule.for_patterns
+            )
+        if rule.to_return:
+            self.item_matcher[rule.to_return].add_or_update(rule_id, rule.for_patterns)
+
+    # TODO: These URL matching functionalities could be moved to web-poet.
+    # ``overrides_for`` in web-poet could be ``str`` or ``_Url`` subclass.
+
+    def rules_for_url(
+        self, url: str, url_matcher: Dict[Any, URLMatcher]
```
kmike, line 118: url_matcher -> matchers?

scrapy_poet/overrides.py
```diff
@@ -89,35 +70,64 @@ class OverridesRegistry(OverridesRegistryBase):
...
+                result[target] = self._rules[rule_id].use
+        return result
+
+    def overrides_for(self, request: Request) -> Mapping[Callable, Callable]:
+        return self.rules_for_url(request.url, self.overrides_matcher)
+
+    def page_object_for_item(
+        self, url: str, po_cls: Type[Any]
+    ) -> Optional[Callable[..., Any]]:
+        return self.rules_for_url(url, self.item_matcher).get(po_cls)
```
kmike, line 133: Can it be optimized? Currently it runs matcher.match for everything in the registry; it seems running it once, for the item_cls, should be enough.

scrapy_poet/overrides.py
```diff
@@ -89,35 +70,64 @@ class OverridesRegistry(OverridesRegistryBase):
...
-                overrides[instead_of] = self.rules[rule_id].use
-        return overrides
+                result[target] = self._rules[rule_id].use
+        return result
+
+    def overrides_for(self, request: Request) -> Mapping[Callable, Callable]:
+        return self.rules_for_url(request.url, self.overrides_matcher)
+
+    def page_object_for_item(
+        self, url: str, po_cls: Type[Any]
```
kmike, line 131: hm, isn't it item_cls, not po_cls?

scrapy_poet/overrides.py
BurnzZ: I'll address this in a separate PR in `web-poet` which migrates all of these functionalities.

scrapy_poet/overrides.py
BurnzZ: Updated how it's implemented in https://github.com/scrapinghub/web-poet/pull/112 wherein its now returning a `Mapping` and doesn't accept an `item_cls` anymore.

scrapy_poet/overrides.py
BurnzZ: Fixed in https://github.com/scrapinghub/web-poet/pull/112.

scrapy_poet/overrides.py
BurnzZ: Fixed in https://github.com/scrapinghub/web-poet/pull/112

scrapy_poet/overrides.py
BurnzZ: Fixed in https://github.com/scrapinghub/web-poet/pull/112

scrapy_poet/page_input_providers.py
```diff
@@ -263,3 +282,77 @@ class ResponseUrlProvider(PageObjectInputProvider):
...
+        # The key that's used here is the ``scrapy.Request`` instance to ensure
+        # that the cached instances under it are properly garbage collected
+        # after processing such request.
+        self._cached_instances = WeakKeyDictionary()
+
+    def provided_classes(self, cls):
+        """If the item is in any of the ``to_return`` in the rules, then it can
+        definitely provide by using the corresponding page object in ``use``.
+        """
+        return isclass(cls) and self.registry.search(to_return=cls)
```
kmike, line 304: Won't it be expensive to call self.registry.search all the time? It's now O(N), going through all the defined rules. And it also seems that provided_classes is called for each request, if I'm not mistaken.

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,1173 @@
...
+handle_urls(URL, instead_of=RightPage)(LeftPage)
+
+
+@inlineCallbacks
+def test_mutual_overrides() -> None:
+    """Two page objects that override each other should not present any problems.
+
+    In practice, this isn't useful at all.
+
+    TODO: We could present a warning to the user if this is detected.
```
kmike, line 245: Raising an error is also a good option, might be even better. I don't see a use case for this; it seems this can get confusing.

kmike: Though no strong opinion on that.

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,1173 @@
...
+class PrioritySubclassProductPage(PriorityParentProductPage):
+    @field
+    def name(self) -> str:
+        return "priority subclass product name"
+
+
+@inlineCallbacks
+def test_item_return_subclass_priority() -> None:
+    """Same case as with ``test_item_return_subclass()`` but now the parent PO
+    uses a higher priority of 600 than the default 500.
```
kmike, line 452: Should we check the case where the subclass priority is set higher? It seems that's the "happy path", while using a high priority in the base class is an edge case, with little practical application.

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,1173 @@
...
+    """Even if ``@handle_urls`` could derive the value for the ``to_return``
+    parameter when the class inherits from something like ``ItemPage[ItemType]``,
+    any value passed through its ``to_return`` parameter should take precedence.
+
+    Note that that this produces some inconsistencies between the rule's item
+    class vs the class that is actually returned. Using the ``to_return``
+    parameter in ``@handle_urls`` isn't recommended because of this.
+
+    This also causes an ``UndeclaredProvidedTypeError`` since the item provider
+    has received a different type of item class from the page object.
```
kmike, line 492: I think the expected behavior for to_return should be to "lie", and actually return ReplacedProduct, not fail. I'm fine with not supporting it, but then probably we should have an xfail test, not a test which enforces the current UndeclaredProvidedTypeError.

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,1173 @@
...
+    @field
+    def name(self) -> str:
+        return "wubba lubba dub dub"
+
+
+@inlineCallbacks
+def test_item_return_with_overrides() -> None:
+    item, deps = yield crawl_item_and_deps(Morty)
+    assert item == Morty(name="wubba lubba dub dub")
+    assert_deps(deps, {"item": RickSanchezC137Page})
```
kmike, line 611: I assume the "wubba lubba dub dub" Morty is returned because of an override, not because of a priority clash (both pages are registered for Morty item). Would it make sense to check that it's the case, by adding priority to the example, or adding another test with priorities?

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,1173 @@
...
+@handle_urls(URL)
+class PageObjectDependencyPage(ItemPage):
+    async def to_item(self) -> dict:
+        return {"name": "item dependency"}
+
+
+@attrs.define
+class MainProductA:
+    name: str
+    item_from_po_dependency: dict
```
kmike, line 667: I'm not sure we should be supporting dicts as items in injection. There seem to be too many pitfalls.

* ItemPage uses dict by default
* There can be "dict" attributes in general, which are not related at all to items.

So, combination of this could mean that we start to do some weird and unexpected injection. I think we shouldn't inject dicts.

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,1173 @@
...
+    PageObjectCounterMixin.assert_instance_count(1, ProductDeepDependencyPage)
+    PageObjectCounterMixin.assert_instance_count(1, ProductDuplicateDeepDependencyPage)
+    assert ItemDependencyPage.to_item_call_count == 1
+    assert ProductWithItemDepPage.to_item_call_count == 1
+    assert ProductDeepDependencyPage.to_item_call_count == 1
+    assert ProductDuplicateDeepDependencyPage.to_item_call_count == 1
+
+
+@attrs.define
+class ChickenPage:
```
kmike, line 1047: This should probably be named ChickenItem, then same for EggItem

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,1173 @@
...
+    @field
+    def other(self) -> str:
+        return self.other_injected.name
+
+
+@inlineCallbacks
+def test_page_object_with_item_dependency_deadlock(caplog) -> None:
+    """Items with page objects which depend on each other resulting in a deadlock
+    should have a corresponding error raised.
+    """
```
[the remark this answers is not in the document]
kmike: Or maybe even a case when both pages use items as dependencies.

kmike:
@BurnzZ I finished the review! :) The implementation looks good overall. You had a good suggestion to review the tests carefully; that's great that you added many of them!

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,1173 @@
...
+
+    # There should be a warning to the user about clashing rules.
+    rules = [
+        ApplyRule(URL, use=ParentProductPage, to_return=ProductFromParent),
+        ApplyRule(URL, use=SubclassProductPage, to_return=ProductFromParent),
+    ]
+    msg = f"Consider updating the priority of these rules: {rules}"
+
+    with warnings.catch_warnings(record=True) as caught_warnings:
+        item, deps = yield crawl_item_and_deps(ProductFromParent)
```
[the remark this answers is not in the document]
BurnzZ: As per https://github.com/scrapinghub/web-poet/pull/112/files#diff-22f478b54a07b372ea47ead887ca01488362ad501692903a9d303511f9d1fe24R157-R163, it would be during import time when the rules are being added to the registry:

```python
from web_poet import consume_modules, default_registry

consume_modules("package_A", "package_B")

# user warning above since:
# 1. @handle_urls are being executed
# 2. which calls add_rule()
# 3. which in turn checks and raises the user warning
```

The tests are set up in a way such that the registry is created each time, alongside the spider via `crawl_item_and_deps()`. This allows us to test each case individually with abstracted setups.

tests/test_web_poet_rules.py
BurnzZ: Yeah, leaving it as-is for now until users do raise use cases for them.

tests/test_web_poet_rules.py
```diff
@@ -0,0 +1,1173 @@
...
+
+    async def to_item(self) -> ProductFromInjectable:
+        return await item_from_fields(self, item_cls=ProductFromInjectable)
+
+
+@inlineCallbacks
+def test_item_return_from_injectable() -> None:
+    """The case wherein a PageObject inherits directly from ``web_poet.Injectable``
+    should also work, provided that the ``to_item`` method is similar with
+    ``web_poet.ItemPage``:
```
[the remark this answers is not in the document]
BurnzZ: I added them here to function as regression checks to err out if something changes in the future. Leaving it as-is here since we get support on it for free with the current implementation. Although I doubt users are going to have this use case.

scrapy_poet/page_input_providers.py
BurnzZ: Good call!

I've updated the PR https://github.com/scrapinghub/scrapy-poet/pull/103 instead since it's directly using the new `web_poet.RulesRegistry` which offers `O(1)` look-up time. https://github.com/scrapinghub/scrapy-poet/pull/103/commits/3c6fdaec67cf1f386de46ddc1af34671c1ef5546

tests/test_web_poet_rules.py
BurnzZ: https://github.com/scrapinghub/scrapy-poet/pull/88/commits/ef01f11591242309802136185ee0fd5bbe15be4f

scrapy_poet/page_input_providers.py
kmike: It'd be better not to use a private method. We can either introduce another method, or (probably better) optimize .search for this use case.

BurnzZ: > or (probably better) optimize .search for this use case.

That's indeed much better 👍

Optimized the `RulesRegistry.search()` method in https://github.com/scrapinghub/web-poet/pull/112/commits/05a832b82beae4242a3facc1920992cee91e00d2 to have a faster look-up when `to_return` or `instead_of` are used.

tests/test_web_poet_rules.py
BurnzZ: Updated 3 variations of this test with an xfail. https://github.com/scrapinghub/scrapy-poet/pull/88/commits/4d25d8c7c59c9f075238acfcf1fbdb5f37b9c91a

tests/test_web_poet_rules.py
BurnzZ: I've re-written these tests to support 3 different variations. https://github.com/scrapinghub/scrapy-poet/pull/88/commits/e184c6f09d210cdd84b1ae2abd5c9e904d0b4ce3

tests/test_web_poet_rules.py
BurnzZ: Good catch! https://github.com/scrapinghub/scrapy-poet/pull/88/commits/bf9b7bfd93e1687adf74f5522cd1efeab225129a

tests/test_web_poet_rules.py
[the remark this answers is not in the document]
BurnzZ: Good one! The current test only checks if both POs have deadlock due to both of them requiring items.

I've added a new variation which is one PO requires items while the other requires a PO. https://github.com/scrapinghub/scrapy-poet/pull/88/commits/33a03910fd60b183ceaff79a435ab4b2411cc644

If two POs both need each others, it'd be hard to make this mistake since Python would err out one PO as not being defined. If they're defined in separate modules and imports one another, Python would also raise cyclic imports.

tests/test_web_poet_rules.py
BurnzZ: I might be missing something but we're currently not supporting `dict`s to be injected.

`item_from_po_dependency` in this case is populated via the page object dependency as:

```python
@field
async def item_from_po_dependency(self) -> dict:
    return await self.injected_page.to_item()
```

It'd be hard for the `ItemProvider` to identify which PO returns the given `dict` item. We only support the following at the moment:

```python
@attrs.define
class MainProductB:                                                                                                                           
    name: str                                                                                                                                 
    item_dependency: ItemDependency
```

docs/rules-from-web-poet.rst
```diff
@@ -187,21 +203,103 @@ Take note that since ``SCRAPY_POET_OVERRIDES`` is structured as
...
+Registry
+========
+
+As mentioned above, scrapy-poet uses the registry from web-poet called
+:class:`web_poet.rules.RulesRegistry`.
+
+This registry implementation can be changed if needed. A different registry can
+be configured by passing its class path to the ``SCRAPY_POET_REGISTRY`` setting.
+Such registries must be a subclass of :class:`web_poet.rules.RulesRegistry`
+to ensure the expected methods and its types are properly accounted for.
```
kmike, line 305: I don't think it's enough to describe the API. It's not clear which methods are needed, and which are implementation details. A few options:

1. Remove SCRAPY_POET_REGISTRY setting
2. Keep the setting, but remove the detailed notes about the registry ("Such registries must be a subclass of ..."); leave it semi-documented.
3. Provide an official API which users need to implement, probably as an abc class. Or, maybe, make sure that all methods which are not used directly are _underscored.

What do you think? I'm +0.5 to (2).

docs/settings.rst
```diff
@@ -21,26 +21,36 @@ More info on this at this section: :ref:`providers`.
...
 ---------------------
 
+Deprecated. Use ``SCRAPY_POET_RULES`` instead.
+
+SCRAPY_POET_RULES
+-----------------
+
 Default: ``None``
 
 Mapping of overrides for each domain. The format of the such ``dict`` mapping
```
kmike, line 31: Isn't it a list of AppltRule instances, not a mapping of overrides for each domain, and not a dict?

docs/settings.rst
```diff
@@ -21,26 +21,36 @@ More info on this at this section: :ref:`providers`.
...
+SCRAPY_POET_REGISTRY
+--------------------
+
 Defaut: ``None``
 
-Sets an alternative Registry to replace the default :class:`~.OverridesRegistry`.
-To use this, set a ``str`` which denotes the absolute object path of the new
-Registry.
+Sets an alternative Registry to replace the default
+:class:`web_poet.rules.RulesRegistry`. To use this, set a ``str`` which denotes
```
kmike, line 50: If I'm not mistaken, like all other Scrapy settings, the class object could also work, not just a str import path.

docs/rules-from-web-poet.rst
```diff
@@ -187,21 +203,103 @@ Take note that since ``SCRAPY_POET_OVERRIDES`` is structured as
...
+For more information about all the expected behavior for the ``to_return``
+parameter in :class:`web_poet.rules.ApplyRule`, check out web-poet's tutorial
+regarding :ref:`rules-item-class-example`.
+
+
+Registry
+========
+
+As mentioned above, scrapy-poet uses the registry from web-poet called
+:class:`web_poet.rules.RulesRegistry`.
```
kmike, line 300: It seems this might cause some confusion. In web-poet we have RulesRegistry class, and a default instance of this class (web_poet.default_registry). Users can create multiple registries, by creating other instances of RulesRegistry. I think in web-poet [docs](https://web-poet.readthedocs.io/en/stable/page-objects/rules.html) "registry" means "RulesRegistry instance", while here "registry" means "RulesRegistry class".

So, when I see SCRAPY_POET_REGISTRY option, I'd probably expect that it allows to use registries other than default_registry. But what it does is that it allows to load all rules from default_registry (or from SCRAPY_POET_RULES), and then create some other class from these rules, instead of RulesRegstry. 

To avoid the confusion, we should at least rename the option to SCRAPY_POET_REGISTRY_CLASS. But actually I'm not sure what's the use case for this option. I can see use cases for using registries other than default_registry (which is not supported directly - you need to do it through SCRAPY_POET_RULES), but why would users want to change the registry class?

docs/rules-from-web-poet.rst
kmike: Although (1) might be even better; see https://github.com/scrapinghub/scrapy-poet/pull/88/files#r1073900090

docs/rules-from-web-poet.rst
kmike: We may also allow to pass registry instances to SCRAPY_POET_RULES. It might make SCRAPY_POET_REGISTRY redundant (as you can pass an instance of a subclass if needed), and also simplify some of the use cases, with  `default_registry.get_rules()`.

docs/settings.rst
BurnzZ: Good call! https://github.com/scrapinghub/scrapy-poet/pull/88/commits/00d5dd6224dfe247e02cdff5f706dfe201128574

docs/settings.rst
BurnzZ: Good catch! I missed out on this. It should actually be an `Iterable[ApplyRule]`, even from before ([reference](https://github.com/scrapinghub/scrapy-poet/pull/88/files#diff-4906cda8a37b6f984ea889eab98b820f6b9f6908f7994ab26e50adfdeff95004L92)). 

https://github.com/scrapinghub/scrapy-poet/pull/88/commits/00d5dd6224dfe247e02cdff5f706dfe201128574

docs/rules-from-web-poet.rst
BurnzZ: As discussed in the call, it'd be probably best to remove the support for replacing the registry. https://github.com/scrapinghub/scrapy-poet/pull/88/commits/e3b7a8e03c284c605005394caf8e0b13af7bc8d5

docs/rules-from-web-poet.rst
BurnzZ: https://github.com/scrapinghub/scrapy-poet/pull/88#discussion_r1081051564

docs/settings.rst
```diff
@@ -21,26 +21,18 @@ More info on this at this section: :ref:`providers`.
...
-------------------------------
-
-Defaut: ``None``
+Default: ``None``
 
-Sets an alternative Registry to replace the default :class:`~.OverridesRegistry`.
-To use this, set a ``str`` which denotes the absolute object path of the new
-Registry.
+Accepts an ``Iterable[ApplyRule]`` to which sets the rules to use. The default
+is currently  :class:`web_poet.rules.RulesRegistry`.
```
kmike, line 32: > The default is currently  :class:`web_poet.rules.RulesRegistry`

Hm, what does it mean?

scrapy_poet/injection.py
```diff
@@ -139,7 +139,11 @@ def build_plan(self, request: Request) -> andi.Plan:
             callback,
             is_injectable=is_injectable,
             externally_provided=self.is_class_provided_by_any_provider,
-            overrides=self.overrides_registry.overrides_for(request).get,
+            # Ignore the type since andi.plan expects overrides to be
+            # Callable[[Callable], Optional[Callable]] but the registry
+            # returns a more accurate typing for this scenario:
+            # Mapping[Type[ItemPage], Type[ItemPage]]
```
kmike, line 145: Is it a typing issue to be fixed in andi?

scrapy_poet/page_input_providers.py
```diff
@@ -263,3 +283,81 @@ class ResponseUrlProvider(PageObjectInputProvider):
...
+        self.registry = self.injector.registry
+
+        # The key that's used here is the ``scrapy.Request`` instance to ensure
+        # that the cached instances under it are properly garbage collected
+        # after processing such request.
+        self._cached_instances = WeakKeyDictionary()
+
+    def provided_classes(self, cls):
+        """If the item is in any of the ``to_return`` in the rules, then it can
+        definitely provide by using the corresponding page object in ``use``.
```
kmike, line 303:
```suggestion
        """If the item is in any of the ``to_return`` in the rules, then it can
        be provided by using the corresponding page object in ``use``.
```

scrapy_poet/injection.py
BurnzZ: Yes, this part: [https://github.com/scrapinghub/andi/blob/master/andi/andi.py#L131](https://github.com/scrapinghub/andi/blob/master/andi/andi.py#L131)

Specifically, I think andi needs to support the signature of `dict.get()` as well.

docs/settings.rst
BurnzZ: Good catch! Updated in a46b1e2

scrapy_poet/page_input_providers.py
BurnzZ: a46b1e2

tests/test_web_poet_rules.py
kmike: I misread the code, sorry! Makes sense.

tests/test_web_poet_rules.py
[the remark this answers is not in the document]
kmike: Would it make sense to check the for items as well?

```py
@handle_urls(URL)
@attrs.define
class ChickenDeadlockPage(ItemPage[ChickenItem]):
    egg: EggItem

    @field
    def name(self) -> str:
        return "chicken"

    @field
    def other(self) -> str:
        return self.egg.name


@handle_urls(URL)
@attrs.define
class EggDeadlockPage(ItemPage[EggItem]):
    chicken: ChickenItem  # depend on item, not on a page

    @field
    def name(self) -> str:
        return "egg"

    @field
    def other(self) -> str:
        return self.chicken.name
```

kmike (approved):
👍 🎉 🚀

tests/test_web_poet_rules.py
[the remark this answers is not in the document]
BurnzZ: Yes, we have it in https://github.com/scrapinghub/scrapy-poet/pull/88/files#diff-a5ce1f1bde36b639be5be17fed1ed35ffb018c1fefcf4544c0e3fac120ca58ebR1314-R1339.

kmike:
Any idea what's up with the CI @BurnzZ? I tried to restart jobs, but it didn't help. No logs, nothing :)

PR #100, opened 2022-12-11
fix bug by passing spider instance to to_dict( ... ) method of scrapy Request

### LINKED ISSUE #93: HttpResponseProvider.fingerprint() broken?
When trying to run a simple spider with caching enabled I get this exception:

```Python traceback
File ".env/lib/python3.10/site-packages/scrapy_poet/page_input_providers.py", line 184, in fingerprint
  k: str(v) for k, v in request.to_dict().items() if k in request_keys
File ".env/lib/python3.10/site-packages/scrapy/http/request/__init__.py", line 184, in to_dict
  "callback": _find_method(spider, self.callback) if callable(self.callback) else self.callback,
File ".env/lib/python3.10/site-packages/scrapy/http/request/__init__.py", line 208, in _find_method
  raise ValueError(f"Function {func} is not an instance method in: {obj}")
```

So if the request has `callback` set, `to_dict()` expects a spider object which isn't passed here. Not sure whether it worked originally, before #74. It's also possible I'm just doing something wrong.

Note that the comment in `Request.to_dict()` is misleading, but it was always that way since `scrapy.utils.reqser.request_to_dict()` was introduced and is probably a copypasting error from `request_from_dict()`.

A minimal spider:

```Python
import scrapy
from web_poet.pages import ItemWebPage


class BookPage(ItemWebPage):
    def to_item(self):
        return {
            'url': self.url,
        }


class BooksSpider(scrapy.Spider):
    name = 'poet1'
    start_urls = ['http://books.toscrape.com/']

    custom_settings = dict(
        DOWNLOADER_MIDDLEWARES={
            'scrapy_poet.InjectionMiddleware': 543,
        },
        SPIDER_MIDDLEWARES={
            "scrapy_poet.RetryMiddleware": 275,
        }
    )

    def parse_book(self, response, book_page: BookPage):
        yield book_page.to_item()

    def parse(self, response):
        links = response.css('.image_container a')
        yield from response.follow_all(links, self.parse_book)
```

Run it with `-s SCRAPY_POET_CACHE=foo.sqlite`

ankitjavalkar:
Fixes #93

### DIFF
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -16,7 +16,6 @@ import attr
 from scrapy import Request
 from scrapy.crawler import Crawler
 from scrapy.http import Response
-from scrapy.utils.request import request_fingerprint
 from web_poet import (
     HttpClient,
     HttpResponse,
@@ -168,6 +167,13 @@ class HttpResponseProvider(PageObjectInputProvider, CacheDataProviderMixin):
     provided_classes = {HttpResponse}
     name = "response_data"
 
+    def __init__(self, crawler: Crawler):
+        if hasattr(crawler, "request_fingerprinter"):
+            self._fingerprint = crawler.request_fingerprinter.fingerprint
+        else:
+            from scrapy.utils.request import request_fingerprint
+            self._fingerprint = request_fingerprint
+
     def __call__(self, to_provide: Set[Callable], response: Response):
         """Builds a ``HttpResponse`` instance using a Scrapy ``Response``"""
         return [
@@ -186,7 +192,7 @@ class HttpResponseProvider(PageObjectInputProvider, CacheDataProviderMixin):
             k: str(v) for k, v in _request.to_dict().items() if k in request_keys
         }
         fp_data = {
-            "SCRAPY_FINGERPRINT": request_fingerprint(_request),
+            "SCRAPY_FINGERPRINT": self._fingerprint(_request).hex(),
             **request_data,
         }
         return json.dumps(fp_data, ensure_ascii=False, sort_keys=True)
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -169,8 +169,10 @@ class HttpResponseProvider(PageObjectInputProvider, CacheDataProviderMixin):
 
     def __init__(self, crawler: Crawler):
         if hasattr(crawler, "request_fingerprinter"):
+
             def fingerprint(x):
                 return crawler.request_fingerprinter.fingerprint(x).hex()
+
             self._fingerprint = fingerprint
         else:
             from scrapy.utils.request import request_fingerprint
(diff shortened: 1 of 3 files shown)


scrapy_poet/page_input_providers.py
```diff
@@ -179,10 +179,10 @@ def __call__(self, to_provide: Set[Callable], response: Response):
             )
         ]
 
-    def fingerprint(self, to_provide: Set[Callable], request: Request) -> str:
+    def fingerprint(self, to_provide: Set[Callable], request: Request, spider: Spider) -> str:
         request_keys = {"url", "method", "body"}
         request_data = {
-            k: str(v) for k, v in request.to_dict().items() if k in request_keys
+            k: str(v) for k, v in request.to_dict(spider=spider).items() if k in request_keys
```
Gallaecio, line 185: Given the callback is not actually meant to be stored by any of these methods, I am thinking creating a copy of the original request without callback or errback, and using that for serialization purposed, may be best here.

e.g. (untested)

```
_request = request.replace(callback=None, errback=None)
```

Then you use `_request` in the rest of the method instead of `request`.

ankitjavalkar: OK. Makes sense I will make that change. @Gallaecio However, I will still have to pass ```spider``` to the ```to_dict``` method, am I right? In that case do you recommend the argument style used here, or should I use the crawler method passed to the parent object?

Gallaecio: > I will still have to pass spider to the to_dict method, am I right?

[The spider argument is only needed for requests with a callback or errback](https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Request.to_dict), so the idea is not to need to pass it at all after you set those to `None`. I am hoping there will be no exception provided those methods are set to `None`.

ankitjavalkar: Yes, thank you for the suggestion. I have made this change now.

docs/providers.rst
```diff
@@ -91,7 +91,7 @@ would lead to the following code:
                 )
             ]
 
-        def fingerprint(self, to_provide: Set[Callable], request: Request) -> str:
+        def fingerprint(self, to_provide: Set[Callable], request: Request, spider: Spider) -> str:
```
Gallaecio, line 94: No longer needed, right?

scrapy_poet/page_input_providers.py
```diff
@@ -14,7 +14,7 @@
 
 import attr
 from scrapy import Request
-from scrapy.crawler import Crawler
+from scrapy.crawler import Crawler, Spider
```
Gallaecio, line 17: No longer needed either (this is why flake8 fails).

Gallaecio (approved):
Looks good to me.

Please, try to think how you could include a test for the fix. I am OK with merging without a test here if adding one is not trivial. But, if we could have one, that would be awesome.

Gallaecio:
I have included changes to use Scrapy 2.7 centralized request fingerprinting when available, to prevent a warning in the new test otherwise.

wRAR (approved):

kmike:
Thanks @ankitjavalkar and @Gallaecio!

PR #102, opened 2022-12-14
Command for creating test fixtures.

wRAR:
Proof of concept, tied to https://github.com/scrapinghub/web-poet/pull/110 and changed together with it.

Not sure if this needs some more settings/code to work with some customized scrapy_poet projects or not.


wRAR:
The Scrapy command may need a better name.

docs/settings.rst
```diff
@@ -78,3 +78,14 @@ developing Page Objects.
...
 out on sporadic errors.
+
+
+SCRAPY_POET_TESTS_DIR
+---------------------
+
+Default: ``None``
+
+Sets the location where the ``create_po_test`` command creates tests. When set
+to ``None``, uses the ``fixtures`` directory in the current directory.
```
wRAR, line 89: We use `settings.get("SCRAPY_POET_TESTS_DIR", "fixtures")`, does this mean that the default is `"fixtures"`? I don't think so, but I don't think we can make a non-None default here?

docs/testing.rst
```diff
@@ -0,0 +1,32 @@
...
+provided by ``web-poet``.
+
+Configuring the test location
+=============================
+
+The ``SCRAPY_POET_TESTS_DIR`` setting specifies where to create the tests. It
+can be set in the project settings or with the ``-s`` command argument.
+
+Handling time fields
+====================
```
wRAR, line 26: Should we also document this feature in web-poet docs? Or maybe also remove the explanation from here and only say that we set this value?

scrapy_poet/commands.py
```diff
@@ -0,0 +1,115 @@
...
+        )
+
+
+def spider_for(injectable: Type[ItemPage]) -> Type[scrapy.Spider]:
+    class InjectableSpider(scrapy.Spider):
+        name = "injectable"
+
+        url = None
+        custom_settings = {
+            "SCRAPY_POET_PROVIDERS": DEFAULT_PROVIDERS,
```
Gallaecio, line 60: Are we preventing custom providers on purpose?

scrapy_poet/commands.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def additional_settings() -> dict:
+    return {
+        "ITEM_PIPELINES": {
+            SavingPipeline: 100,
+        },
+        "DOWNLOADER_MIDDLEWARES": {
+            SavingInjectionMiddleware: 543,
+        },
```
Gallaecio, line 78: As a side effect of defining these, I take it we override project-configured values. This could be a good thing, but also a bad thing.

I cannot imagine an item pipeline that would be of use here, so I think disabling those could be good.

However, if a Scrapy project is using the scrapy-zyte-smartproxy plugin, I take it this effectively prevents that plugin from being used to download the URL, which I think could be a problem.

If so, we may need to use a more convoluted approach to define the downloader middlewares, i.e. add the middleware to the existing DOWNLOADER_MIDDLEWARES of self.settings below, instead of redefining DOWNLOADER_MIDDLEWARES.

docs/settings.rst
wRAR: My point is we don't have default values for settings (non-None value returned by get() when the user didn't set it) but in this case we handle unset values as set to "fixtures". 

It's different from having a true default value of "fixtures" because you can't change it to None in your settings.py.

Gallaecio: You mean that setting `SCRAPY_POET_TESTS_DIR=None` or `SCRAPY_POET_TESTS_DIR=""` in `settings.py` would break this? If so, we could use `settings.get("SCRAPY_POET_TESTS_DIR") or "fixtures"` instead.

scrapy_poet/commands.py
wRAR: No, this is one of the parts where I'm not sure I'm supporting various customizations well.

wRAR: (this code was combined from existing tests code and some new tests code from the unmerged Kevin's PR and then heavily trimmed down so something was lost or kept where it shouldn't be)

scrapy_poet/commands.py
wRAR: Makes sense.

docs/settings.rst
wRAR: Well, my initial question was only "how to document this properly", as replacing the current doc with just "Default: "fixtures"" sounds nice but will be technically wrong (as, yes, setting SCRAPY_POET_TESTS_DIR=None won't change the setting).

BurnzZ: How about setting `Default: "fixtures"` in the docs and when `SCRAPY_POET_TESTS_DIR=None`, the command fails?

docs/testing.rst
```diff
@@ -0,0 +1,32 @@
...
+
+======================
+Tests for Page Objects
+======================
+
+``web-poet`` provides :ref:`tools for testing Page Objects
+<web-poet:web-poet-testing>`. ``scrapy-poet`` projects can use a Scrapy command
+to easily generate tests::
+
+    scrapy create_po_test my_project.po.MyItemPage http://example.com/page1.html
```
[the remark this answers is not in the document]
BurnzZ: +1 on `savefixture`

docs/testing.rst
```diff
@@ -0,0 +1,32 @@
...
+
+This will request the provided page, create an instance of the provided Page
+Object for this page, request its :meth:`~web_poet.pages.ItemPage.to_item`
+method and save both the Page Object dependencies and the resulting item as a
```
BurnzZ, line 15: Welcome to the `Page Object` naming club! 🎉 I used this naming convention as well before.😄 But @Gallaecio , @kmike , and I came to an agreement to use the lower case on this alongside `item class` in another PR (or was it from a call?).

scrapy_poet/commands.py
BurnzZ: We'd need to support both `DEFAULT_PROVIDERS` and `SCRAPY_POET_PROVIDERS`, similar to https://github.com/scrapinghub/scrapy-poet/blob/master/scrapy_poet/injection.py#L53-L57

docs/settings.rst
wRAR: @BurnzZ I don't think we can see if it's None because the user passed None or because nothing was set?

BurnzZ: I see, I totally missed that. In that case, I think documenting it with `Default: None` would be alright.

scrapy_poet/commands.py
wRAR: I think we can just remove this line so that the project settings are used?

BurnzZ: Yes, that works.

docs/testing.rst
```diff
@@ -0,0 +1,29 @@
...
+
+======================
+Tests for Page Objects
+======================
+
+``web-poet`` provides :ref:`tools for testing page objects
+<web-poet:web-poet-testing>`. ``scrapy-poet`` projects can use a Scrapy command
+to easily generate tests::
+
+    scrapy savefixture my_project.po.MyItemPage http://example.com/page1.html
```
kmike, line 11:
```suggestion
    scrapy savefixture my_project.pages.MyItemPage 'https://example.com/page1.html'
```

see also what we needed to do for Scrapy docs: https://docs.scrapy.org/en/latest/intro/tutorial.html#extracting-data

tests/test_commands.py
```diff
@@ -0,0 +1,47 @@
...
+
+def test_savefixture(tmp_path) -> None:
+    project_name = "foo"
+    cwd = Path(tmp_path)
+    call_scrapy_command(str(cwd), "startproject", project_name)
+    cwd /= project_name
+    type_name = "foo.po.BTSBookPage"
+    (cwd / project_name / "po.py").write_text(
+        """
+import attr
```
BurnzZ, line 23:
```suggestion
import attrs
```

Let's have the tests use the new API.

tests/test_commands.py
```diff
@@ -0,0 +1,47 @@
...
+from web_poet import ResponseUrl
+from web_poet.pages import WebPage
+
+
+@attr.define
```
BurnzZ, line 28:
```suggestion
@attrs.define
```

BurnzZ (approved):

kmike:
hey! What's the behavior for the dependencies which we don't currently support, such as HttpClient?

wRAR:
> hey! What's the behavior for the dependencies which we don't currently support, such as HttpClient? Are we raising an error for them?

We don't check this in scrapy-poet, but the called web-poet code will raise an exception:

> NotImplementedError: Serialization for <class 'web_poet.page_inputs.client.HttpClient'> is not implemented

kmike:
Thanks @wRAR!

PR #103, opened 2022-12-16
Docs and deprecations after supporting to_return=Item

BurnzZ:
Built on top of https://github.com/scrapinghub/scrapy-poet/pull/88. It's separated to keep the original PR clean.

Also uses the branch from https://github.com/scrapinghub/web-poet/pull/112 which contains the new registry features which hereby allows the removal of scrapy-poet's own registry.

# TODO:

- [ ] update `setup.py` and `tox.ini` to point to the new **web-poet** version containing new registry features.

### DIFF
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -301,7 +301,8 @@ class ItemProvider(PageObjectInputProvider):
         """If the item is in any of the ``to_return`` in the rules, then it can
         definitely provide by using the corresponding page object in ``use``.
         """
-        return isclass(cls) and self.registry.search(to_return=cls)
+        # avoid using ``web_poet.RulesRegistry.search()`` since it's O(N)
+        return isclass(cls) and self.registry._item_matchers.get(cls)
 
     def update_cache(self, request: Request, mapping: Dict[Type, Any]) -> None:
         if request not in self._cached_instances:


scrapy_poet/registry.py
```diff
@@ -0,0 +1,150 @@
...
+
+        if rule.instead_of:
+            self.overrides_matcher[rule.instead_of].add_or_update(
+                rule_id, rule.for_patterns
+            )
+        if rule.to_return:
+            self.item_matcher[rule.to_return].add_or_update(rule_id, rule.for_patterns)
+
+    # TODO: These URL matching functionalities could be moved to web-poet.
+    # ``overrides_for`` in web-poet could be ``str`` or ``_Url`` subclass.
```
BurnzZ, line 127: Thoughts:
- `OverridesRegistryBase` won't be needed if the `overrides_for` functionality is moved to **web-poet**.
- We can immediately retire `OverridesRegistryBase` from **scrapy-poet** if that's the case.

BurnzZ: I'm considering on moving this functionality right away in **web-poet** so that **scrapy-poet**'s next release already has all of the essential registry functionalities out of the box.

scrapy_poet/registry.py
```diff
@@ -0,0 +1,149 @@
...
+    def overrides_for(self, request: Request) -> Mapping[Callable, Callable]:
+        """
+        Return a ``Mapping`` (e.g. a ``dict``) with type translation rules.
+        The key is the source type that is wanted to be replaced by
+        value, which is also a type.
+        """
+        pass
+
+
+class OverridesAndItemRegistry(OverridesRegistryBase, RulesRegistry):
```
BurnzZ, line 29: I'm thinking if there's a better name for this, such that it's general enough that we won't deprecate it again when a new feature comes out again, as well as it's distinct enough against `web_poet.RulesRegistry`.

docs/rules-from-web-poet.rst
```diff
@@ -187,21 +203,124 @@ Take note that since ``SCRAPY_POET_OVERRIDES`` is structured as
...
+
+    * Page objects can directly ask for items as dependencies.
+
+      This is made possible by the ``ApplyRule("example.com", use=ProductImagePage,
+      to_return=Image)`` instance created from the ``@handle_urls`` decorator
+      on ``ProductImagePage``.
+
+      The ``Image`` item instance directly comes from ``ProductImagePage`` (by
+      calling its ``.to_item()`` method behind the scenes) and is provided to
+      ``ProductPage``.
```
Gallaecio, line 274: I think it may make sense to drop this point altogether, and simplify the example above accordingly. Page object definition is the scope of web-poet, in scrapy-poet we should probably focus only on what is specific to scrapy-poet.

Similarly, we may be able to simplify this documentation by linking to the upstream web-poet docs for documentation on how to declare the item that a page object returns, and here focus only on how to use those in a Scrapy spider, i.e. simply cover that you can use the returned item as a type hint to get it.

scrapy_poet/downloadermiddlewares.py
```diff
@@ -49,14 +50,25 @@ def __init__(self, crawler: Crawler) -> None:
...
-        self.overrides_registry = create_instance(registry_cls, settings, crawler)
+
+        if settings.get("SCRAPY_POET_OVERRIDES_REGISTRY"):
+            msg = (
+                "The SCRAPY_POET_OVERRIDES_REGISTRY setting is deprecated. "
+                "Use SCRAPY_POET_REGISTRY instead."
+            )
+            warnings.warn(msg, DeprecationWarning, stacklevel=2)
+            registry_cls = load_object(
+                settings.get("SCRAPY_POET_OVERRIDES_REGISTRY", OverridesAndItemRegistry)
```
Gallaecio, line 61: At this point we know it is defined.

```suggestion
                settings["SCRAPY_POET_OVERRIDES_REGISTRY"]
```

scrapy_poet/registry.py
Gallaecio: Given the import path gives additional context, I am fine with: `Registry`, `RuleRegistry`, `RulesRegistry` (a “registry of rules” is a “rule registry”, but for consistency with upstream this name may be best). Prepending `Scrapy` to any of those is also fine by me.

CHANGELOG.rst
```diff
@@ -64,17 +64,21 @@ In line with this, the following changes were made:
...
 
+    * The ``scrapy_poet.injection.Injector``'s attribute and constructor parameter 
+      called ``overrides_registry`` is now simply called ``registry``.
+      This is backwards incompatible.
     * An item type is now supported by ``scrapy_poet.callback_for`` alongside
       the usual page objects. This means that it won't raise a ``TypeError``
       anymore when not passing a subclass of ``web_poet.ItemPage``.
-    * ``scrapy_poet.overrides.OverridesRegistry`` has been overhauled:
+    * ``scrapy_poet.overrides.OverridesRegistry`` has been deprecated and
+      overhauled into ``scrapy_poet.rules.OverridesAndItemRegistry``:
```
Gallaecio, line 74:
```suggestion
      overhauled into ``scrapy_poet.registry.OverridesAndItemRegistry``:
```

scrapy_poet/registry.py
```diff
@@ -0,0 +1,149 @@
...
+    def from_crawler(cls, crawler: Crawler) -> Crawler:
+        if "SCRAPY_POET_OVERRIDES" in crawler.settings:
+            msg = (
+                "The SCRAPY_POET_OVERRIDES setting is deprecated. "
+                "Use SCRAPY_POET_RULES instead."
+            )
+            warn(msg, DeprecationWarning, stacklevel=2)
+            rules = crawler.settings.getlist("SCRAPY_POET_OVERRIDES", [])
+        else:
+            rules = crawler.settings.getlist("SCRAPY_POET_RULES", [])
```
Gallaecio, line 76: I wonder, for these deprecated settings, if we should consider a scenario where both settings are declared. Currently, the properly-named setting is ignored. Maybe the old-named setting should be ignored instead, in that scenario.

scrapy_poet/registry.py
Gallaecio: Sounds good to me.

docs/rules-from-web-poet.rst
BurnzZ: I think it'd be great to still frame both web-poet and scrapy-poet together in the same example. In that way, the user could get the whole picture (if they're new to this) about how both packages interact with one another.

scrapy_poet/registry.py
BurnzZ: Good call! Updated to prioritize the new setting first in https://github.com/scrapinghub/scrapy-poet/pull/103/commits/41412398440ed3a517bb3040483390c16cdbd479

docs/rules-from-web-poet.rst
Gallaecio: Makes sense.

But can we at least remove the code and explanation about page objects supporting items as dependencies? While I see how it is related, I don’t think it is relevant for this section about how Scrapy callbacks can ask for items. And removing it would simplify the section significantly.

BurnzZ: I see your point. Updated it in https://github.com/scrapinghub/scrapy-poet/pull/103/commits/dae69d8ba3e672732cc2105dbca77e8253b33ffe

CHANGELOG.rst
```diff
@@ -73,17 +73,22 @@ In line with this, the following changes were made:
         * ``is_provided()`` is now an instance method instead of a class
           method.
 
+    * The ``scrapy_poet.injection.Injector``'s attribute and constructor parameter 
+      called ``overrides_registry`` is now simply called ``registry``.
+      This is backwards incompatible.
+    * An item type is now supported by ``scrapy_poet.callback_for`` alongside
     * An item class is now supported by ``scrapy_poet.callback_for`` alongside
```
Gallaecio, line 80:
```suggestion
    * An item class is now supported by ``scrapy_poet.callback_for`` alongside
```

scrapy_poet/utils/_deprecation.py
```diff
@@ -0,0 +1,110 @@
...
+from warnings import warn
+
+
+def _clspath(cls, forced=None):
+    if forced is not None:
+        return forced
+    return f"{cls.__module__}.{cls.__name__}"
+
+
+def _create_deprecated_class(
```
kmike, line 11: I think it might be better to use the same function either from scrapy or from web-poet, instead of copying it here. Scrapy's might be a bit better, because it's public.

BurnzZ: Good call! Removed this. https://github.com/scrapinghub/scrapy-poet/pull/103/commits/190e3a67aebff3be29440317b19533a78b98791b

scrapy_poet/registry.py
BurnzZ: https://github.com/scrapinghub/web-poet/pull/112

scrapy_poet/registry.py
BurnzZ: Plot twist: scrapy-poet's registry has been removed in lieu of https://github.com/scrapinghub/web-poet/pull/112.

Ref: https://github.com/scrapinghub/scrapy-poet/pull/103/commits/4c1fa0283eba727c5dcb7c83a61969218ac97c07

docs/rules-from-web-poet.rst
```diff
@@ -187,21 +203,103 @@ Take note that since ``SCRAPY_POET_OVERRIDES`` is structured as
...
-must implement the method :meth:`scrapy_poet.overrides.OverridesRegistryBase.overrides_for`.
-As other Scrapy components, it can be initialized from the ``from_crawler`` class
-method if implemented. This might be handy to be able to access settings, stats,
-request meta, etc.
+Item Returns
+============
+
+scrapy-poet also supports a convenient way of asking for items directly. This
+is made possible by the ``to_return`` parameter of :class:`web_poet.rules.ApplyRule`.
+The ``to_return`` specifies which item a page object is capable of returning for
```
Gallaecio, line 211:
```suggestion
The ``to_return`` parameter specifies which item a page object is capable of returning for
```

or 

```suggestion
``to_return`` specifies which item a page object is capable of returning for
```

Gallaecio (approved):
Looks good to me, pending the web-poet changes.

BurnzZ:
Merging this to https://github.com/scrapinghub/scrapy-poet/pull/88 and will continue future discussions there.

PR #104, opened 2022-12-16
official support for Python 3.11

BurnzZ:
I'm excited to have better error message in our CI as per this new feature: https://docs.python.org/3/whatsnew/3.11.html#pep-657-fine-grained-error-locations-in-tracebacks

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,17 +21,12 @@ jobs:
           toxenv: "min"
         - python-version: "3.7"
           toxenv: "asyncio-min"
-
         - python-version: "3.8"
           toxenv: "py"
         - python-version: "3.9"
           toxenv: "py"
-
         - python-version: "3.10"
           toxenv: "py"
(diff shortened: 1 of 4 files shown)


wRAR (approved):

.github/workflows/test.yml
```diff
@@ -32,6 +32,11 @@ jobs:
         - python-version: "3.10"
           toxenv: "asyncio"
 
+        - python-version: "3.11"
+          toxenv: "py"
+        - python-version: "3.11"
+          toxenv: "asyncio"
```
kmike, line 38: Should we remove 3.10 asyncio env?

BurnzZ: Good call. Synced it up with scrapy's CI setup: https://github.com/scrapy/scrapy/blob/master/.github/workflows/tests-ubuntu.yml#L17-L25

kmike:
Thanks @BurnzZ!

PR #106, opened 2022-12-20
Update behavior when dealing with `request.callback=None`

### LINKED ISSUE #48: Unintended interaction with RobotsTxtMiddleware, FilesPipeline, and ImagePipeline
When `def parse(self, response: DummyResponse):` annotation is used, downloading of robots.txt file is skipped; I'm not sure this is intended.

It seems it is cased by this code: https://github.com/scrapy/scrapy/blob/23cfdb058e80a18ee2b66e1b966355f3aca426d0/scrapy/downloadermiddlewares/robotstxt.py#L65 - RobotsTxtMiddleware uses scrapy.Request without setting a callback, which means "parse" by default, but in fact the callback is unused. Then injector inspects the callback signature and decides that response is not needed, while this signature is irrelevant, as the callback is never called.

BurnzZ:
Resolves #48 .

The implementation of the workaround in **scrapy-poet** might change due to the additional issues discovered mentioned [in this comment](https://github.com/scrapinghub/scrapy-poet/issues/48#issuecomment-1358856408).

TODO:
- [x] Add workarounds for `FilesPipeline` and `ImagePipeline`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,15 +6,15 @@ TBR
 ---
 
 * Some changes in behavior when the callback isn't specified (i.e. ``None``) in
-  a ``Request`` which defaults to the ``parse()`` method.
+  a ``Request`` which defaults to the ``parse()`` method. A ``UserWarning``
+  would be issued if the following are encountered:
 
     * Fixed the issue where downloading the response is skipped when the
       ``parse()`` method is annotated with ``scrapy_poet.DummyResponse``.
       For instance: ``def parse(self, response: DummyResponse)``.
     * The ``InjectionMiddleware`` won't attempt to build any dependencies anymore
-      when ``Request.callback == None`` and the ``parse()`` method has any
-      annotated dependencies.
-    * A ``UserWarning`` would be issued if these things are encountered.
+      when ``Request.callback == None`` and the ``parse()`` method has annotated
+      dependencies that involves any of ``SCRAPY_POET_PROVIDERS``.
 
 * Official support for Python 3.11
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,15 +6,25 @@ TBR
 ---
 
 * Some changes in behavior when the callback isn't specified (i.e. ``None``) in
-  a ``Request`` which defaults to the ``parse()`` method. A ``UserWarning``
-  would be issued if the following are encountered:
+  a :class:`scrapy.http.Request` which defaults to the ``parse()`` method. A
+  ``UserWarning`` would be issued if the following are encountered:
 
     * Fixed the issue where downloading the response is skipped when the
-      ``parse()`` method is annotated with ``scrapy_poet.DummyResponse``.
+      ``parse()`` method is annotated with :class:`scrapy_poet.api.DummyResponse`.
       For instance: ``def parse(self, response: DummyResponse)``.
-    * The ``InjectionMiddleware`` won't attempt to build any dependencies anymore
-      when ``Request.callback == None`` and the ``parse()`` method has annotated
-      dependencies that involves any of ``SCRAPY_POET_PROVIDERS``.
+    * The :class:`scrapy_poet.downloadermiddlewares.InjectionMiddleware` won't
+      attempt to build any dependencies anymore when ``request.callback == None``
+      and the ``parse()`` method has annotated dependencies that involves any of
+      ``SCRAPY_POET_PROVIDERS``.
+
+  This affects the following built-in Scrapy components since they use
+  :class:`scrapy.http.Request` without setting any callback value:
+
+    * :class:`scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware`
+    * :class:`scrapy.pipelines.images.ImagesPipeline`
+    * :class:`scrapy.pipelines.files.FilesPipeline`
+
+  See the new :ref:`pitfalls` documentation for more information.
 
 * Official support for Python 3.11
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,10 @@ Changelog
 TBR
 ---
 
+* Fix interaction with ``scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware``
+  where it could skip downloading the "robots.txt" file if the spider has annotated
+  the ``parse()`` method with ``scrapy_poet.DummyResponse``. For instance:
+  ``def parse(self, response: DummyResponse)``.
 * Official support for Python 3.11
 
 0.6.0 (2022-11-24)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,9 +5,9 @@ Changelog
 TBR
 ---
 
-* Fix interaction with ``scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware``
-  where it could skip downloading the "robots.txt" file if the spider has annotated
-  the ``parse()`` method with ``scrapy_poet.DummyResponse``. For instance:
+* Fix cases where some downloads are skipped for ``Requests`` that doesn't specify
+  a callback which defaults to the ``parse()`` method. It only happens when the
+  ``parse()`` method is annotated with ``scrapy_poet.DummyResponse``. For instance:
   ``def parse(self, response: DummyResponse)``.
 * Official support for Python 3.11
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,9 +5,8 @@ Changelog
 TBR
 ---
 
-* Scrapy uses the ``parse()`` method when dealing with :class:`scrapy.http.Request`
-  instances having a callback of ``None``. On such cases, Scrapy uses the spider's
-  ``parse()`` method as the callback.
+* Scrapy uses the spider's ``parse()`` method when the :class:`scrapy.http.Request`
+  callback is set to ``None``.
 
   There are some built-in Scrapy features and components that do this, as of
   Scrapy 2.7:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,9 +12,9 @@ TBR
       ``parse()`` method is annotated with ``scrapy_poet.DummyResponse``.
       For instance: ``def parse(self, response: DummyResponse)``.
     * The ``InjectionMiddleware`` won't attempt to build any dependencies anymore
-      when ``Request.callback == None`` and the ``parse()`` method is annotated
-      with ``response: DummyResponse``.
-    * A warning would be issued if these things are encountered.
+      when ``Request.callback == None`` and the ``parse()`` method has any
+      annotated dependencies.
+    * A ``UserWarning`` would be issued if these things are encountered.
 
 * Official support for Python 3.11
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,24 +5,79 @@ Changelog
 TBR
 ---
 
-* Some changes in behavior when the callback isn't specified (i.e. ``None``) in
-  a :class:`scrapy.http.Request` which defaults to the ``parse()`` method. A
-  ``UserWarning`` would be issued if the following are encountered:
+* Scrapy uses the ``parse()`` method when dealing with :class:`scrapy.http.Request`
+  instances having a callback of ``None``. On such cases, Scrapy uses the spider's
+  ``parse()`` method as the callback.
 
-    * Fixed the issue where downloading the response is skipped when the
-      ``parse()`` method is annotated with :class:`scrapy_poet.api.DummyResponse`.
-      For instance: ``def parse(self, response: DummyResponse)``.
-    * The :class:`scrapy_poet.downloadermiddlewares.InjectionMiddleware` won't
-      attempt to build any dependencies anymore when ``request.callback == None``
-      and the ``parse()`` method has annotated dependencies that involves any of
-      ``SCRAPY_POET_PROVIDERS``.
+  There are some built-in Scrapy features and components that do this, as of
+  Scrapy 2.7:
 
-  This affects the following built-in Scrapy components since they use
-  :class:`scrapy.http.Request` without setting any callback value:
+    * Using the ``parse()`` method which relies on ``start_urls``. For example:
 
+        .. code-block:: python
+
+            class MySpider(scrapy.Spider):
+                name = "my_spider"
+                start_urls = ["https://books.toscrape.com"]
+
+                def parse(self, response):
+                    ...
+    
     * :class:`scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware`
     * :class:`scrapy.pipelines.images.ImagesPipeline`
     * :class:`scrapy.pipelines.files.FilesPipeline`
+
+  There are some backward incompatible changes when this occurs:
+
+    * When the ``parse()`` method has its response argument annotated with
+      :class:`scrapy_poet.api.DummyResponse`, for instance:
+      ``def parse(self, response: DummyResponse)``, the response is downloaded
+      instead of being skipped.
+
+      This was a necessary change since
+      :class:`scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware` might
+      not work properly and **not** visit the ``robots.txt`` file from a website.
+      Moreover, it avoids the problem of missing images or files when the following
+      pipelines are used:
+
+        * :class:`scrapy.pipelines.images.ImagesPipeline`
+        * :class:`scrapy.pipelines.files.FilesPipeline`
+
+      To avoid this new behavior, the **recommended** approach is to avoid
+      defining a ``parse()`` method and instead choose any other name. 
+
+    * When the ``parse()`` method has dependencies that are provided by
+      any active providers from ``SCRAPY_POET_PROVIDERS``, the
+      :class:`scrapy_poet.downloadermiddlewares.InjectionMiddleware` won't
+      attempt to build any dependencies anymore.
+
+      This causes the following code to have this error ``TypeError: parse()
+      missing 1 required positional argument: 'page'.``:
+
+        .. code-block:: python
+
+            class MySpider(scrapy.Spider):
+                name = "my_spider"
+                start_urls = ["https://books.toscrape.com"]
+
+                def parse(self, response: scrapy.http.Response, page: MyPage):
+                    ...
+        
+      To avoid this issue, you can:
+
+        * avoid defining the ``parse()`` method and instead use any other name.
+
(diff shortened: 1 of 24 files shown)


scrapy_poet/injection.py
```diff
@@ -121,6 +121,10 @@ def is_scrapy_response_required(self, request: Request):
         """
         Check whether the request's response is going to be used.
         """
+        # See: https://github.com/scrapinghub/scrapy-poet/issues/48
+        if request.callback is None and request.url.endswith("robots.txt"):
```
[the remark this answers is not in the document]
BurnzZ: > I'd prefer not to try to fix Scrapy components on-by-one here; it's a losing battle, as there might be third-party and project-specific extensions which use the same approach.

I agree. There's also the notion of unable to upgrade to the latest Scrapy version with the fix.

> Do you think it would work?

(1) and (2) are good ideas.

I've implemented it in this new commit: https://github.com/scrapinghub/scrapy-poet/pull/106/commits/0ce6de8b0fc64b3acc9d4726ef5664e98f28491b

I'm also thinking if we should introduce an escape hatch for this, but I soon realized that the escape hatch itself is to avoid using the `parse()` method completely. The user warning should be enough to inform the user about this.

On the other hand, it'd hard to make them work for your example [from this comment](https://github.com/scrapinghub/scrapy-poet/issues/48#issuecomment-1341061227) since the `Response` would now actuallly be downloaded but would result in a 404, completely ignoring the `parse()` callback. I think this could be okay, since if users are experiencing 404s, then they could:

- add 404s to the allowed HTTP Response status codes, or
- specify an explicit `callback` and `errback` to handle them.

What do you think?

----

(3) and (4) are all good to me. 👍

> One more question: what about page objects? Is this issue only about DummyResponse? For example, if parse method needs a certain Page Object to be created, would robots.txt middleware trigger creation of this Page Object? What if the page object doesn't use HttpResponse dependency?

The injector still creates them as it's being called by `InjectionMiddleware.process_response`. Good call!

Right now, it doesn't present any problems, apart from some minor wasted compute time to build the dependencies which aren't used at all. It's difficult to address this right now without (3) since simply relying on `request.callback == None` raises some potential false positives.

scrapy_poet/injection.py
```diff
@@ -326,6 +327,16 @@ def is_callback_requiring_scrapy_response(callback: Callable):
         return True
 
     if issubclass(first_parameter.annotation, DummyResponse):
+        # See: https://github.com/scrapinghub/scrapy-poet/issues/48
+        if request.callback is None:
+            warnings.warn(
+                f"{request} has callback=None which defaults to the parse() "
```
kmike, line 333: Would this error be logged for every FilePipeline request? What can user do about this warning?

tests/test_downloader.py
```diff
@@ -403,3 +416,106 @@ async def parse(self, response, page: ItemPage):
...
+                    "scrapy_poet.InjectionMiddleware": 543,
+                },
+            }
+
+            def parse(self, response: DummyResponse):
+                collected["response"] = response
+
+        crawler = make_crawler(TestSpider, {})
+
+        with warnings.catch_warnings(record=True) as w:
```
kmike, line 451: can you use pytest.warns?

tests/test_response_required_logic.py
```diff
@@ -185,21 +187,52 @@ def test_is_provider_using_response():
...
     # Callbacks created with the callback_for function won't make use of
     # the response, but their providers might use them.
-    assert is_callback_requiring_scrapy_response(spider.callback_for_parse) is False
+    assert (
+        is_callback_requiring_scrapy_response(spider.callback_for_parse, mock_request)
+        is False
+    )
+
+    # See: https://github.com/scrapinghub/scrapy-poet/issues/48
+    mock_request.callback = None
```
kmike, line 211: can we use a real Request instad of mock?

BurnzZ: Good call! https://github.com/scrapinghub/scrapy-poet/pull/106/commits/e353f81df552fea79d7ec12a394c9b6509bc2db2

tests/test_downloader.py
BurnzZ: https://github.com/scrapinghub/scrapy-poet/pull/106/commits/e353f81df552fea79d7ec12a394c9b6509bc2db2

tests/test_downloader.py
```diff
@@ -286,6 +294,9 @@ class TestSpider(Spider):
                 },
             }
 
+            def start_requests(self):
+                yield Request(server.root_url, self.parse)
```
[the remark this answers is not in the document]
BurnzZ: Good call! I've updated the docs and examples to be in line and consistent with this. https://github.com/scrapinghub/scrapy-poet/pull/106/commits/8efbc929e7a67bab28a6e1a171244ff0e7ad8419

scrapy_poet/injection.py
BurnzZ: Yes, it currently logs everything.

1. I was thinking of moving this `is_callback_requiring_scrapy_response()` function as a method in the `Injector`. In this way, it can have access to a new setting similar to `DUPEFILTER_DEBUG` which could be toggled to show all succeeding duplicate requests.
    - Since we now have access to the `stats`, we could also log the number of `DummyResponse` that were actually downloaded. Though, this would be harder to debug than the actual logged warning messages.

2. A similar approach to (1) but instead of introducing a new setting similar to `DUPEFILTER_DEBUG`, users could manually filter the warning logs via https://docs.scrapy.org/en/latest/topics/logging.html#advanced-customization. This takes more steps than (1).
    - (1) and (2) can be combined as well, where the new setting is available but the user could opt not to use it as they could customize the filtering themselves.

3. Another approach is to subclass `UserWarning` and use that specific warning category to warn the user. They could use `warnings.filterwarnings()` for it but it's a clunky to integrate with a Scrapy project.

scrapy_poet/downloadermiddlewares.py
```diff
@@ -107,6 +107,13 @@ def process_response(
         and an injectable attribute,
         the user-defined ``cb_kwargs`` takes precedence.
         """
+        if request.callback is None:
+            spider.logger.warning(
+                f"{request} has callback=None. No dependencies will be built "
+                f"by scrapy-poet."
+            )
```
Gallaecio, line 114: Instead of warning here, could we move this to `process_request`, and only warn the first time if the `parse` method has something that would cause action from the injection middleware? (i.e. DummyResponse or something that would be subject to be resolved in process_response)

BurnzZ: I've updated this part to only warn once when the `parse()` method has been annotated with `DummyResponse`.

I think we should keep the warning here in `process_response()` since there could be custom middlewares in some project that tries to replace the `callback=None` with something else.

scrapy_poet/injection.py
BurnzZ: After our call, I've updated this to only warn once by having the same message (i.e. rewording the warning message to remove the request from it).

scrapy_poet/downloadermiddlewares.py
```diff
@@ -107,6 +108,19 @@ def process_response(
...
         """
+        if request.callback is None:
+            callback = get_callback(request, spider)  # should return parse()
+            response_annotation = get_response_annotation(callback)
+
+            if issubclass(response_annotation.annotation, DummyResponse):
+                warnings.warn(
+                    "A request has been encountered with callback=None which "
+                    "defaults to the parse() method. On such cases, when the "
+                    "parse() method is annotated with DummyResponse, "
```
kmike, line 119: It's not about DummyResponse only, right?

BurnzZ: Do you on its subclasses as well? Good catch! Updated in https://github.com/scrapinghub/scrapy-poet/pull/106/commits/59076de43c8d71e58dcedeb58a07bb7dc3c0a18a

Gallaecio: What we discussed off GitHub was to discourage the use of `parse` as a callback that gets injected stuff, and to warn if `parse` has _anything_ that would require injection, be it DummyResponse, HttpResponse, or anything else.

By the way, to do it properly, I am thinking that we need to ensure we do it not only for callback=None, but also for callback='parse' and callback=spider.parse.

kmike: As @Gallaecio said, it's about anything we cn provide, not just DummyResponse.

> By the way, to do it properly, I am thinking that we need to ensure we do it not only for callback=None, but also for callback='parse' and callback=spider.parse.

Why would callback=spider.parse or "parse" be an issue? If user wants to use this callback explicitly, would it be possible to provide the dependencies, without any warnings?

Gallaecio: OK, so we would not be warning against the use of resolvable parameters in `parse` _in general_, but only when there is at least one detected use of callback=None. Works for me.

BurnzZ: > OK, so we would not be warning against the use of resolvable parameters in parse in general, but only when there is at least one detected use of callback=None. Works for me.

Yes + only if there are any annotated parameters in the `parse()` method.

Thanks for clarifying this gents!

scrapy_poet/downloadermiddlewares.py
```diff
@@ -107,6 +108,16 @@ def process_response(
         and an injectable attribute,
         the user-defined ``cb_kwargs`` takes precedence.
         """
+        if request.callback is None and spider.parse.__annotations__:
```
kmike, line 111: Would it be possible to only warn if annotated types are going to be handled by scrapy-poet? For example, one can annotate Response as scrapy.Response, or there might be another argument, passed through cb_kwargs.

BurnzZ: Good call! Updated in https://github.com/scrapinghub/scrapy-poet/pull/106/commits/b03296b72297a46825fc67381992162a74f2e12c.

scrapy_poet/downloadermiddlewares.py
```diff
@@ -91,6 +93,34 @@ def process_request(
...
+        self, request: Request, response: Response, spider: Spider
+    ) -> bool:
+        """See: https://github.com/scrapinghub/scrapy-poet/issues/48"""
+
+        # No need to skip if the callback doesn't default to the parse() method
+        if request.callback is not None:
+            return False
+
+        # If the Request.cb_kwargs possess all of the cb dependencies, then no
+        # warning message should be issued.
```
BurnzZ, line 106: I just realized while writing this part that the `Injector` doesn't take into account any of the `request.cb_kwargs`. This means it's building it all of the dependencies which could be replaced later on by any `request.cb_kwargs` that are present.

kmike: Do you see any issue with that?

BurnzZ: There could be some use cases where building a particular dependency might be too expensive and one approach to avoid such a cost is to provide it manually via `cb_kwargs`.

It also removes the user's escape hatch of avoiding the providers from being used on certain requests since they can provide it themselves within the spider's request.

We could address it later in another PR.

scrapy_poet/downloadermiddlewares.py
```diff
@@ -107,6 +137,16 @@ def process_response(
         and an injectable attribute,
         the user-defined ``cb_kwargs`` takes precedence.
         """
+        if self._skip_dependency_creation(request, response, spider):
```
Gallaecio, line 140: The implementation of `_skip_dependency_creation` does not seem to need the response object. I wonder if we can move this code to process_request, and get rid of the changes to is_callback_requiring_scrapy_response, i.e. emit the warning from a single place in code.

BurnzZ: > The implementation of _skip_dependency_creation does not seem to need the response object.

Good catch! It's curious as to why our flake8 hook hasn't captured this. Probably time to look at bumping out all the pre-commit hook versions (e.g. flake8 is already v6 but we're using v4).

BurnzZ: > I wonder if we can move this code to process_request, and get rid of the changes to is_callback_requiring_scrapy_response, i.e. emit the warning from a single place in code.

One concern of having it in `process_request()` is that some user's project with custom middlewares declared after `InjectionMiddleware` that replaces the `Request.callback` value could mess up the check.

It could rarely happen, but we still need to write about such peculiarity somewhere in the docs to prevent users from stumbling to it. I think having it in `process_response()` is worth the extra code diffs in this case.

Gallaecio: > One concern of having it in `process_request()` is that some user's project with custom middlewares declared after `InjectionMiddleware` that replaces the `Request.callback` value could mess up the check.

That’s a good point. And also explains why we need the warning in process request as well for the specific case of DummyResponse, since that does not generate a response at all.

scrapy_poet/injection.py
```diff
@@ -303,7 +304,14 @@ def get_callback(request, spider):
     return request.callback
 
 
-def is_callback_requiring_scrapy_response(callback: Callable):
+def get_response_annotation(callback: Callable):
```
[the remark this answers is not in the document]
BurnzZ: I think `get_response_signature_parameter` is more apt since it's returning `inspect.Parameter`.

We can keep this public, as it's quite handy for **scrapy-poet** users to also have out-of-the-box response introspection if they need it. I also don't see any reason to make it private as well.

CHANGELOG.rst
```diff
@@ -5,6 +5,17 @@ Changelog
...
+* Some changes in behavior when the callback isn't specified (i.e. ``None``) in
+  a ``Request`` which defaults to the ``parse()`` method. A ``UserWarning``
+  would be issued if the following are encountered:
+
+    * Fixed the issue where downloading the response is skipped when the
+      ``parse()`` method is annotated with ``scrapy_poet.DummyResponse``.
+      For instance: ``def parse(self, response: DummyResponse)``.
+    * The ``InjectionMiddleware`` won't attempt to build any dependencies anymore
+      when ``Request.callback == None`` and the ``parse()`` method has annotated
+      dependencies that involves any of ``SCRAPY_POET_PROVIDERS``.
```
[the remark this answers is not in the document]
BurnzZ: Updated the **CHANGELOG.rst** to mention `RobotsTxtMiddleware`/`ImagesPipeline`/`FilesPipeline`.

For the **documentation**, I created a new **Pitfalls** doc page to house all peculiarities of this package.

For the **warning**, I think we can omit mentioning these three (3) Scrapy components to have the message more concise. Instead, I think we can point the user to the **Pitfalls** docs for a more in-depth explanation.

> For this same reason, I wonder if in code examples we should also use something other than parse, as you did in a prior version of these changes, to minimize the chance of issues when users copy examples.

I think the current set of examples should be fine even if users copy them since we already fixed the incorrect behavior. The main thing we need to update is dependencies not being built as `parse()` needs to be declared as a callback manually in `start_requests()`. This is done.

If the user fails to do this by writing something like the following _(which has now since been removed in our docs/examples)_:

```python
class Spidey(Spider):
    name = "milesmorales"
    start_urls = ["https://dailybugle.com"]

    def parse(self, response: DummyResponse, page: NewPage):
        ...
```

They'd be met with a `TypeError` since `page` isn't injected + a `UserWarning` explaining what happened:

> A request has been encountered with callback=None which defaults to the parse() method. On such cases, annotated dependencies in the parse() method won't be built by scrapy-poet. However, if the request has callback=parse, the annotated dependencies will be built.

Gallaecio (approved):
Great work! I specially love the pitfalls page.

CHANGELOG.rst
```diff
@@ -5,6 +5,27 @@ Changelog
...
+      ``SCRAPY_POET_PROVIDERS``.
+
+  This affects the following built-in Scrapy components since they use
+  :class:`scrapy.http.Request` without setting any callback value:
+
+    * :class:`scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware`
+    * :class:`scrapy.pipelines.images.ImagesPipeline`
+    * :class:`scrapy.pipelines.files.FilesPipeline`
+
+  See the new :ref:`pitfalls` documentation for more information.
```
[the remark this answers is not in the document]
BurnzZ: Thanks! I've adapted the CHANGELOG with some of your suggestions. I like that we're now specifying explicitly how to avoid the warnings within the CHANGELOG.

It's a bit longer now but the user could immediately see if their code is affected and is able to update it quickly.

CHANGELOG.rst
```diff
@@ -5,6 +5,82 @@ Changelog
 TBR
 ---
 
+* Scrapy uses the ``parse()`` method when dealing with :class:`scrapy.http.Request`
+  instances having a callback of ``None``. On such cases, Scrapy uses the spider's
+  ``parse()`` method as the callback.
```
kmike, line 10:
```suggestion
* Scrapy uses spider's parse() method when :class:`scrapy.http.Request` callback is set to None.
```

scrapy_poet/downloadermiddlewares.py
```diff
@@ -91,6 +93,27 @@ def process_request(
...
+
+        # No need to skip if the callback doesn't default to the parse() method
+        if request.callback is not None:
+            return False
+
+        # If the Request.cb_kwargs possess all of the cb dependencies, then no
+        # warning message should be issued.
+        signature_iter = iter(inspect.signature(spider.parse).parameters)
+        next(signature_iter)  # skip the first arg: response
+        cb_param_names = set(signature_iter)
```
kmike, line 107: We might want to compute cb_param_names once, because inspection is costly sometimes (on pypy?). But let's not do it now.

scrapy_poet/injection.py
```diff
@@ -303,7 +304,7 @@ def get_callback(request, spider):
     return request.callback
 
 
-def is_callback_requiring_scrapy_response(callback: Callable):
+def is_callback_requiring_scrapy_response(callback: Callable, request: Request):
```
kmike, line 307: Can we get callback from the request, why is it required to pass both?

kmike (approved):
The PR looks good; feel free to merge it @BurnzZ when you're happy. This is such a tricky issue!

scrapy_poet/injection.py
BurnzZ: This is tricky since the original `callback` parameter currently accepts the fact that it could receive a value of **not** `None` since we're providing the spider's `parse()` method if `request.callback=None`.

I've refactored this again to instead accept an optional `raw_callback` value meant to accept the actual value of `request.callback`. I think this is more backward compatible overall. https://github.com/scrapinghub/scrapy-poet/pull/106/commits/e8e22fa700c6c4bcc9d4b878d9ef6cc8c8c7f2b0

The other alternative is to move this as a classmethod to `Injector` so that it has access to the `spider` instance which in turn grants access to the spider's `parse()` method if needed. I think this approach is more limiting than the one above.

PR #111, opened 2023-01-04
WIP: `Annotated[Item, PickFields("x", "y")]` to decide which fields to populate in callback

BurnzZ:
A continuation for https://github.com/scrapinghub/scrapy-poet/pull/88 but built on top of https://github.com/scrapinghub/scrapy-poet/pull/103 to reduce the chances of merge conflicts later on when writing the docs.

# TODO:

- [ ] avoid returning cached full-items as-is, vice-versa
- [ ] raise error when FieldPick has field not in PO
- [ ] handle the inconsistencies of `Annotated` in different Python versions
- [ ] more tests

When we're happy with the API:

- [ ] Changelog
- [ ] Docs

### DIFF
--- a/docs/api_reference.rst
+++ b/docs/api_reference.rst
@@ -11,6 +11,10 @@ API
    :members:
    :no-special-members:
 
+.. autoclass:: scrapy_poet.PickFields
+   :members:
+   :no-special-members:
+
 Injection Middleware
 ====================
 
--- a/scrapy_poet/__init__.py
+++ b/scrapy_poet/__init__.py
@@ -1,4 +1,4 @@
-from .api import DummyResponse, callback_for
+from .api import DummyResponse, PickFields, callback_for
 from .downloadermiddlewares import InjectionMiddleware
 from .page_input_providers import (
     CacheDataProviderMixin,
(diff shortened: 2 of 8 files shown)


BurnzZ:
There's still some stuff to do but I think this is ready for review to check if we're happy with the API:

```python
from typing Annotated, Optional

import attrs
import scrapy
from scrapy_poet import ItemPage, PickFields, handle_urls, field

@attrs.define
class BigItem:
    x: str
    y: Optional[str] = None

@handle_urls("example.com")
@attrs.define
class BigPage(ItemPage[BigItem]):
    @field
    def x(self) -> str:
        return "x"

    @field
    def y(self) -> str:
        return "y"

class SomeSpider(scrapy.Spider):
    name = "somespider"

    def start_requests(self):
        yield scrapy.Request("https://example.com", self.parse_item)

    def parse_item(self, response, item: Annotated[BigItem, PickFields("x")]):
        yield item  # should return BigItem(x="x", y=None)
```

Gallaecio:
The API looks good to me.

I think `NotPickFields` can be a nice API as well. In fact, I imagine it could potentially be used more often than `PickFields`. And I think we should definitely prevent mixing the two with an error or warning.

Issue #112, opened 2023-01-11
raising `web_poet.exceptions.Retry` doesn't create a new page object with new response


BurnzZ:
# Problem

Raising `Retry` doesn't refresh the page object (i.e. not creating a new instance) which contains the new response.

See examples below which demonstrates that the instance IDs are the same in the logs.

## Example 1: `raise Retry` inside spider

```python
import scrapy
from web_poet import WebPage, field
from web_poet.exceptions import Retry
from scrapy_poet import DummyResponse


class BookPage(WebPage):
    pass


class BooksSpider(scrapy.Spider):
    name = "books_spider"

    custom_settings = {
        "DOWNLOADER_MIDDLEWARES": {
            "scrapy_poet.InjectionMiddleware": 543,
        },
        "SPIDER_MIDDLEWARES": {
            "scrapy_poet.RetryMiddleware": 275,
        },
    }

    keep_instances_alive = []

    def start_requests(self):
        yield scrapy.Request('http://books.toscrape.com/', self.parse)

    def parse(self, response: DummyResponse, page: BookPage):
        self.keep_instances_alive.append(response)
        self.keep_instances_alive.append(page)
        self.keep_instances_alive.append(page.response)
        print(f"response ID: {id(response)}")
        print(f"page ID: {id(page)}")
        print(f"page.response ID: {id(page.response)}")

        raise Retry
```

```
2023-01-11 13:35:21 [scrapy.core.engine] INFO: Spider opened
2023-01-11 13:35:21 [scrapy.extensions.logstats] INFO: Crawled 0 pages (at 0 pages/min), scraped 0 items (at 0 items/min)
2023-01-11 13:35:21 [scrapy.extensions.telnet] INFO: Telnet console listening on 127.0.0.1:6023
2023-01-11 13:35:21 [scrapy.core.engine] DEBUG: Crawled (404) <GET http://books.toscrape.com/robots.txt> (referer: None)
2023-01-11 13:35:22 [scrapy.core.engine] DEBUG: Crawled (200) <GET http://books.toscrape.com/> (referer: None)
response ID: 140047906624848
page ID: 140047897561744
page.response ID: 140047897562032
2023-01-11 13:35:22 [scrapy.downloadermiddlewares.retry] DEBUG: Retrying <GET http://books.toscrape.com/> (failed 1 times): page_object_retry
[...]

Issue #110, opened 2023-01-04
Injector doesn't take into account user-provided dependencies from `request.cb_kwargs`


BurnzZ:
From https://github.com/scrapinghub/scrapy-poet/pull/106#discussion_r1060013740

Gallaecio:
Related: https://github.com/scrapinghub/scrapy-poet/issues/98

Issue #98, opened 2022-11-28
Define expected behavior when passing PO as part of cb_kwargs


VMRuiz:
I was working on a project where I manually had to retry some requests like: 

```python
def parse(self, response, parameter, po: PoetClass):
   if broken:
       return Request(response.url, cb_kwargs=response.request.cb_kwargs, self.parse)
    # [...]
```

In this case, the current `po` is passed to the next invokation of `parse` so its content is not recreated with the new value of `response`. I was talking with @Gallaecio about this and while I think this behavior is not wrong, it took me a while to figure out why I was not getting new data in `po` until I realised my poor decision. 

So, I think this should be allowed because there may be a case where someone want to pass and already created Poet Object instead of expecting a new one to be created. Nevertheless, as this is a quite rare case, maybe the library could log a warning to let the user know that he may be doing something wrong in most of the other cases.

Gallaecio:
@lampssy noticed that this issue no longer applies since https://github.com/scrapinghub/scrapy-poet/pull/113.

PR #113, opened 2023-01-11
fix `web_poet.exceptions.Retry` not working

BurnzZ:
Fixes:
- https://github.com/scrapinghub/scrapy-poet/issues/112

Also kind of related to:
- https://github.com/scrapinghub/scrapy-poet/issues/98
- https://github.com/scrapinghub/scrapy-poet/issues/110

### DIFF
--- a/scrapy_poet/downloadermiddlewares.py
+++ b/scrapy_poet/downloadermiddlewares.py
@@ -146,11 +146,6 @@ class InjectionMiddleware:
             request,
             response,
         )
-        # If it's the first request which is not retried, copy the original
-        # request.cb_kwargs which contains user-specificied dependencies. This
-        # should persist across all the retries.
-        if not request.meta.get("retry_times"):
-            request.meta["orig_cb_kwargs"] = request.cb_kwargs.copy()
         self._merge_dependencies(final_kwargs, request)
         return response
 
@@ -160,11 +155,5 @@ class InjectionMiddleware:
         dependencies provided by the user via ``scrapy.Request.cb_kwargs``.
         """
         for arg, value in final_kwargs.items():
-            # Ensures that user provided dependencies via ``Request.cb_kwargs``
-            # are used instead of the dependencies created by the providers.
-            if arg in request.meta.get("orig_cb_kwargs", {}):
-                continue
-            if not request.meta.get("retry_times") and arg in request.cb_kwargs:
-                continue
             request.cb_kwargs[arg] = value
             # TODO: check if all arguments are fulfilled somehow?
--- a/scrapy_poet/downloadermiddlewares.py
+++ b/scrapy_poet/downloadermiddlewares.py
@@ -5,7 +5,7 @@ are executed.
 import inspect
 import logging
 import warnings
-from typing import Dict, Generator, Optional, Type, TypeVar
+from typing import Generator, Optional, Type, TypeVar
 
 from scrapy import Spider, signals
 from scrapy.crawler import Crawler
@@ -146,14 +146,8 @@ class InjectionMiddleware:
             request,
             response,
         )
-        self._merge_dependencies(final_kwargs, request)
-        return response
-
-    @staticmethod
-    def _merge_dependencies(final_kwargs: Dict, request: Request) -> None:
-        """This merges the dependencies created by the providers alongside the
-        dependencies provided by the user via ``scrapy.Request.cb_kwargs``.
-        """
+        # Fill the callback arguments with the created instances
         for arg, value in final_kwargs.items():
             request.cb_kwargs[arg] = value
             # TODO: check if all arguments are fulfilled somehow?
+        return response
--- a/scrapy_poet/downloadermiddlewares.py
+++ b/scrapy_poet/downloadermiddlewares.py
@@ -150,7 +150,7 @@ class InjectionMiddleware:
         for arg, value in final_kwargs.items():
             # If scrapy-poet can't provided the dependency, allow the user to
             # give it.
-            if value is None and request.cb_kwargs.get(arg) is not None:
+            if value is None and arg in request.cb_kwargs:
                 continue
             request.cb_kwargs[arg] = value
             # TODO: check if all arguments are fulfilled somehow?
--- a/scrapy_poet/downloadermiddlewares.py
+++ b/scrapy_poet/downloadermiddlewares.py
@@ -150,7 +150,7 @@ class InjectionMiddleware:
         for arg, value in final_kwargs.items():
             # If scrapy-poet can't provided the dependency, allow the user to
             # give it.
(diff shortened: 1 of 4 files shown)


tests/test_middleware.py
```diff
@@ -263,6 +263,28 @@ def test_multi_args_callbacks(settings):
...
+    "final_kwargs,cb_kwargs,expected",
+    [
+        ({}, {}, {}),
+        ({"x": 1}, {}, {"x": 1}),
+        ({}, {"x": 2}, {"x": 2}),
+        ({"x": 1}, {"x": None}, {"x": 1}),
+        ({"x": 1}, {"x": False}, {"x": 1}),
+        ({"x": 1}, {"x": 2}, {"x": 1}),
+        ({"x": None}, {"x": 2}, {"x": 2}),
+        ({"x": False}, {"x": 2}, {"x": False}),
```
BurnzZ, line 276: This part is tricky since in order for `Retry` to work, the `InjectorMiddleware` needs to replace the `Request.cb_kwargs` values to ensure fresh page objects, responses, and other dependencies.

However, the user loses control when wanting to specifically provide a dependency instead of relying to scrapy-poet's providers.

I'll look into this further.

Gallaecio: Maybe we should make it so that by default re-injection happens, and later on provide a way to explicitly ask for a cb_kwarg not to be re-injected (maybe some meta key).

BurnzZ: I was thinking if we should:
- disallow users from overriding dependencies created by providers via `Request.cb_kwargs`
- instead, only allow dependencies from being overridden by something like `Request(..., meta={"SCRAPY_POET_DEPS": {page: MyPage()}})`

This means that:
- https://github.com/scrapinghub/scrapy-poet/issues/110 won't be needed since the `Injector` needs to use the meta key `SCRAPY_POET_DEPS`
- https://github.com/scrapinghub/scrapy-poet/issues/98 would cease to confuse users.

Though I'm not sure if this breaks some important use cases for some users.

BurnzZ: nvm, I found a way to persist the original user-specified `request.cb_kwargs` across multiple retries. https://github.com/scrapinghub/scrapy-poet/pull/113/commits/0425a70926487e40ffd215c0c76231f8b7380cbd 

I think this solution is much cleaner since:
- it doesn't add a new setting which could make **scrapy-poet** a bit more complex.
- user's have direct control on overriding the callback dependencies via `request.cb_kwargs` itself.

BurnzZ: As discussed off-GitHub, we're going with the first route discussed previously:

- disallow users from overriding dependencies created by providers via `Request.cb_kwargs`. This avoids the confusion during retries where the values in `cb_kwargs` come from the past `Request` or are actually provided by the user.

This avoids the complexities of doing https://github.com/scrapinghub/scrapy-poet/commit/0425a70926487e40ffd215c0c76231f8b7380cbd which creates complexities and confusion when new middlewares are added before/after the `InjectionMiddleware`.

This approach creates a backward incompatible change of not support user provided dependencies anymore via `Request.cb_kwargs`. https://github.com/scrapinghub/scrapy-poet/pull/113/commits/a1c48c1122ee4e02dfd66bc4b2aef44bb2c145cc

scrapy_poet/downloadermiddlewares.py
```diff
@@ -146,11 +146,14 @@ def process_response(
             request,
             response,
         )
-        # Fill the callback arguments with the created instances
+        self._merge_dependencies(final_kwargs, request)
+        return response
+
+    @staticmethod
+    def _merge_dependencies(final_kwargs: Dict, request: Request) -> None:
```
kmike, line 153: I think we can remove this method, as it doesn't really merge the user-provided dependencies anymore

tests/test_retries.py
```diff
@@ -21,15 +23,22 @@ class BaseSpider(Spider):
...
 
 
+def _assert_all_unique_instances(instances: List[Any]):
+    assert len({id(instance) for instance in instances}) == len(instances)
+
+
 @inlineCallbacks
 def test_retry_once():
     retries = deque([True, False])
+    po_instances, po_response_instances = [], []
```
kmike, line 33: page_instances, pages?

tests/test_middleware.py
```diff
@@ -259,7 +259,7 @@ def test_multi_args_callbacks(settings):
     item, _, _ = yield crawl_single_item(MultiArgsCallbackSpider, ProductHtml, settings)
     assert type(item["product"]) == ProductPage
     assert type(item["provided"]) == ProvidedWithDeferred
-    assert item["cb_arg"] == "arg!"
+    assert item["cb_arg"] is None
```
kmike, line 262: Why is this value dropped, and not preserved?

BurnzZ: It's because we're not allowing parameters from `Request.cb_kwargs` to replace things in the callbacks anymore.

Do you think we should have a special behavior for `Request.cb_kwargs` to only replace the callback values when they're `None`?

scrapy_poet/downloadermiddlewares.py
```diff
@@ -148,6 +148,11 @@ def process_response(
         )
         # Fill the callback arguments with the created instances
         for arg, value in final_kwargs.items():
+            # If scrapy-poet can't provided the dependency, allow the user to
+            # give it.
+            if value is None and request.cb_kwargs.get(arg):
```
kmike, line 153: It seems we can replace user-provided False or 0 with None, because of the way condition is written.

BurnzZ: Good catch! https://github.com/scrapinghub/scrapy-poet/pull/113/commits/800d512f97d0e9a71df0f0dd07d480b674beaa39

scrapy_poet/downloadermiddlewares.py
```diff
@@ -148,9 +148,11 @@ def process_response(
         )
         # Fill the callback arguments with the created instances
         for arg, value in final_kwargs.items():
-            # Precedence of user callback arguments
-            if arg not in request.cb_kwargs:
-                request.cb_kwargs[arg] = value
+            # If scrapy-poet can't provided the dependency, allow the user to
+            # give it.
+            if value is None and request.cb_kwargs.get(arg) is not None:
```
kmike, line 153: I think this is a correct implementation, but maybe `arg in request.cb_kwargs` is simpler?

BurnzZ: Updated in https://github.com/scrapinghub/scrapy-poet/pull/113/commits/af6905a01f6ba6fe008c35d37fcc5889fdc39443

kmike:
Thanks @BurnzZ!

PR #114, opened 2023-01-11
simplify changelog entry related to GH-106

kmike:
Hey! I tried to simplify the changelog entry a bit, to put the recommendations first.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,7 @@ TBR
   :class:`scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware`,
   :class:`scrapy.pipelines.images.ImagesPipeline` or
   :class:`scrapy.pipelines.files.FilesPipeline`. It is also a reason
-  ``HttpClient`` might not be working in page objects.
+  :class:`web_poet.page_inputs.client.HttpClient` might not be working in page objects.
   Now these cases are detected, and a warning is issued.
 
   As of Scrapy 2.7, it is not possible to fix the issue completely
@@ -24,10 +24,10 @@ TBR
       callback is assumed normally. But sometimes callback=None is used
       when :class:`scrapy.http.Request` is added to the Scrapy's downloader
       directly, in which case no callback is used. Middlewares, including
-      scrapy-poet's, can't distinguish between these two cases, which causes
+      **scrapy-poet**'s, can't distinguish between these two cases, which causes
       all kinds of issues.
 
-  We recommend all ``scrapy-poet`` users to modify their code to
+  We recommend all **scrapy-poet** users to modify their code to
   avoid the issue. Please **don't** define ``parse``
   method with arguments which are supposed to be filled by ``scrapy-poet``,
   and rename the existing ``parse`` methods if they have such arguments.
@@ -38,7 +38,7 @@ TBR
 
   There are backwards-incompatible changes related to this issue.
   They only affect you if you don't follow the advice of not using ``parse``
-  method with scrapy-poet.
+  method with **scrapy-poet**.
 
     * When the ``parse()`` method has its response argument annotated with
       :class:`scrapy_poet.api.DummyResponse`, for instance:
@@ -46,7 +46,7 @@ TBR
       instead of being skipped.
 
     * When the ``parse()`` method has dependencies that are provided by
-      ``scrapt-poet``, the :class:`scrapy_poet.downloadermiddlewares.InjectionMiddleware` won't
+      ``scrapy-poet``, the :class:`scrapy_poet.downloadermiddlewares.InjectionMiddleware` won't
       attempt to build any dependencies anymore.
 
       This causes the following code to have this error ``TypeError: parse()
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,12 +22,12 @@ TBR
   in page objects. Now these cases are detected, and a warning is issued.
 
   As of Scrapy 2.7, it is not possible to fix the issue completely
-  in scrapy-poet. Fixing it would require Scrapy changes; some 3rd party
+  in **scrapy-poet**. Fixing it would require Scrapy changes; some 3rd party
   libraries may also need to be updated.
 
   .. note::
 
-      The root of the issue is that when request.callback is None, ``parse``
+      The root of the issue is that when request.callback is ``None``, ``parse``
       callback is assumed normally. But sometimes callback=None is used
       when :class:`scrapy.http.Request` is added to the Scrapy's downloader
       directly, in which case no callback is used. Middlewares, including
@@ -35,7 +35,7 @@ TBR
       all kinds of issues.
 
   We recommend all **scrapy-poet** users to modify their code to
-  avoid the issue. Please **don't** define ``parse``
+  avoid the issue. Please **don't** define ``parse()``
   method with arguments which are supposed to be filled by **scrapy-poet**,
   and rename the existing ``parse`` methods if they have such arguments.
   Any other name is fine. It avoids all possible issues, including
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,24 +27,24 @@ TBR
 
   .. note::
 
-      The root of the issue is that when request.callback is ``None``, ``parse``
-      callback is assumed normally. But sometimes callback=None is used
-      when :class:`scrapy.http.Request` is added to the Scrapy's downloader
-      directly, in which case no callback is used. Middlewares, including
-      **scrapy-poet**'s, can't distinguish between these two cases, which causes
-      all kinds of issues.
+      The root of the issue is that when request.callback is ``None``,
+      ``parse()`` callback is assumed normally. But sometimes callback=None
+      is used when :class:`scrapy.http.Request` is added to the Scrapy's
+      downloader directly, in which case no callback is used. Middlewares,
+      including **scrapy-poet**'s, can't distinguish between these two cases,
+      which causes all kinds of issues.
 
   We recommend all **scrapy-poet** users to modify their code to
   avoid the issue. Please **don't** define ``parse()``
   method with arguments which are supposed to be filled by **scrapy-poet**,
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -61,30 +60,13 @@ TBR
...
-
-        * avoid defining the ``parse()`` method and instead use any other name.
-
-            * This is the **recommended** approach since it's easier and avoids
-              any :class:`scrapy.http.Request` with ``callback`` set to ``None``
-              which could come from anywhere (e.g. introducing new 3rd-party
-              middlewares or pipelines).
-
-        * ensure that all your :class:`scrapy.http.Request` has a ``callback``
-          **not** set to ``None``.
```
kmike, line 75: Won't users get warning from built-in components if they do it?

CHANGELOG.rst
```diff
@@ -5,49 +5,48 @@ Changelog
 TBR
 ---
 
-* Scrapy uses the spider's ``parse()`` method when the :class:`scrapy.http.Request`
-  callback is set to ``None``.
+* Improved ``scrapy-poet`` behavior when ``spider.parse()`` method
+  arguments are annotated. Previously, they were causing unnecessary work
```
BurnzZ, line 9: We should mention that if it's only annotated by `DummyResponse` andanything else that **scrapy-poet** provides. If it's annotated with `scrapy.http.Response`, there are no issues with it. [Reference](https://github.com/scrapinghub/scrapy-poet/blob/master/tests/test_downloader.py#L492-L515).

CHANGELOG.rst
```diff
@@ -5,49 +5,48 @@ Changelog
...
-        * :class:`scrapy.pipelines.images.ImagesPipeline`
-        * :class:`scrapy.pipelines.files.FilesPipeline`
-
-      To avoid this new behavior, the **recommended** approach is to avoid
-      defining a ``parse()`` method and instead choose any other name. 
-
     * When the ``parse()`` method has dependencies that are provided by
-      any active providers from ``SCRAPY_POET_PROVIDERS``, the
-      :class:`scrapy_poet.downloadermiddlewares.InjectionMiddleware` won't
+      ``scrapt-poet``, the :class:`scrapy_poet.downloadermiddlewares.InjectionMiddleware` won't
```
BurnzZ, line 49:
```suggestion
      ``scrapy-poet``, the :class:`scrapy_poet.downloadermiddlewares.InjectionMiddleware` won't
```

CHANGELOG.rst
```diff
@@ -5,49 +5,48 @@ Changelog
...
 
-* Scrapy uses the spider's ``parse()`` method when the :class:`scrapy.http.Request`
-  callback is set to ``None``.
+* Improved ``scrapy-poet`` behavior when ``spider.parse()`` method
+  arguments are annotated. Previously, they were causing unnecessary work
+  in unexpected places like
+  :class:`scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware`,
+  :class:`scrapy.pipelines.images.ImagesPipeline` or
+  :class:`scrapy.pipelines.files.FilesPipeline`. It is also a reason
+  ``HttpClient`` might not be working in page objects.
```
BurnzZ, line 14:
```suggestion
  :class:`web_poet.page_inputs.client.HttpClient` might not be working in page objects.
```

CHANGELOG.rst
```diff
@@ -5,49 +5,48 @@ Changelog
...
+  .. note::
+
+      The root of the issue is that when request.callback is None, ``parse``
+      callback is assumed normally. But sometimes callback=None is used
+      when :class:`scrapy.http.Request` is added to the Scrapy's downloader
+      directly, in which case no callback is used. Middlewares, including
+      scrapy-poet's, can't distinguish between these two cases, which causes
+      all kinds of issues.
+
+  We recommend all ``scrapy-poet`` users to modify their code to
```
BurnzZ, line 30:
```suggestion
  We recommend all **scrapy-poet** users to modify their code to
```

I think we should probably format it like this so it doesn't look like the inline code we have.

CHANGELOG.rst
```diff
@@ -5,49 +5,48 @@ Changelog
...
+  in scrapy-poet. Fixing it would require Scrapy changes; some 3rd party
+  libraries may also need to be updated.
+
+  .. note::
+
+      The root of the issue is that when request.callback is None, ``parse``
+      callback is assumed normally. But sometimes callback=None is used
+      when :class:`scrapy.http.Request` is added to the Scrapy's downloader
+      directly, in which case no callback is used. Middlewares, including
+      scrapy-poet's, can't distinguish between these two cases, which causes
```
BurnzZ, line 27:
```suggestion
      **scrapy-poet**'s, can't distinguish between these two cases, which causes
```

Keeping them consistent

CHANGELOG.rst
```diff
@@ -5,49 +5,48 @@ Changelog
...
-    
-    * :class:`scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware`
-    * :class:`scrapy.pipelines.images.ImagesPipeline`
-    * :class:`scrapy.pipelines.files.FilesPipeline`
+  See the new :ref:`pitfalls` documentation for more information.
 
-  There are some backward incompatible changes when this occurs:
+  There are backwards-incompatible changes related to this issue.
+  They only affect you if you don't follow the advice of not using ``parse``
+  method with scrapy-poet.
```
BurnzZ, line 41:
```suggestion
  method with **scrapy-poet**.
```

CHANGELOG.rst
```diff
@@ -5,49 +5,48 @@ Changelog
...
+      The root of the issue is that when request.callback is None, ``parse``
+      callback is assumed normally. But sometimes callback=None is used
+      when :class:`scrapy.http.Request` is added to the Scrapy's downloader
+      directly, in which case no callback is used. Middlewares, including
+      scrapy-poet's, can't distinguish between these two cases, which causes
+      all kinds of issues.
+
+  We recommend all ``scrapy-poet`` users to modify their code to
+  avoid the issue. Please **don't** define ``parse``
+  method with arguments which are supposed to be filled by ``scrapy-poet``,
```
BurnzZ, line 32: The main recommendation in the [Pitfalls](https://scrapy-poet.readthedocs.io/en/latest/intro/pitfalls.html) doc is to simply avoid defining the `parse()` method and choose any other name.

I think this is much simpler than a user checking if their `parse()` method has some dependencies fulfilled by **scrapy-poet**.

I also think it's less error prone in the future when users forget that this issue exists and they'd be adding dependencies to it. If they'd renamed the `parse()` method from the start, they'd be immune to future issues.

CHANGELOG.rst
BurnzZ: Yeah, they'd still get warnings from things like RobotsMiddlewares

CHANGELOG.rst
kmike: > I think this is much simpler than a user checking if their parse() method has some dependencies fulfilled by scrapy-poet.

Checking if a dependency is used looks simple enough: you find parse method, and see if it has arguments other than response. 

> I also think it's less error prone in the future when users forget that this issue exists and they'd be adding dependencies to it. If they'd renamed the parse() method from the start, they'd be immune to future issues.

I think the current workaround won't be there forever; we'll make a fix in Scrapy (hopefully, pretty soon) to avoid this issue, so that users will be able to use parse methods again. It also looks like if they add dependencies to parse method, a warning is going to be issued, and it won't work anyways, so hopefully the error would be obvious.

Is it ok with you to keep the recommendation of changing parse method only if it uses scrapy-poet provided arguments, at least in the changelog, to minimize the suggested changes users need to do to upgrade?

CHANGELOG.rst
```diff
@@ -10,51 +10,50 @@ TBR
...
+  to be provided by **scrapy-poet**. Previously, it was causing
+  unnecessary work in unexpected places like
+  :class:`scrapy.downloadermiddlewares.robotstxt.RobotsTxtMiddleware`,
+  :class:`scrapy.pipelines.images.ImagesPipeline` or
+  :class:`scrapy.pipelines.files.FilesPipeline`. It is also a reason
+  :class:`web_poet.page_inputs.client.HttpClient` might not be working
+  in page objects. Now these cases are detected, and a warning is issued.
+
+  As of Scrapy 2.7, it is not possible to fix the issue completely
+  in scrapy-poet. Fixing it would require Scrapy changes; some 3rd party
```
BurnzZ, line 25:
```suggestion
  in **scrapy-poet**. Fixing it would require Scrapy changes; some 3rd party
```

Last one for consistency. :)

CHANGELOG.rst
BurnzZ: > Is it ok with you to keep the recommendation of changing parse method only if it uses scrapy-poet provided arguments, at least in the changelog, to minimize the suggested changes users need to do to upgrade?

Yeah, let's go with it in the changelog. The docs provide more information anyways which could also help them determine the best course of action in their use case. :)

CHANGELOG.rst
```diff
@@ -10,51 +10,50 @@ TBR
...
+
+      The root of the issue is that when request.callback is None, ``parse``
+      callback is assumed normally. But sometimes callback=None is used
+      when :class:`scrapy.http.Request` is added to the Scrapy's downloader
+      directly, in which case no callback is used. Middlewares, including
+      **scrapy-poet**'s, can't distinguish between these two cases, which causes
+      all kinds of issues.
+
+  We recommend all **scrapy-poet** users to modify their code to
+  avoid the issue. Please **don't** define ``parse``
```
BurnzZ, line 38:
```suggestion
  avoid the issue. Please **don't** define ``parse()``
```

There are some unchanged lines that has this style. We could go with either ``parse()`` or ``parse`` but we need to make them consistent throughout the changelog.

kmike: Makes sense. Fixed in https://github.com/scrapinghub/scrapy-poet/pull/114/commits/123aa340e36690643d722a08edf34f90a62ee2b1

BurnzZ (approved):
Thanks @kmike !

PR #119, opened 2023-02-01
include `SCRAPY_POET_RULES` in install docs

BurnzZ:
I tried setting up a Scrapy project from scratch using our tutorials, i.e., copy-pasting the code from the installation section.

Took me a while to figure out why the Item Provider wasn't working. It was due to not declaring the `SCRAPY_POET_RULES` in the settings. :sweat\_smile: 

I think we can prevent this from happening by including it directly in the docs, especially now that we're encouraging items than overrides.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,6 +64,9 @@ Add the following inside Scrapy's ``settings.py`` file:
         "scrapy_poet.RetryMiddleware": 275,
     }
 
+    from web_poet import default_registry
+    SCRAPY_POET_RULES = default_registry.get_rules()
+
 Developing
 ==========
 
--- a/docs/intro/install.rst
+++ b/docs/intro/install.rst
@@ -33,6 +33,9 @@ of your Scrapy project:
         "scrapy_poet.RetryMiddleware": 275,
     }
 
+    from web_poet import default_registry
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -64,6 +64,9 @@ Add the following inside Scrapy's ``settings.py`` file:
         "scrapy_poet.RetryMiddleware": 275,
     }
 
+    from web_poet import default_registry
+    SCRAPY_POET_RULES = default_registry.get_rules()
```
kmike, line 68: That's actually a question I had in a previous PR, but haven't asked. Would it make sense to use these rules by default?

BurnzZ: > Would it make sense to use these rules by default?

I'm not quite sure at the moment since I'm not sure how users would use their rules. For example, they might only be picking some of them using `default_registry.search()` or directly creating the `ApplyRule` instances form here.

Gallaecio: It may still make sense to use the default registry rules by default. The other options are still possible by changing the value.

BurnzZ: I've created another PR in #120 to explore this alternative.

BurnzZ:
Closing in favor of the merged PR in https://github.com/scrapinghub/scrapy-poet/pull/120

PR #120, opened 2023-02-02
Default: `SCRAPY_POET_RULES = web_poet.default_registry.get_rules()`

BurnzZ:
Stemming from the discussion in https://github.com/scrapinghub/scrapy-poet/pull/119#discussion_r1093194207.

~Also addresses the failing tests as mentioned in https://github.com/scrapinghub/scrapy-poet/issues/123.~  Moved into another PR: https://github.com/scrapinghub/scrapy-poet/pull/125

### DIFF
--- a/docs/intro/basic-tutorial.rst
+++ b/docs/intro/basic-tutorial.rst
@@ -511,14 +511,15 @@ and store the :py:class:`web_poet.rules.ApplyRule` for you. All of the
     # rules from other packages. Otherwise, it can be omitted.
     # More info about this caveat on web-poet docs.
     consume_modules("external_package_A", "another_ext_package.lib")
-    SCRAPY_POET_RULES = default_registry.get_rules()
 
 .. note::
 
-    By default, ``SCRAPY_POET_RULES`` has values already set to
-    :meth:`web_poet.default_registry.get_rules`. However, if you need to add 
-    rules from other external packages, make sure to use
-    :func:`web_poet.rules.consume_modules` as shown above.
+    By default, ``SCRAPY_POET_RULES`` already has some rules set from the
+    return value of :meth:`web_poet.default_registry.get_rules()
+    <web_poet.rules.RulesRegistry.get_rules>`. However, if you need to add rules
+    from other external packages, make sure to use
+    :func:`web_poet.consume_modules <web_poet.rules.consume_modules>` as shown
+    above inside your ``settings.py``.
 
 For more info on this, you can refer to these docs:
 
--- a/docs/intro/basic-tutorial.rst
+++ b/docs/intro/basic-tutorial.rst
@@ -513,6 +513,13 @@ and store the :py:class:`web_poet.rules.ApplyRule` for you. All of the
     consume_modules("external_package_A", "another_ext_package.lib")
     SCRAPY_POET_RULES = default_registry.get_rules()
 
+.. note::
+
+    By default, ``SCRAPY_POET_RULES`` has values already set to
+    :meth:`web_poet.default_registry.get_rules`. However, if you need to add 
+    rules from other external packages, make sure to use
+    :func:`web_poet.rules.consume_modules` as shown above.
+
 For more info on this, you can refer to these docs:
 
     * ``scrapy-poet``'s :ref:`rules-from-web-poet` Tutorial section.
--- a/docs/intro/basic-tutorial.rst
+++ b/docs/intro/basic-tutorial.rst
@@ -498,12 +498,15 @@ is accepted and it would restrict the override only to category pages.
     Also see the `url-matcher <https://url-matcher.readthedocs.io/en/stable/>`_
     documentation for more information about the patterns syntax.
 
-Manually defining overrides like this would be inconvenient, most
-especially for larger projects. Fortunately, `web-poet`_ has a cool feature to
-annotate Page Objects like :py:func:`web_poet.handle_urls` that would define
-and store the :class:`web_poet.ApplyRule <web_poet.rules.ApplyRule>` for you.
-All of the :class:`web_poet.ApplyRule <web_poet.rules.ApplyRule>` rules could
-then be simply read as:
+Manually defining overrides like this would be inconvenient, most especially for
+larger projects. Fortunately, ``scrapy-poet`` already retrieves the rules defined
+from `web-poet`_'s ``default_registry``. This is done by setting the default
+value of the ``SCRAPY_POET_RULES`` setting as
+:meth:`web_poet.default_registry.get_rules() <web_poet.rules.RulesRegistry.get_rules>`.
+
+However, if you need to add rules from other external packages, make sure to use
+:func:`web_poet.consume_modules <web_poet.rules.consume_modules>` inside your
+``settings.py`` file as shown below.
 
 .. code:: python
 
@@ -513,15 +516,6 @@ then be simply read as:
     # rules from other packages. Otherwise, it can be omitted.
     # More info about this caveat on web-poet docs.
     consume_modules("external_package_A", "another_ext_package.lib")
-
-.. note::
-
-    By default, ``SCRAPY_POET_RULES`` already has some rules set from the
-    return value of :meth:`web_poet.default_registry.get_rules()
-    <web_poet.rules.RulesRegistry.get_rules>`. However, if you need to add rules
-    from other external packages, make sure to use
-    :func:`web_poet.consume_modules <web_poet.rules.consume_modules>` as shown
-    above inside your ``settings.py``.
 
 For more info on this, you can refer to these docs:
 
--- a/docs/intro/basic-tutorial.rst
+++ b/docs/intro/basic-tutorial.rst
@@ -506,7 +506,7 @@ value of the ``SCRAPY_POET_RULES`` setting as
 
 However, this only works if page objects are annotated using the
 :func:`web_poet.handle_urls` decorator. You also need to set the
-``SCRAPY_POET_MODULES`` setting so that these rules could be properly imported.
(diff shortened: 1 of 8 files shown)


tests/test_downloader.py
```diff
@@ -434,11 +434,11 @@ class BaseSpider(Spider):
```
BurnzZ, line 434: Some tests need to be updated since it's now getting other warnings since we've used `default_registry.get_rules()`.

docs/settings.rst
```diff
@@ -26,9 +26,30 @@ Deprecated. Use ``SCRAPY_POET_RULES`` instead.
...
+
+        from web_poet import default_registry, consume_modules
+
+        # The consume_modules() must be called first if you need to properly import
+        # rules from other packages. Otherwise, it can be omitted.
+        # More info about this caveat on web-poet docs.
+        consume_modules("external_package_A", "another_ext_package.lib")
+
+        # To get all of the Override Rules that were declared via annotations.
+        SCRAPY_POET_RULES = default_registry.get_rules()
```
Gallaecio, line 49: Isn’t the point that this line is not necessary given the default value?

BurnzZ: It's still necessary if rules from external packages are needed due to the `consume_modules()` needing to be called.

Otherwise, the default rules inside `SCRAPY_POET_RULES` is incomplete.

Gallaecio: But at which point is `get_rules` as a default value called? If it is called after settings have been defined, then defining `consume_modules` in `settings.py` should still work, right? If the current implementation calls get_rules too early, maybe we could change that. But I would check first, because I am not sure the current code needs to be changed.

BurnzZ: Aha, I see your point. The `default_registry.get_rules()` is only called after the settings has been loaded, so yes, we can avoid redefining `SCRAPY_POET_RULES` here.

Updated in c42618e.

docs/intro/basic-tutorial.rst
```diff
@@ -511,7 +511,15 @@ and store the :py:class:`web_poet.rules.ApplyRule` for you. All of the
...
-    SCRAPY_POET_RULES = default_registry.get_rules()
+
+.. note::
+
+    By default, ``SCRAPY_POET_RULES`` already has some rules set from the
+    return value of :meth:`web_poet.default_registry.get_rules()
+    <web_poet.rules.RulesRegistry.get_rules>`. However, if you need to add rules
+    from other external packages, make sure to use
+    :func:`web_poet.consume_modules <web_poet.rules.consume_modules>` as shown
+    above inside your ``settings.py``.
```
Gallaecio, line 522:
```suggestion
    above inside ``settings.py``.
```

or 

```suggestion
    above inside your ``settings.py`` file.
```

Gallaecio (approved):

docs/intro/basic-tutorial.rst
```diff
@@ -511,7 +511,15 @@ and store the :py:class:`web_poet.rules.ApplyRule` for you. All of the
     # rules from other packages. Otherwise, it can be omitted.
     # More info about this caveat on web-poet docs.
     consume_modules("external_package_A", "another_ext_package.lib")
-    SCRAPY_POET_RULES = default_registry.get_rules()
```
kmike, line 514: The text above tells

> All of the
:py:class:`web_poet.rules.ApplyRule` rules could then be simply read as:

But the example no longer reads the rules.

docs/settings.rst
```diff
@@ -26,9 +26,17 @@ Deprecated. Use ``SCRAPY_POET_RULES`` instead.
...
-Default: ``None``
+Default: :meth:`web_poet.default_registry.get_rules()
+<web_poet.rules.RulesRegistry.get_rules>`
+
+.. warning::
+
+    Although ``SCRAPY_POET_RULES`` already has values set from the return value of
+    :meth:`web_poet.default_registry.get_rules() <web_poet.rules.RulesRegistry.get_rules>`,
+    make sure you call :func:`web_poet.consume_modules <web_poet.rules.consume_modules>`
+    inside your ``settings.py`` if you're using other rules from other external packages.
```
kmike, line 37: What does external packages mean in this context? Does it mean "not in the same Scrapy project"?

docs/intro/basic-tutorial.rst
BurnzZ: Good catch, I've cleaned up the paragraphs above.

docs/settings.rst
BurnzZ: Added some text to clarify that the external packages contain page objects annotated with the `@handle_urls`

docs/rules-from-web-poet.rst
```diff
@@ -170,12 +170,18 @@ The :py:func:`web_poet.handle_urls` decorator in this case is indicating that
...
-For example:
+``scrapy-poet`` automatically uses the rules defined by the ``@handle_urls``
+annotation by having the default value of the ``SCRAPY_POET_RULES`` setting set to
+:meth:`web_poet.default_registry.get_rules() <web_poet.rules.RulesRegistry.get_rules>`.
+The :py:meth:`get_rules() <web_poet.rules.RulesRegistry.get_rules>` method of the
+``default_registry`` returns a ``List[ApplyRule]``.
+
+However, if you need to add rules from other external packages, make sure to use
+:func:`web_poet.consume_modules <web_poet.rules.consume_modules>` inside your
+``settings.py`` file as shown below.
```
kmike, line 184: I'm still unsure about this "external packages" thing - what are these rules external to. 

Let's say you have a repo with a Scrapy project and page objects

```
my-project-repo/
├── project-pages/
│   ├── project_pages/
│   │   ├── __init__.py
│   │   └── website1.py
│   └── setup.py
├── my-scrapy-project/
│   ├── my_scrapy_project/
│   │   ├── spiders/
│   │   │   └── website1.py
│   │   ├── inline_pages/
│   │   │   └── website2/
│   │   │       ├── product.py
│   │   │       └── product_list.py
│   │   ├── __init__.py
│   │   ├── items.py
│   │   └── middleware.py
│   └── settings.py
└── setup.py
```

and also a separate installable library with page objects:

```
pages-lib/
├── pages_lib/
│   ├── website1/
│   │   ├── product.py
│   │   └── product_list.py
│   └── website2/
│       ├── product.py
│       └── product_list.py
├── setup.py
└── README.rst      
```

All the page objects use `handle_urls` decorator.

Which of them should be considered external? It seems one should use consume_modules for all packages which are not a part of Scrapy project ("external" means "external to the Scrapy project"?)

For modules inside a Scrapy project it is not needed only because default SpiderLoader calls walk_tree for the entries of SPIDER_MODULES setting, which is (by the way) empty by default, and is only set in the default settings.py template. So, I guess there could be cases when even the page objects defined in the project are not loaded - e.g. a case when a spider is executed in a script via CrawlerProcess, and there is no need to define SPIDER_MODULES.

It seems the safest would be to ask users to call consume_modules for all packages which may contain page objects, and either clarify "external" vs "internal" terminology, or stop using it. I'd also not go into details explaining SPIDER_MODULES in our docs.

It may also be not a bad idea to add PAGE_OBJECT_PACKAGES (or PAGE_OBJECT_MODULES) option. If it's easier to add this option than to fix the documentaion, it makes sense to do it in this PR; otherwise, it's better to do it separately.

BurnzZ: I see what you mean. Thanks for laying it down.

> It seems the safest would be to ask users to call consume\_modules for all packages which may contain page objects

Yeah, it would seem that would ensure it would work for all cases.

> It may also be not a bad idea to add PAGE\_OBJECT\_PACKAGES (or PAGE\_OBJECT\_MODULES - not precise, but more similar to SPIDER\_MODULES) option

That's actually much easier to explain. It's worth going this route than directly exposing `consume_modules`.:+1: 

> It also might be a good idea to separate the CI fixes to a different PR, though not required.

Good idea. I've moved it here: #125

BurnzZ: What do you think about the new `SCRAPY_POET_MODULES` implemented in af2cace ?

docs/settings.rst
```diff
@@ -35,13 +35,19 @@ Accepts a ``List[ApplyRule]`` which sets the rules to use.
...
 There are sections dedicated for this at :ref:`intro-tutorial` and
 :ref:`rules-from-web-poet`.
 
+SCRAPY_POET_MODULES
+-------------------
+
+Default: ``[]``
+
+A list of modules (i.e. ``List[str]``) which scrapy-poet will look for page objects
+annotated with the :func:`web_poet.handle_urls` decorator. Each module is passed
```
kmike, line 49: I think we should mention 2 things here:

1. It may not be just modules, but packages as well. All modules in a package are loaded recursively.
2. It is possible to get page ojects in the registry even if the module is not in SCRAPY_POET_MODULES. We may say that for handle_urls to work a module needs to be imported, and SCRAPY_POET_MODULES just makes sure that all modules/packages listed there are imported.

BurnzZ: Added more details in f16dbf2

tests/test_downloader.py
```diff
@@ -508,10 +517,10 @@ def parse(self, response: scrapy.http.Response):
 
         crawler = make_crawler(TestSpider, {})
 
-        with warnings.catch_warnings(record=True) as warning_msg:
+        with pytest.warns(UserWarning) as record:
```
kmike, line 520: Are these changes related to the pull request? 
The test docstring contradicts the test behavior.

BurnzZ: Yes, wherein it should not encounter particular warning messages by setting `not_existing=True`.

kmike: aha, I see, thanks!

kmike:
Thanks @BurnzZ!

PR #122, opened 2023-02-02
Don't put microseconds into frozen_time.

wRAR:


Gallaecio (approved):
I hope 1 second of leeway is enough :crossed_fingers:

kmike:
Thanks @wRAR!

PR #125, opened 2023-02-07
fix failing tests after new warning phrases for dupe rules

### LINKED ISSUE #123: Update tests for the new web-poet
web-poet changed some messages and so tests in test_web_poet_rules.py that expect a "Consider updating the priority of these rules:" message fail and tests that expect it to be absent are not checking the correct thing too.

BurnzZ:
Resolves #123


kmike:
Thanks @BurnzZ!

PR #126, opened 2023-02-08
officially support scrapy.http.request.NO_CALLBACK on scrapy>=2.8

### LINKED ISSUE #118: Update behavior when encountering `Request.callback = NO_CALLBACK`
This is an issue tracker to follow-up on:
- https://github.com/scrapinghub/scrapy-poet/pull/106

after Scrapy 2.8 has been released bearing this feature:
- https://github.com/scrapy/scrapy/pull/5798

BurnzZ:
Resolves #118

~Apparently, there's no need to update the `InjectionMiddleware` ([ref](https://github.com/scrapinghub/scrapy-poet/pull/106/files#diff-04d90547880d12f9db87df2cdd5d6cf986ee4471acfe380b29a12645063503dbR100)) and `is_callback_requiring_scrapy_response` ([ref](https://github.com/scrapinghub/scrapy-poet/pull/106/files#diff-48a2d08707cccfa87db10d30b57025420baa731b27434a78d8f529e6649796e8R336)) since the [pitfall behavior](https://scrapy-poet.readthedocs.io/en/stable/intro/pitfalls.html) only kicks in when the request callback is set to `None`.~

~This means that using `NO_CALLBACK` without the changes from this PR, should work. This PR simply makes the support for `NO_CALLBACK` official.~

Only added Scrapy 2.7 and 2.8 for this PR which borders when the `NO_CALLBACK` has been introduced.


docs/intro/pitfalls.rst
```diff
@@ -15,9 +21,9 @@ Pitfalls
     The TLDR; recommendation is to simply avoid defining a ``parse()`` method
     and instead choose another name.
 
-Scrapy supports declaring :class:`scrapy.Request <scrapy.http.Request>` instances
-without setting any callbacks (i.e. ``None``). For these instances, Scrapy uses
-the ``parse()`` method as its callback.
+Scrapy < 2.8 supports declaring :class:`scrapy.Request <scrapy.http.Request>`
```
kmike, line 24: Scrapy 2.8+ also supports it.

docs/intro/pitfalls.rst
```diff
@@ -7,6 +7,12 @@ Pitfalls
 ``scrapy.Request`` without callback
 ===================================
 
+.. tip::
+
+    Note that the pitfalls discussed in this section isn't applicable to
```
kmike, line 12:
```suggestion
    Note that the pitfalls discussed in this section aren't applicable to
```

tests/test_downloader.py
```diff
@@ -651,3 +651,28 @@ def parse(
...
+    except ImportError:
+        pytest.skip("NO_CALLBACK not available in Scrapy < 2.8")
+
+    with MockServer(EchoResource) as server:
+
+        class TestSpider(BaseSpider):
+            start_urls = [server.root_url]
+
+            def start_requests(self):
+                yield Request(server.root_url, callback=NO_CALLBACK)
```
kmike, line 671: Is it possible to support regular spiders now, e.g. with start_urls without start_requests?

tox.ini
```diff
@@ -24,11 +24,26 @@ deps =
     andi==0.4.1
     attrs==21.3.0
     parsel==1.5.0
-    scrapy==2.6.0
```
kmike, line 27: Why is min scrapy version removed here? I think we should be running at least one test with the minimum Scrapy version which we support.

BurnzZ: addressed in d17adcb

docs/intro/pitfalls.rst
BurnzZ: e0a4fc9

docs/intro/pitfalls.rst
BurnzZ: Good catch! e0a4fc9

tests/test_downloader.py
BurnzZ: Addressed in 19b7cf4

.github/workflows/test.yml
```diff
@@ -43,10 +43,8 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install tox
     - name: tox
-      env:
-        TOXENV: ${{ matrix.toxenv }}
       run: |
-        tox -e py
+        tox -e ${{ matrix.toxenv || 'py' }}
```
BurnzZ, line 47: These changes were necessary since before, GitHub actions was not using some of the pinned configurations of `min`, `asyncio`, and `asyncio-min`.

For example, `min` was not using `scrapy==2.6.0` but the latest version at the current time: https://github.com/scrapinghub/scrapy-poet/actions/runs/4047094013/jobs/6960692434

tests/test_downloader.py
BurnzZ: Addressed in 6c5895e, but some of the asyncio toxenvs are still failing.

.github/workflows/test.yml
BurnzZ: Moved this into a separate PR: #129

tests/test_downloader.py
BurnzZ: I've created a separate PR in #129 to address the unrelated CI failures.

.github/workflows/test.yml
Gallaecio: The fix here was to change `tox -e py` to `tox`, then `TOXENV` should work as expected.

wRAR (approved):

Gallaecio (approved):

CHANGELOG.rst
```diff
@@ -80,6 +80,24 @@ TBR
 * Moved some of the utility functions from the test module into
   ``scrapy_poet.utils.testing``.
 
+* Supports the new :func:`scrapy.http.request.NO_CALLBACK` introduced in
+  **Scrapy 2.8**. This means that the :ref:`pitfalls` (introduced in
+  ``scrapy-poet==0.7.0``) when using the ``parse()`` method won't apply if
+  :class:`scrapy.Request <scrapy.http.Request>`'s callback is set to
+  :func:`scrapy.http.request.NO_CALLBACK`.
```
kmike, line 87: I wonder if it'd be clear for the users what it means. Previously, we were saying "rename your parse methods". Now we say that "in Scrapy 2.8 there are no pitfalls if callback is set to NO_CALLBACK". If I'm reading it, I might think that I need to set the callback to NO_CALLBACK somewhere to avoid pitfalls, but I have no idea what is NO_CALLBACK, and when to use it.

Is it possible to say that if you use Scrapy 2.8, then pitfalls don't apply, unless you use third-party middlewares which send Requests directly to downloader and which are not updated to use NO_CALLBACK? 

I think this applies to the pitfalls doc as well. 

In Scrapy 2.8+ the pitfall is this: everything works as usual if you write spiders. But if there is some tricky code which adds Requests to downloader directly without a callback, it must use NO_CALLBACK, not None.

In Scrapy < 2.8 - pitfalls are explained in the existing pitfalls doc.

BurnzZ: Addressed in 7058bba

.github/workflows/test.yml
BurnzZ: I've gotten inspiration from [https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/.github/workflows/test.yml#L20-L48.](https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/.github/workflows/test.yml#L20-L48.)

Gallaecio: Hmm… Well, the -e approach now does allow to avoid quite some duplicate lines, so you are right, it is probably the best approach.

kmike:
Thanks @BurnzZ!

PR #127, opened 2023-02-08
Support HttpClient in savefixture

wRAR:
Based on https://github.com/scrapinghub/web-poet/pull/128

### DIFF
--- a/scrapy_poet/commands.py
+++ b/scrapy_poet/commands.py
@@ -84,6 +84,7 @@ class SaveFixtureCommand(ScrapyCommand):
         spider_cls = spider_for(cls)
         self.settings["ITEM_PIPELINES"][SavingPipeline] = 100
         self.settings["DOWNLOADER_MIDDLEWARES"][SavingInjectionMiddleware] = 543
+        self.settings["_SCRAPY_POET_SAVEFIXTURE"] = True
 
         frozen_time = datetime.datetime.now(datetime.timezone.utc).replace(
             microsecond=0
--- a/scrapy_poet/commands.py
+++ b/scrapy_poet/commands.py
@@ -83,6 +83,10 @@ class SaveFixtureCommand(ScrapyCommand):
 
         spider_cls = spider_for(cls)
         self.settings["ITEM_PIPELINES"][SavingPipeline] = 100
+        self.settings["DOWNLOADER_MIDDLEWARES"][
+            "scrapy_poet.downloadermiddlewares.InjectionMiddleware"
+        ] = None
+        self.settings["DOWNLOADER_MIDDLEWARES"][InjectionMiddleware] = None
         self.settings["DOWNLOADER_MIDDLEWARES"][SavingInjectionMiddleware] = 543
         self.settings["_SCRAPY_POET_SAVEFIXTURE"] = True
 
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -49,10 +49,7 @@ class BTSBookPage(WebPage):
     assert fixture.is_valid()
     assert (fixture.input_path / "HttpResponse-body.html").exists()
     assert fixture.meta_path.exists()
-    assert (
-        fixture.input_path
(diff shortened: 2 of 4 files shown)


tests/test_commands.py
```diff
@@ -47,6 +49,10 @@ def to_item(self):
     assert fixture.is_valid()
     assert (fixture.input_path / "HttpResponse-body.html").exists()
     assert fixture.meta_path.exists()
+    assert (
+        fixture.input_path
+        / "HttpClient-4a4c68d01cec0684e4eb2d155534005ac2db9a33-body.html"
```
kmike, line 54: What do you think about saving the requests as well, and not using fingerprints on the file system?

Pros:

* tests are easier to inspect
* easier to spot issues
* if the fingerprinting algorithm has an issue, fixing it doesn't require regenerating all the test cases, because fingerprints are only used internally

Cons:

* more data to save
* probably, more code

I.e. you'd have something like this on disk:

```
HttpClient-1-HttpRequest.json
HttpClient-1-HttpResponse-body.html
HttpClient-1-HttpResponse-other.json
```

When running tests, you load all the requests and responses; they can be matched using the id ("1" in this case). To compare a request issued with a stored request we can use fingerprints or something else; we're free to change the algorithm at any point, because we have complete information about requests, and we just need to compare them.

BurnzZ: An alternative that I could think of is bundling all information between request-response pairs in one file. For example:

```
HttpClient-1.json
HttpClient-2.json
HttpClient-3.json
...
```

Pros:

*   No need to link different files together

Cons:

*   More data inside a single file, though could get ugly on diffs

Although I think there's merit in having the fingerprint in the filename, for example, `HttpClient-4a4c68d01cec0684e4eb2d155534005ac2db9a33.json` since it offers an quick visual inspection that diffs with different request-response pairs should have different fingerprints. As opposed to a user deleting `HttpClient-72.json` and then adding a new fixture for a totally different request-response pair but with the same `HttpClient-72.json` filename.

kmike: The issue with a single file is that response body becomes unreadable (and unwritable by humans) in the most common case - JSON AJAX requests. 

When saving responses in general, we tried to make sure that if the response is html, you can just open it directly, for quick inspection, it's stored as a stand-alone html file. The same applies here, to the AJAX payloads. Though the file name shouldn't be .html probably - what do you think @wRAR @BurnzZ ?

We actually have the same issue with requests, so it could make sense to use 2 files as well if there is request body.

wRAR: > Though the file name shouldn't be .html probably

It's .txt in my current unpublished code, unless it's empty in which case there is no file.

And yeah, it seems much better to save the request body separately than inside a JSON.

BurnzZ: > When saving responses in general, we tried to make sure that if the response is html, you can just open it directly, for quick inspection

Makes sense. :+1:

Gallaecio (approved):

scrapy_poet/commands.py
```diff
@@ -83,7 +83,12 @@ def run(self, args, opts):
 
         spider_cls = spider_for(cls)
         self.settings["ITEM_PIPELINES"][SavingPipeline] = 100
+        self.settings["DOWNLOADER_MIDDLEWARES"][
+            "scrapy_poet.downloadermiddlewares.InjectionMiddleware"
+        ] = None
+        self.settings["DOWNLOADER_MIDDLEWARES"][InjectionMiddleware] = None
         self.settings["DOWNLOADER_MIDDLEWARES"][SavingInjectionMiddleware] = 543
+        self.settings["_SCRAPY_POET_SAVEFIXTURE"] = True
```
kmike, line 91: I wonder if it'd be simpler if we have a single InjectionMiddleware, which reads the value of _SCRAPY_POET_SAVEFIXTURE setting and acts accordingly.

kmike:
Thanks @wRAR!

PR #129, opened 2023-02-09
Fix CI failing on `min`, `asyncio`, and `asyncio-min` envs

BurnzZ:
Built on top of #126 since it needs the `is_min_scrapy_version()` utility function.

These changes are necessary since otherwise, GitHub actions was not using some of the pinned configurations of `min`, `asyncio`, and `asyncio-min`.

For example, `min` was not using `scrapy==2.6.0` but the latest version at the current time: [https://github.com/scrapinghub/scrapy-poet/actions/runs/4047094013/jobs/6960692434](https://github.com/scrapinghub/scrapy-poet/actions/runs/4047094013/jobs/6960692434)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,9 +4,8 @@
 name: tox
 
 on:
-  push:
-    branches: [ master ]
   pull_request:
+  push:
     branches: [ master ]
 
 jobs:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -88,6 +88,8 @@ TBR
 
 * Documentation improvements.
 
+* Various test and CI Fixes.
+
 * Deprecations:
 
     * The ``SCRAPY_POET_OVERRIDES`` setting has been replaced by
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -93,10 +93,10 @@ TBR
 * Fix the ``twisted.internet.error.ReactorAlreadyInstalledError`` error raised
   when using Twisted < 21.2.0.
 
-* Fix CI configuration that doesn't follow it's intended commands in these
-  tox environments: ``min``, ``asyncio-min``, and ``asyncio``. This ensures
-  that page objects using ``asyncio`` should work properly, alongside the minimum
-  specified Twisted version.
+* Fix test configuration that doesn't follow the intended commands and dependencies
+  in these tox environments: ``min``, ``asyncio-min``, and ``asyncio``. This
+  ensures that page objects using ``asyncio`` should work properly, alongside
+  the minimum specified Twisted version.
 
 * Documentation improvements.
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -86,6 +86,10 @@ TBR
   :class:`scrapy.Request <scrapy.http.Request>`'s callback is set to
   :func:`scrapy.http.request.NO_CALLBACK`.
 
+* Fix ``TypeError`` when using Twisted <= 21.7.0 since scrapy-poet was using
+  ``twisted.internet.defer.Deferred[object]`` type annotation before which was
+  not subscriptable.
+
 * Documentation improvements.
 
 * Various test and CI Fixes.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -90,9 +90,15 @@ TBR
   ``twisted.internet.defer.Deferred[object]`` type annotation before which was
   not subscriptable.
 
-* Documentation improvements.
+* Fix the ``twisted.internet.error.ReactorAlreadyInstalledError`` error raised
+  when using Twisted < 21.2.0.
 
-* Various test and CI Fixes.
+* Fix CI configuration that doesn't follow it's intended commands in these
+  tox environments: ``min``, ``asyncio-min``, and ``asyncio``. This ensures
+  that page objects using ``asyncio`` should work properly, alongside the minimum
+  specified Twisted version.
+
+* Documentation improvements.
 
 * Deprecations:
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -91,7 +91,7 @@ TBR
   not subscriptable.
 
 * Fix the ``twisted.internet.error.ReactorAlreadyInstalledError`` error raised
-  when using Twisted < 21.2.0.
+  when using the ``scrapy savefixture`` command and Twisted < 21.2.0 is installed.
 
 * Fix test configuration that doesn't follow the intended commands and dependencies
   in these tox environments: ``min``, ``asyncio-min``, and ``asyncio``. This
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -17,6 +17,7 @@ from typing import (
     Callable,
     ClassVar,
     Dict,
+    Generator,
     List,
     Optional,
     Sequence,
@@ -32,6 +33,7 @@ import attr
 from scrapy import Request
 from scrapy.crawler import Crawler
 from scrapy.http import Response
+from twisted.internet.defer import Deferred, ensureDeferred, inlineCallbacks
 from web_poet import (
     HttpClient,
     HttpResponse,
@@ -329,12 +331,13 @@ class ItemProvider(PageObjectInputProvider):
     def get_from_cache(self, request: Request, cls: Callable) -> Optional[Any]:
         return self._cached_instances.get(request, {}).get(cls)
 
-    async def __call__(
+    @inlineCallbacks
+    def __call__(
         self,
         to_provide: Set[Callable],
         request: Request,
         response: Response,
-    ) -> List[Any]:
+    ) -> Generator[Deferred[object], object, List[Any]]:
         results = []
         for cls in to_provide:
             item = self.get_from_cache(request, cls)
@@ -361,7 +364,7 @@ class ItemProvider(PageObjectInputProvider):
             )
 
             try:
-                po_instances = await self.injector.build_instances(
+                po_instances = yield from self.injector.build_instances(
                     request, response, plan
                 )
             except RecursionError:
@@ -371,7 +374,7 @@ class ItemProvider(PageObjectInputProvider):
                 )
 
             page_object = po_instances[page_object_cls]
-            item = await page_object.to_item()
+            item = yield from ensureDeferred(page_object.to_item())
 
             self.update_cache(request, po_instances)
             self.update_cache(request, {type(item): item})
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -9,6 +9,7 @@ different providers in order to acquire data from multiple external sources,
 for example, from scrapy-playwright or from an API for automatic extraction.
 """
 import abc
+import asyncio
 import json
 from dataclasses import make_dataclass
 from inspect import isclass
@@ -17,7 +18,6 @@ from typing import (
     Callable,
     ClassVar,
     Dict,
-    Generator,
     List,
     Optional,
     Sequence,
@@ -33,7 +33,7 @@ import attr
 from scrapy import Request
 from scrapy.crawler import Crawler
 from scrapy.http import Response
-from twisted.internet.defer import Deferred, ensureDeferred, inlineCallbacks
+from scrapy.utils.defer import maybe_deferred_to_future
 from web_poet import (
     HttpClient,
     HttpResponse,
@@ -308,6 +308,11 @@ class ItemProvider(PageObjectInputProvider):
 
     name = "item"
 
+    template_deadlock_msg = (
+        "Deadlock detected! A loop has been detected to "
+        "trying to resolve this plan: {plan}"
+    )
+
     def __init__(self, injector):
         super().__init__(injector)
         self.registry = self.injector.registry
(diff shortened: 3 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -4,9 +4,8 @@
 name: tox
 
 on:
-  push:
-    branches: [ master ]
   pull_request:
+  push:
```
BurnzZ, line 8: This enables the CI runs on pull requests not inteded to be merged into the `master` branch.

BurnzZ, line 8: This enables the CI runs on pull requests not inteded to be merged into the `master` branch.

scrapy_poet/page_input_providers.py
```diff
@@ -32,6 +32,7 @@
 from scrapy import Request
 from scrapy.crawler import Crawler
 from scrapy.http import Response
+from twisted.internet.defer import ensureDeferred, inlineCallbacks
```
BurnzZ, line 35: Needed to be compatible with the `asyncio` reactor.

scrapy_poet/page_input_providers.py
```diff
@@ -329,12 +331,13 @@ def update_cache(self, request: Request, mapping: Dict[Type, Any]) -> None:
     def get_from_cache(self, request: Request, cls: Callable) -> Optional[Any]:
         return self._cached_instances.get(request, {}).get(cls)
 
-    async def __call__(
+    @inlineCallbacks
+    def __call__(
```
wRAR, line 335: What problem was happening before this change?

BurnzZ: It was when using `--reactor=asyncio` with pytest which fails the `asyncio` and `asyncio-min` toxenvs.

Specifically, it fails for all tests which uses `ItemProvider`:

```builtins.RuntimeError: Task got bad yield: <Deferred at 0x7f32afef6410>```

Comment (wRAR) [reply scrapy_poet/page_input_providers.py:335]: This sounds like https://docs.scrapy.org/en/latest/topics/asyncio.html#awaiting-on-deferreds (in which case wrapping the deferred into a Future using `maybe_deferred_to_future()` should be enough).

Comment (BurnzZ) [reply scrapy_poet/page_input_providers.py]: aha, that works! Thanks! 1bc3ee3 I was trying to use `@inlineCallbacks` similar to how the `Injector` is being set up.

Comment (BurnzZ) [reply scrapy_poet/page_input_providers.py]: Ohh no there's an issue though where it gets stuck when running the tests on deadlock.

Do you know in particular if `maybe_deferred_to_future()` raises a `RecursionError`?

Reference: https://github.com/scrapinghub/scrapy-poet/pull/129/files#diff-a38b072945ddeae120d2462383b70ddc0f40967b53c849c49d6a3cbdf1917f53R368

Comment (wRAR) [reply scrapy_poet/page_input_providers.py]: Hmm no, I have no idea about this

Comment (BurnzZ) [reply scrapy_poet/page_input_providers.py]: I'll look into some solutions for this tomorrow. Otherwise, perhaps we can revert and put on hold the change in 1bc3ee3 for the future since it was working without problems before. In this way, it wouldn't block the upcoming release.

I think the main problem is that `asyncio.Future` doesn't increase the recursion depth which leads to the `RecursionError` that we're checking here. Another solution is to create a new check other than `RecursionError` as a proxy for detecting deadlocks but this would take more time to implement and test.

File: tox.ini [hunk 4]
```diff
@@ -27,6 +23,13 @@ deps =
     url-matcher==0.2.0
     web-poet==0.7.0
 
+    # https://stackoverflow.com/a/73046084
+    Twisted==21.7.0
```
kmike, line 27: for min environment we probably want an earlier twisted, right?

scrapy_poet/page_input_providers.py
```diff
@@ -371,7 +374,7 @@ async def __call__(
                 )
 
             page_object = po_instances[page_object_cls]
-            item = await page_object.to_item()
+            item = yield from ensureDeferred(page_object.to_item())
```
kmike, line 377: Does it work with page objects which use asyncio?

tox.ini
BurnzZ: Addressed it on a92f771. However, it fails the tests for the SaveFixtureCommand with:

```
2023-02-10 15:45:32 [scrapy.crawler] INFO: Overridden settings:
{'BOT_NAME': 'foo',
 'NEWSPIDER_MODULE': 'foo.spiders',
 'ROBOTSTXT_OBEY': True,
 'SPIDER_MODULES': ['foo.spiders']}
Traceback (most recent call last):
  File "/home/kevinoxy/.pyenv/versions/3.7.2/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/kevinoxy/.pyenv/versions/3.7.2/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/kevinoxy/dev/z/scrapy-poet/.tox/min/lib/python3.7/site-packages/scrapy/cmdline.py", line 168, in <module>
    execute()
  File "/home/kevinoxy/dev/z/scrapy-poet/.tox/min/lib/python3.7/site-packages/scrapy/cmdline.py", line 145, in execute
    _run_print_help(parser, _run_command, cmd, args, opts)
  File "/home/kevinoxy/dev/z/scrapy-poet/.tox/min/lib/python3.7/site-packages/scrapy/cmdline.py", line 100, in _run_print_help
    func(*a, **kw)
  File "/home/kevinoxy/dev/z/scrapy-poet/.tox/min/lib/python3.7/site-packages/scrapy/cmdline.py", line 153, in _run_command
    cmd.run(args, opts)
  File "/home/kevinoxy/dev/z/scrapy-poet/.tox/min/lib/python3.7/site-packages/scrapy_poet/commands.py", line 95, in run
    self.crawler_process.crawl(spider_cls, url=url)
  File "/home/kevinoxy/dev/z/scrapy-poet/.tox/min/lib/python3.7/site-packages/scrapy/crawler.py", line 205, in crawl
    crawler = self.create_crawler(crawler_or_spidercls)
  File "/home/kevinoxy/dev/z/scrapy-poet/.tox/min/lib/python3.7/site-packages/scrapy/crawler.py", line 238, in create_crawler
    return self._create_crawler(crawler_or_spidercls)
  File "/home/kevinoxy/dev/z/scrapy-poet/.tox/min/lib/python3.7/site-packages/scrapy/crawler.py", line 313, in _create_crawler
    return Crawler(spidercls, self.settings, init_reactor=True)
  File "/home/kevinoxy/dev/z/scrapy-poet/.tox/min/lib/python3.7/site-packages/scrapy/crawler.py", line 82, in __init__
    default.install()
  File "/home/kevinoxy/dev/z/scrapy-poet/.tox/min/lib/python3.7/site-packages/twisted/internet/epollreactor.py", line 246, in install
    installReactor(p)
  File "/home/kevinoxy/dev/z/scrapy-poet/.tox/min/lib/python3.7/site-packages/twisted/internet/main.py", line 32, in installReactor
    raise error.ReactorAlreadyInstalledError("reactor already installed")
twisted.internet.error.ReactorAlreadyInstalledError: reactor already installed
```

I think it's due to **Twisted 18.9.0**'s default reactor value.

It could be fixed by setting `self.settings["TWISTED_REACTOR"] = "twisted.internet.epollreactor.EPollReactor"` in SaveFixtureCommand. bf1a40b

I'm not quite knowledgeable about the behavioral specifics of TWISTED\_REACTOR in Scrapy, let me know if there's a better alternative.

kmike: If I'm reading https://docs.scrapy.org/en/latest/topics/asyncio.html correctly, it means that savefixture won't work with asyncio page objects if we install EPollReactor :( @wRAR is it the case? What would you recommend?

BurnzZ: > Was the only change the Twisted version downgrade

Yes, this one in particular: a92f771

kmike:
@wRAR could ypu please pick up this PR from @BurnzZ?

BurnzZ:
We could split it in half, I could add more tests to ensure page objects using asyncio fully works ([ref](https://github.com/scrapinghub/scrapy-poet/pull/129#discussion_r1102084934)).

tox.ini
wRAR: Twisted < 21.2.0 imports (and thus installs) the reactor in `twisted.web.client`, which we import in several places in Scrapy. I don't know yet why this is not a problem for e.g. the `asyncio-pinned` Scrapy tests, maybe some init order is different.

scrapy_poet/page_input_providers.py
BurnzZ: Addressed in eaf0d40

scrapy_poet/page_input_providers.py
BurnzZ: Learned the hardway when addressing [https://github.com/scrapinghub/scrapy-poet/pull/129#discussion_r1102084934](https://github.com/scrapinghub/scrapy-poet/pull/129#discussion_r1102084934) that we really need to go into this direction. Fortunately, I've implemented a solution to solve the problem of detecting deadlocks when using asyncio's `Future` instead of Twisted's `Deferred`. eaf0d40

scrapy_poet/page_input_providers.py
```diff
@@ -329,6 +344,20 @@ def update_cache(self, request: Request, mapping: Dict[Type, Any]) -> None:
     def get_from_cache(self, request: Request, cls: Callable) -> Optional[Any]:
         return self._cached_instances.get(request, {}).get(cls)
 
+    def check_if_deadlock(self, request: Request) -> bool:
+        """Should only be used when ``AsyncioSelectorReactor`` is the reactor."""
+        if request not in self._build_instances_call_counter:
+            self._build_instances_call_counter[request] = 0
+        self._build_instances_call_counter[request] += 1
```
BurnzZ, line 351: I've thought about storing the `andi.Plan` here and checking how many times we've seen it before to see if there's a potential deadlock. However, it take up a lot of memory on some cases.

Tracking and storing a simple `int` value like this per `scrapy.Request` is much lighter on the memory.

scrapy_poet/page_input_providers.py
```diff
@@ -361,13 +390,24 @@ async def __call__(
...
+                # Need to check before awaiting on the ``asyncio.Future``
+                # before it gets stuck on a potential deadlock.
+                if asyncio.isfuture(deferred_or_future):
+                    if self.check_if_deadlock(request):
+                        raise ProviderDependencyDeadlockError(
+                            self.template_deadlock_msg.format(plan=plan)
+                        )
+
+                po_instances = await deferred_or_future
             except RecursionError:
```
BurnzZ, line 408: I'm thinking about removing this try-except block which checks for `RecursionError` and simply use `check_if_deadlock()` for both Futures and Deferreds. However, I also think that there's a much better workaround for `ayncio` that I haven't found yet. 

Might be worth having 2 ways to catch deadlocks here for Futures and Deferreds for the time being.

tox.ini
wRAR: Found it: scrapy-poet imports it earlier than Scrapy does, via `scrapy_poet/__init__.py` -> `scrapy_poet/spidermiddlewares.py` -> `scrapy.downloadermiddlewares.retry`.

Moving `from scrapy.downloadermiddlewares.retry import get_retry_request` inside `scrapy_poet.spidermiddlewares.RetryMiddleware.process_spider_exception()` seems to fix this (we could add a comment saying this is needed on Twisted < 21.2.0 only).

BurnzZ: Awesome work @wRAR ! Let me try to test this out

CHANGELOG.rst
```diff
@@ -86,8 +86,14 @@ TBR
   :class:`scrapy.Request <scrapy.http.Request>`'s callback is set to
   :func:`scrapy.http.request.NO_CALLBACK`.
 
+* Fix ``TypeError`` when using Twisted <= 21.7.0 since scrapy-poet was using
+  ``twisted.internet.defer.Deferred[object]`` type annotation before which was
+  not subscriptable.
+
 * Documentation improvements.
 
+* Various test and CI Fixes.
```
kmike, line 95: I think there are also actual bug fixes in this PR

BurnzZ: Addressed in https://github.com/scrapinghub/scrapy-poet/pull/129/commits/de56e9046a82a646fda44635c221df6e590a8bc1. I've omitted the fixes for the unreleased features like https://github.com/scrapinghub/scrapy-poet/pull/88

kmike (approved):
Looks great, thanks for spotting and fixing it @BurnzZ! And thanks @wRAR for your help with debugging!

tox.ini
```diff
@@ -27,11 +23,17 @@ deps =
     url-matcher==0.2.0
     web-poet==0.7.0
 
+    # https://github.com/scrapy/scrapy/issues/5635
+    pyopenssl==22.0.0
```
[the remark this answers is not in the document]
BurnzZ: Nice catch! It's indeed only applicable in Scrapy 2.6. Fixed it on a94cb32

wRAR (approved):

tox.ini
Gallaecio: @wRAR Should we make changes in Scrapy to avoid issues like these?

wRAR: @Gallaecio it may be useful to move `twisted.web.client` imports inside methods, if that's actually fixes this problem (so if there are no other things that break this and cannot be fixed).

Issue #128, opened 2023-02-09
TypeError: 'type' object is not subscriptable


kishan3:
More details about this error:

```
  File "/home/kishan/anaconda3/envs/data_feeds/lib/python3.10/site-packages/scrapy_poet/__init__.py", line 2, in <module>
    from .downloadermiddlewares import InjectionMiddleware
  File "/home/kishan/anaconda3/envs/data_feeds/lib/python3.10/site-packages/scrapy_poet/downloadermiddlewares.py", line 37, in <module>
    class InjectionMiddleware:
  File "/home/kishan/anaconda3/envs/data_feeds/lib/python3.10/site-packages/scrapy_poet/downloadermiddlewares.py", line 95, in InjectionMiddleware
    ) -> Generator[Deferred[object], object, Response]:
TypeError: 'type' object is not subscriptable

```

scrapy-poet = 0.5.0
web-poet = 0.7.2
scrapy = 2.6.1
python = 3.10.9

Any advice on this would be helpful. Thanks.

BurnzZ:
Hi @kishan3 , could you provide more details about your code? A minimum reproducible code would help here.

kishan3:
Hi @BurnzZ, please correct me if I am doing it wrong 

```
from scrapy import Spider
from web_poet import ItemWebPage


class AmazonProductPage(ItemWebPage):
    def to_item(self):
        return {
            "url": self.url,
        }


class AmazonProductListPage(ItemWebPage):
    def to_item(self):
        return {
            "url": self.url,
        }


class AmazonTest(Spider):
    name = "amazon_test"
    start_urls = ["https://www.amazon.com/s?k=iphone"]

    @classmethod
    def update_settings(cls, settings):
        super().update_settings(settings)
        settings["DOWNLOADER_MIDDLEWARES"]["scrapy_poet.InjectionMiddleware"] = 543
        settings["SPIDER_MIDDLEWARES"]["scrapy_poet.RetryMiddleware"] = 275
        settings["SCRAPY_POET_OVERRIDES"] = [("amazon.com", AmazonProductListPage, AmazonProductPage)]

```

BurnzZ:
I'm not what aspect of your code contributes to the `TypeError: 'type' object is not subscriptable` error you've specified. I'm guessing it's because one of your dependencies, my hunch is on **Twisted.**

Do you mind listing out all the dependency versions here?

In any case, I've tried a similar code with what you have and the following had no problems at all.

```python
[...]

PR #130, opened 2023-02-09
bump min Twisted 18.9.0 → 21.7.0

BurnzZ:
Avoids #128 from happening

I've tried all twisted versions from 18.9.0 and it seems 21.7.0 is the oldest one that makes `Deferred` subscriptable.

This is also compatible with our pinned dependency on #129, which uses `Twisted==21.7.0`. ([code reference](https://github.com/scrapinghub/scrapy-poet/pull/129/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R27)).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -80,6 +80,11 @@ TBR
 * Moved some of the utility functions from the test module into
   ``scrapy_poet.utils.testing``.
 
+* Now requires Twisted >= 21.7.0 instead of >= 18.9.0. This prevents a ``TypeError``
+  from happening since scrapy-poet uses the ``twisted.internet.defer.Deferred[object]`` 
+  type annotation in its code. On older Twisted versions, ``Deferred`` is not
+  subscriptable.
+
 * Documentation improvements.
 
 * Deprecations:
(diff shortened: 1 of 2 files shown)


wRAR (approved):

CHANGELOG.rst
```diff
@@ -80,6 +80,11 @@ TBR
 * Moved some of the utility functions from the test module into
   ``scrapy_poet.utils.testing``.
 
+* Now requires Twisted >= 21.7.0 instead of >= 18.9.0. This prevents a ``TypeError``
+  from happening since scrapy-poet uses the ``twisted.internet.defer.Deferred[object]``
```
kmike, line 84: A more conservative fix would be to not use this type annotation. Is it essential?

BurnzZ: Yes, I think that's a better solution.

I'll make this change instead in #129 since it adds another type annotation using `Deferred`. Commit: efb6476

BurnzZ:
Closing in favor of efb6476 which was made in #129

BurnzZ:
Closing in favor of efb6476 which was made in #129

PR #133, opened 2023-02-27
Only run to_item() under the time machine

wRAR:
As initializing and downloading is no longer done while the time machine time is ticking, this should help greatly with time differences in the field values.

On the other hand, I'm now not sure how will we be able to switch to running real project spiders in savefixture instead of `InjectableSpider`, maybe it will be possible by moving the time machine part into some deeper scrapy-poet code.

### DIFF
--- a/scrapy_poet/commands.py
+++ b/scrapy_poet/commands.py
@@ -1,5 +1,4 @@
 import datetime
-from inspect import iscoroutinefunction
 from pathlib import Path
 from typing import Type
 
@@ -15,6 +14,7 @@ from scrapy.utils.misc import load_object
 from twisted.internet.defer import inlineCallbacks
 from web_poet import ItemPage
 from web_poet.testing import Fixture
+from web_poet.utils import ensure_awaitable
 
 from scrapy_poet import DummyResponse
 from scrapy_poet.downloadermiddlewares import DEFAULT_PROVIDERS, InjectionMiddleware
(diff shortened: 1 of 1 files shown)


wRAR:
Looks like the asyncio test problem only happens with newer tox? And it's unrelated to this PR.

scrapy_poet/commands.py
```diff
@@ -93,9 +105,9 @@ def run(self, args, opts):
         frozen_time = datetime.datetime.now(datetime.timezone.utc).replace(
```
kmike, line 105: Are we rewinding the time back, because frozen_time is frozen before the spider runs? Would it be better to compute frozen_time in the callback as well?

kmike:
> maybe it will be possible by moving the time machine part into some deeper scrapy-poet code

See also: https://github.com/scrapinghub/scrapy-poet/pull/127#discussion_r1115701815

scrapy_poet/commands.py
wRAR: I think it doesn't matter where is the value calculated (it shouldn't matter what is the actual value as long as it's saved into meta), but I think it's easier to pass the value into the spider than outside it?

kmike: It's very minor, but there could be some strange cases. 

For example, you downloaded the response, and response contains time information. In the current implementation we freeze time only after the download, but the frozen_time is set before the download, so they don't match; it means that the body will contain time which is in future. If you want to do something with this time, you may get negative offsets, or it may cause some other issues. I don't know, maybe there is some http header (like Expires), which would contain incorrect information with the current implementation. 

Again, quite an edge case, and it's just seconds, but the right thing to do seems to be to have frozen_time to match the time on the clock.

wRAR: Yeah, makes sense. Though we still need to remove the microseconds part in case the time ticks by some small sub-second amount but the seconds value is still increased.

PR #135, opened 2023-03-03
Real spider support for savefixture

wRAR:
This may be considered a proof of concept but may also be merged I think.

`start_requests` is assigned in `__init__` in case the spider `__init__` overrides it. Still, it may be possible to find real spiders that won't work because they use some magic that isn't covered here. 

It's possible that there is a better approach to this, but it's hard to design one without seeing many different real spiders that will be used with this, so maybe we will need to redesign it later.


Gallaecio (approved):

scrapy_poet/commands.py
```diff
@@ -53,34 +52,44 @@ def __init__(self, crawler: Crawler) -> None:
...
 
         async def cb(self, response: DummyResponse, page: injectable):  # type: ignore[valid-type]
             global frozen_time
             frozen_time = datetime.datetime.now(datetime.timezone.utc).replace(
                 microsecond=0
             )
             with time_machine.travel(frozen_time):
-                yield await ensure_awaitable(page.to_item())  # type: ignore[attr-defined]
+                item = await ensure_awaitable(page.to_item())  # type: ignore[attr-defined]
+            saved_items.append(item)
```
BurnzZ, line 78: Would it be worth it bringing back `SavingPipeline` to make it less fragile on cases like a custom middleware is enabled in the custom settings which replaces the callback method that was originally passed in `start_requests()`?

BurnzZ: Aha, you're right. 👍

BurnzZ (approved):

PR #136, opened 2023-03-06
add .readthedocs.yml config

BurnzZ:
Fixes failing build in [https://readthedocs.org/projects/scrapy-poet/builds/19661918/](https://readthedocs.org/projects/scrapy-poet/builds/19661918/) since it's not using Python 3.7 which `async-lru` doesn't support

Related to https://github.com/scrapinghub/web-poet/pull/147

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,7 +2,6 @@ version: 2
 formats: all
 sphinx:
   configuration: docs/conf.py
-  fail_on_warning: true
 
 build:
   os: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


.readthedocs.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+formats: all
+sphinx:
+  configuration: docs/conf.py
+  fail_on_warning: true
+
+build:
+  os: ubuntu-22.10
+  tools:
+    python: "3.7"  # use min version for async-lru==1.0.3
```
kmike, line 10: Why not use recent Pythion to build docs?

BurnzZ: Right, I was under the impression we need to ensure Python 3.7 builds works correctly. Updated in dc45a2e.

PR #137, opened 2023-03-09
Fix issue 99

### LINKED ISSUE #99: DummyResponse.replace is broken
Currently DummyResponse only accepts url and request arguments. But the `replace` method (inherited from Response) passes all the attributes to `__init__`, which causes an exception.

FilippoPetroli:
Resolves #99


FilippoPetroli:
Hi @kmike 
I fixed the issue. I saw that there are some silenced test that mention DummyResponse so maybe there are edges to trim. I wrote a basic test but if you prefer, I can write more tests.
Thanks

cc @felipeboffnunes

tests/test_downloader.py
```diff
@@ -501,6 +501,14 @@ def parse(self, response: DummyResponse):
     assert not isinstance(collected["response"], DummyResponse)
 
 
+def test_dummy_response_multiple_args() -> None:
+    dummy = DummyResponse(url="https://google.com", request=None, status=300)
+    assert dummy.status == 300
+
+    dummy = DummyResponse("https://google.com", 400)
+    assert dummy.status == 400
```
kmike, line 509: Hey! The original issue was that .replace doesn't work; could you please add a test for this, or maybe even change this test to check .replace? 

Users shouldn't be creating DummyResponse instances themselves, but they might have code which calls .replace.

kmike:
Looks good overall, thanks @FilippoPetroli!

AdrianAtZyte (approved):

PR #139, opened 2023-04-05
Tests for savefixture with web-poet exceptions

wRAR:
- [x] Test for Retry
- [ ] Test for UseFallback

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -7,11 +7,16 @@ from pathlib import Path
 
 from web_poet.testing import Fixture
 
+from scrapy_poet.utils.mockserver import MockServer
+from scrapy_poet.utils.testing import EchoResource
+
 
 def call_scrapy_command(cwd: str, *args: str) -> None:
     with tempfile.TemporaryFile() as out:
         args = (sys.executable, "-m", "scrapy.cmdline") + args
         subprocess.call(args, stdout=out, stderr=out, cwd=cwd)
+        out.seek(0)
+        pass
 
 
 def test_savefixture(tmp_path) -> None:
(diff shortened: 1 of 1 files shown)


wRAR:
With Retry the fixture contains the resulting item and the successful response.

wRAR:
I've added a test for UseFallback where the exception is caught by the spider. It fails because savefixture doesn't use the spider callback (and it wouldn't be right for savefixture to use data not produced by the PO itself?).

wRAR:
These test the generation of tests but not running them. Running them should fail in both cases because in both cases the exception handling code is external to the page object: it's in a middleware and a callback respectively.

wRAR:
The unrelated test failure is because 

```python

    """The case wherein a PageObject inherits directly from ``web_poet.Injectable``
    should also work, provided that the ``to_item`` method is similar with
    ``web_poet.ItemPage``:
    """
    item, deps = yield crawl_item_and_deps(ProductFromInjectable)
    assert item == ProductFromInjectable(name="product from injectable")
```

now breaks with "AttributeError: 'ProductFromInjectablePage' object has no attribute '_validate_input'" (so I think the assumption in the comment is no longer true).

wRAR:
Alternatively, we could instead make tests that check that `to_item()` raises Retry/UseFallback?

wRAR:
It was decided that web-poet needs to support having tests that check "this PO with this input raises `UseFallback`", then scrapy-poet needs to be able to generate it. Keeping this PR until the web-poet change is merged.

wRAR:
This is now waiting for a web-poet release with https://github.com/scrapinghub/web-poet/pull/166

Gallaecio (approved):

PR #143, opened 2023-04-26
Fix scrapy shell with InjectionMiddleware

### LINKED ISSUE #140: scrapy shell doesn't work with InjectionMiddleware
Currently using `scrapy shell` to fetch a URL causes the following exception:

```
  File ".env/lib/python3.10/site-packages/scrapy/shell.py", line 119, in fetch
    response, spider = threads.blockingCallFromThread(
  File ".env/lib/python3.10/site-packages/twisted/internet/threads.py", line 120, in blockingCallFromThread
    result.raiseException()
  File ".env/lib/python3.10/site-packages/twisted/python/failure.py", line 475, in raiseException
    raise self.value.with_traceback(self.tb)
TypeError: issubclass() arg 1 must be a class
```

This hides the real traceback which points to https://github.com/scrapinghub/scrapy-poet/blob/ee1357fe84465840c7d6ec1b6a0add8e622611a7/scrapy_poet/injection.py#L317. It calls `issubclass(first_parameter.annotation, DummyResponse)` where `first_parameter` is the callback first argument, but `scrapy shell` uses `Deferred.callback` as the request callback: https://github.com/scrapy/scrapy/blob/98571eb946e24edfe5b520c0478e72b695d09a9d/scrapy/shell.py#L207 and its first argument is of type `Union[_DeferredResultT, Failure]`. There are probably multiple way to fix this.

wRAR:
Fixes #140

### DIFF
--- a/scrapy_poet/injection.py
+++ b/scrapy_poet/injection.py
@@ -338,6 +338,10 @@ def is_callback_requiring_scrapy_response(
         # There's no type annotation, so we're probably using response here.
         return True
 
+    if not isinstance(first_parameter.annotation, type):
+        # The annotation is using something like Union, so we're probably using response here.
+        return True
+
     if issubclass(first_parameter.annotation, DummyResponse):
         # See: https://github.com/scrapinghub/scrapy-poet/issues/48
         # See: https://github.com/scrapinghub/scrapy-poet/issues/118
--- a/scrapy_poet/injection.py
+++ b/scrapy_poet/injection.py
@@ -6,6 +6,7 @@ import warnings
 from typing import Any, Callable, Dict, List, Mapping, Optional, Set
 
 import andi
+from andi.typeutils import issubclass_safe
 from scrapy import Request, Spider
 from scrapy.crawler import Crawler
 from scrapy.http import Response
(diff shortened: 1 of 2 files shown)


scrapy_poet/injection.py
```diff
@@ -338,6 +338,10 @@ def is_callback_requiring_scrapy_response(
         # There's no type annotation, so we're probably using response here.
         return True
 
+    if not isinstance(first_parameter.annotation, type):
+        # The annotation is using something like Union, so we're probably using response here.
```
kmike, line 342: It took me some time to underdtand the comment, and the logic. I think the point is that the only case when we should skip the response is when the first argument is annotated exactly as DumyResponse. The mention of Union confused me a lot :) Would it make sense to use https://github.com/scrapinghub/andi/blob/c5a6daae013deabf5c215e87ac1fe671cad665dd/andi/typeutils.py#L28?

wRAR: "like issubclass, but return False if cls is not a class, instead of raising an error" is exactly what we need here :)

wRAR:
Hmm the new test failure is reliably reproduced with the last change.

wRAR:
The actual error (concealed by the large error output from the deadlock tests that doesn't fail tests and I think is old) is https://github.com/pytest-dev/pytest-cov/issues/529

I cannot reproduce it locally, it also disappeared on the third try.

PR #145, opened 2023-05-03
Support adapters in fixtures

wRAR:


docs/testing.rst
```diff
@@ -49,3 +48,12 @@ to schedule requests or modify items returned from
...
 is set to ``True`` and using more simple logic in this case.
+
+.. _fixture-adapter:
+
+Configuring the item adapter
+============================
+
+As documented in :ref:`web-poet:web-poet-testing-adapters`, fixtures can use custom
+item adapters. ``savefixture`` will use the adapter specified in the
+``SCRAPY_POET_TESTS_ADAPTER`` setting when saving the fixture.
```
kmike, line 59:
```suggestion
item adapters. ``savefixture`` command uses the adapter specified in the
``SCRAPY_POET_TESTS_ADAPTER`` setting to save the fixture.
```

tests/test_commands.py
```diff
@@ -167,3 +164,59 @@ def to_item(self):
...
+            "savefixture",
+            type_name,
+            f"{server.root_url}",
+        )
+    fixtures_dir = cwd / "fixtures"
+    fixture_dir = fixtures_dir / type_name / "test-1"
+    fixture = Fixture(fixture_dir)
+    assert fixture.is_valid()
+    item = json.loads(fixture.output_path.read_bytes())
+    assert item == {"name": "chocolate"}
```
kmike, line 222: It seems we're not checking the item in other tests; would it make sense to check it, at least to ensure that the chocolate is not lower-case originally?

kmike: Alternatively, you can just return a constant from the page object, without using mock server.

docs/testing.rst
Gallaecio: Mind that, if you append “command”, you need to prepend an article as well: `savefixture` → _the_ `savefixture` command.

docs/settings.rst
```diff
@@ -105,3 +105,14 @@ Default: ``fixtures``
...
 
 More info at :ref:`testing`.
+
+
+SCRAPY_POET_TESTS_ADAPTER
+-------------------------
+
+Default: ``None``
+
+Sets the class, or it's import path, that will be used as an adapter in the
```
Gallaecio, line 115:
```suggestion
Sets the class, or its import path, that will be used as an adapter in the
```

Gallaecio (approved):

Gallaecio (approved):

PR #148, opened 2023-05-31
Issue 105 web poet serialization in caching

### LINKED ISSUE #105: Switch the cache to web_poet.serialization
We've investigated the possibility to use `web_poet.serialization` for the scrapy-poet cache, either by replacing the existing implementation completely (it would include replacing the API as currently providers are responsible for (de)serialization) or by only using the web-poet code to (de)serialize `HttpResponse`. Currently in scrapy-poet only `HttpResponseProvider` supports caching, but the web-poet implementation supports more dependency types and so replacing the implementation completely looks useful but isn't straightforward. The unanswered questions I see so far:

1. If we e.g. write a web-poet-based replacement of `SqlitedictCache` to store just `HttpResponse` (if it's possible), will it help as a first step in the migration? I suspect that it won't.
2. If we need to change the provider API, by changing or removing `CacheDataProviderMixin`, what will this break?
3. What should be the keys in the new cache? Currently the key includes the provider name and the request fingerprint calculated by the provider.
4. We may need to change/add something at the web-poet side, e.g. add a function that returns a list of deps instead of returning a reconstructed PO (which `serialize()` does). We may also need compressed storage support, we already have a place to implement pluggable storages (as alternatives compatible with `SerializedDataFileStorage`).
5. Also changing the caching interface requires rewriting of `Injector.build_instances_from_providers`, though this part may actually be easy when the new interface is clear.

lampssy:
Resolves #105 

I think this is more or less what we need to switch to a web-poet based caching. As for now, the tests are failing, as I didn't fully adjust them to the new cache. It would be good to get some insights whether the new implementation makes sense, before I proceed further.


scrapy_poet/injection.py
```diff
@@ -72,25 +73,22 @@ def load_providers(self, default_providers: Optional[Mapping] = None):  # noqa:
...
-                get_scrapy_data_path(createdir=True), "scrapy-poet-cache.sqlite3"
-            )
-        if cache_filename:
-            compressed = self.crawler.settings.getbool("SCRAPY_POET_CACHE_GZIP", True)
+        self.cache = {}
+        cache_path = self.crawler.settings.get("SCRAPY_POET_CACHE")
+
+        # SCRAPY_POET_CACHE: True
+        if cache_path and isinstance(cache_path, bool):
+            cache_path = get_scrapy_data_path(createdir=True)
```
wRAR, line 82: I wonder if this should have some more specific name, like the old path including "scrapy-poet-cache" in the file name.

wRAR (commented):
Overall I think this approach is fine but I may be missing some special cases.

scrapy_poet/cache.py
```diff
@@ -20,54 +20,46 @@ def close(self) -> None:  # noqa: B027
...
-        self.db[fingerprint] = value
+    def _get_exception_file_path(self, fingerprint: str) -> str:
+        """Save exception inside self.directory, so that `storage.read()` can read it correctly"""
+        return os.path.join(
+            self._get_directory_path(fingerprint), f"{fingerprint}.error"
+        )
 
-    def close(self) -> None:
-        self.db.close()
+    # TODO: Add option for compressed cache
```
Gallaecio, line 65: @kmike @wRAR Do we want to implement this here or leave it for a future iteration?

Gallaecio (commented):
Amazing work so far!

scrapy_poet/cache.py
wRAR: I think it should be done separately and I see it as some kind of pluggable `SerializedDataFileStorage` implementations in web-poet.

wRAR:
> test_name_first_spider() fails because even though the cache is hit the spider gets the non-cached response

This is because test providers used by this test previously implemented the caching mixin and so returned their provided deps directly from the cache.

scrapy_poet/page_input_providers.py
```diff
@@ -133,6 +122,17 @@ def is_provided(self, type_: Callable) -> bool:
                 f"{self!r}. Expected either 'set' or 'callable'"
             )
 
+    @property
+    def has_cache_support(self):
```
wRAR, line 126: This, and the code that uses it, assumes that if one dep is serializable then all of them are. Maybe we even don't need this function and should just check the deps directly when deciding whether to tryu caching them.

wRAR:
I've fixed the tests that were failing.

lampssy:
I added a more specific name for cache folder and removed checks whether providers have cache support - it looks like this is not needed.

@wRAR @Gallaecio let me know if you think any more adjustments needed.

Gallaecio (approved):
Seems good to me.

wRAR (approved):
Thanks!

PR #149, opened 2023-06-02
Replace example.com domains with Scrapinghub demo toscrape.com domains in the documentation

ankitjavalkar:


ankitjavalkar:
I am little concerned that the changes may dilute the context or the relevance of the examples. Need a second opinion on this.

docs/intro/advanced-tutorial.rst
```diff
@@ -62,7 +62,7 @@ Suppose we have the following Page Object:
 
             # Simulates clicking on a button that says "View All Images"
             response: web_poet.HttpResponse = await self.http.get(
-                f"https://api.example.com/v2/images?id={item['product_id']}"
+                f"https://api.toscrape.com/v2/images?id={item['product_id']}"
```
kmike, line 65: This is not a working URL; is it possible to have a real example? I think the idea to use toscrape.com domain is to actually make the examples runnable.

ankitjavalkar: I agree @kmike  . Though, the situation here is that we do not have a toscrape.com website that supports API calls. Now there are websites that support dummy/test API calls but the mandate of the issue was that we should not be using other external domains.

Can we setup an API based toscrape website?

kmike: Not sure how it fits the example, but I think http://quotes.toscrape.com/scroll example uses an AJAX call.

Gallaecio (approved):

PR #151, opened 2023-06-15
Remove ItemProvider’s Response dependency

### LINKED ISSUE #150: ItemProvider always requires Response
Currently ItemProvider always requires Response. So, if a page object doesn't need a response, and an item produced by this page object is used as a dependency, then Response is still downloaded.

Gallaecio:
Resolves #150.


Gallaecio:
I’ve opened a [separate PR for the docs fix](https://github.com/scrapinghub/scrapy-poet/pull/152).

tests/test_providers.py
```diff
@@ -248,13 +248,20 @@ def test_page_params_provider(settings):
...
+    crawler = get_crawler(Spider, settings)
+    injector = Injector(crawler)
+    provider = ItemProvider(injector)
+    request = scrapy.http.Request("https://example.com")
 
-    We'll only test its caching behavior here if its properly garbage collected.
-    """
+    # The fact that no exception is raised below proves that a Response
+    # parameter is not required by ItemProvider.
+    provider(set(), request)
```
kmike, line 261: Hey! Are there existing tests which ensures that 

1) the original issue is fixed, and 
2) some potential new issues don't appear?

Regarding (2), I was thinking about the following:

```py
class MySpider(scrapy.Spider):
    def parse(self, response: DummyResponse, item: Product):
        # ....


@handle_urls("example.com")
class MyPage(ItemPage[Product]):
    response: HttpResponse
```

i.e. we start to pass DummyResponse, but a real response is needed to create a page object which returns an item.

Gallaecio: OK, so you definitely picked up on an issue, and the “solution” I came up with is getting messy, so I would like to discuss it before I move forward further with it, because I might be missing a better solution.

The “solution” consists of having 2 separate item provider classes, one for responseless items and one for responseful items.

Things get more complicated, though. To properly determine if an item needs a response, we need to get the page object for the item, and then check if the dependencies of that page object (which might include other items) are provided by a provider that requires a response. Moreover, we need to take the request (URL) into account, as that can determine which page object is used for an item.

To be honest, it kind of feels like there should be no item provider, just as there is no page object provider, and instead item resolution should be moved closer to the core, and work the same as page object resolution, by somehow making andi realize how to resolve item dependencies. But I am not very familiar with the code base, and I afraid of wasting too much time exploring in that direction.

wRAR: > it kind of feels like there should be no item provider, just as there is no page object provider, and instead item resolution should be moved closer to the core, and work the same as page object resolution, by somehow making andi realize how to resolve item dependencies.

Yeah, that would solve https://github.com/scrapy-plugins/scrapy-zyte-api/issues/91 automatically (AFAIK).

Gallaecio:
Unnecessary thanks to https://github.com/scrapinghub/scrapy-poet/pull/175

PR #155, opened 2023-08-28
Drop python 3.7 support.

bulatbulat48:


tox.ini
```diff
@@ -22,9 +22,10 @@ deps =
     sqlitedict==1.5.0
     url-matcher==0.2.0
     web-poet==0.7.0
+    Twisted<=22.10.0
```
Gallaecio, line 25: This only affects pinned environments, and in fact ideally we should have the lowest Twisted version fixed here. Where you need to add this change is in [testenv]deps above.

bulatbulat48: Fixed, should I set lowest Twised version here?

Gallaecio (approved):

PR #159, opened 2023-09-11
Stats

Gallaecio:
Follow-up to https://github.com/scrapinghub/web-poet/pull/175.

I wonder if we should include some prefix on stats, like `scrapy_poet/`, `scrapy_poet/<page object import path>/`, or , `scrapy_poet/<page object import path>/stats/`.

### DIFF
--- a/docs/stats.rst
+++ b/docs/stats.rst
@@ -5,4 +5,4 @@ Stats
 =====
 
 scrapy-poet tracks :external+web-poet:ref:`web-poet stats <stats>` as part of
-:ref:`Scrapy stats <topics-stats>`, prefixed with ``scrapy-poet/stats/``.
+:ref:`Scrapy stats <topics-stats>`, prefixed with ``poet/stats/``.
--- a/docs/stats.rst
+++ b/docs/stats.rst
@@ -1,0 +1,8 @@
+.. _stats:
+
+=====
+Stats
+=====
+
+scrapy-poet tracks :external+web-poet:ref:`web-poet stats <stats>` as part of
+:ref:`Scrapy stats <topics-stats>`, prefixed with ``scrapy-poet/stats/``.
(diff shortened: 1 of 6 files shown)


kmike:
> I wonder if we should include some prefix on stats

+1 to add a prefix. The main question is should we include page object name or not, i.e. should we aggregate stats across page objects. As the page object may include its name to the stat name if needed, I think a good default is not to include page object name to the prefix, just use scrapy-poet or scrapy_poet (what's the common practice?) 

As a complete overkill solution, we can have a setting with the stats prefix, and template support, so that user can change the prefix, include/exclude page object name if needed, or even move the stats to the top level. It can even be a list of string templates, so that you can have both per-page-object stats and aggregate stats. But I'd not go there in this PR :)

scrapy_poet/page_input_providers.py
```diff
@@ -333,3 +335,31 @@ async def __call__(
 
             results.append(item)
         return results
+
+
+class CrawlerStatCollector(StatCollector):
```
kmike, line 340: ScrapyPoetStatsCollector?

kmike: Although I got it now, why it's called CrawlerStatsCollector: it uses crawler.stats

Gallaecio:
I’ll *undraft* as soon as the (internal) -/_ battle for the prefix is over.

docs/stats.rst
```diff
@@ -0,0 +1,8 @@
+.. _stats:
+
+=====
+Stats
+=====
+
+scrapy-poet tracks :external+web-poet:ref:`web-poet stats <stats>` as part of
+:ref:`Scrapy stats <topics-stats>`, prefixed with ``poet/stats/``.
```
kmike, line 8: Last thing to check: let's see how it looks like with "standard stats", like the number of dependencies provided:

```
"poet/stats/my-custom-1": 15,
"poet/stats/my-custom-2": 15,
"poet/dependencies/MyPageObject": 15,
"poet/dependencies/HttpResponse": 15,
```

Seems fine to me, so +1 to merge

kmike (approved):

PR #163, opened 2023-09-15
Cover 0.15.1 in the release notes

Gallaecio:


wRAR (approved):

CHANGELOG.rst
```diff
@@ -2,6 +2,15 @@
 Changelog
 =========
 
+0.15.1 (2023-09-15)
+-------------------
```
kmike, line 6: Given that this release contains a new feature and a backwards incompatible change, it would have been better to release it as 0.16.0 :) Not a big deal though.

Gallaecio: Yes, I was not sure whether or not to consider stat name changes backward-incompatible, but they indeed can be, and anyone tracking them with Spidermon might get their expectations broken :facepalm:

PR #164, opened 2023-09-18
Handle item overrides

BurnzZ:
Fixes the case when the following code below results in a `ProviderDependencyDeadlockError`. In this example, the issue stems from the `product: Product` dependency that's declared. However, **scrapy-poet** sees `ProductPage` as the one that provides it which causes the deadlock error.

```python
import attrs
from zyte_common_items import Product
from web_poet import handle_urls, WebPage

@handle_urls("example.com")
@attrs.define 
class ProductPage(WebPage[Product]):
    product: Product

    def name(self) -> str:
        return f"(modified) {self.product.name}"
```

The use case for this is when items like `Product` are produced by a provider and not a page object. For example, **scrapy-zyte-api**'s `ZyteApiProvider` ([reference](https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/scrapy_zyte_api/providers.py#L27-L39)). The `ProductPage` in this scenario tries to override the item produced by `ZyteApiProvider`.

---

**TODO:**

- [x] add more test cases to handle different scenarios
- [x] verify if this is the best implementation to go

### DIFF
--- a/scrapy_poet/downloadermiddlewares.py
+++ b/scrapy_poet/downloadermiddlewares.py
@@ -34,7 +34,7 @@ DEFAULT_PROVIDERS = {
     PageParamsProvider: 700,
     RequestUrlProvider: 800,
     ResponseUrlProvider: 900,
-    ItemProvider: 1000,
+    ItemProvider: 2000,
 }
 
 InjectionMiddlewareTV = TypeVar("InjectionMiddlewareTV", bound="InjectionMiddleware")
--- a/scrapy_poet/injection.py
+++ b/scrapy_poet/injection.py
@@ -151,11 +151,20 @@ class Injector:
         )
 
     @inlineCallbacks
-    def build_instances(self, request: Request, response: Response, plan: andi.Plan):
+    def build_instances(
+        self,
+        request: Request,
+        response: Response,
+        plan: andi.Plan,
+        prev_instances: Optional[Dict] = None,
+    ):
         """Build the instances dict from a plan including external dependencies."""
         # First we build the external dependencies using the providers
         instances = yield from self.build_instances_from_providers(
-            request, response, plan
+            request,
+            response,
+            plan,
+            prev_instances,
         )
         # All the remaining dependencies are internal so they can be built just
         # following the andi plan.
@@ -169,10 +178,14 @@ class Injector:
 
     @inlineCallbacks
     def build_instances_from_providers(
-        self, request: Request, response: Response, plan: andi.Plan
+        self,
+        request: Request,
+        response: Response,
+        plan: andi.Plan,
+        prev_instances: Optional[Dict] = None,
     ):
         """Build dependencies handled by registered providers"""
-        instances: Dict[Callable, Any] = {}
+        instances: Dict[Callable, Any] = prev_instances or {}
         scrapy_provided_dependencies = self.available_dependencies_for_providers(
             request, response
         )
@@ -182,7 +195,11 @@ class Injector:
             provided_classes = {
                 cls for cls in dependencies_set if provider.is_provided(cls)
             }
-            provided_classes -= instances.keys()  # ignore already provided types
+
+            # ignore already provided types if provider doesn't need to use them
+            if not provider.allow_prev_instances:
+                provided_classes -= instances.keys()
+
             if not provided_classes:
                 continue
 
@@ -221,6 +238,8 @@ class Injector:
                     externally_provided=scrapy_provided_dependencies,
                     full_final_kwargs=False,
                 ).final_kwargs(scrapy_provided_dependencies)
+                if provider.allow_prev_instances:
+                    kwargs.update({"prev_instances": instances})
                 try:
                     # Invoke the provider to get the data
                     objs = yield maybeDeferred_coro(
--- a/scrapy_poet/injection.py
+++ b/scrapy_poet/injection.py
@@ -441,6 +441,7 @@ def get_injector_for_testing(
     spider = MySpider()
     spider.settings = settings
     crawler.spider = spider
+    crawler.stats = load_object(crawler.settings["STATS_CLASS"])(crawler)
     if not registry:
         registry = create_registry_instance(RulesRegistry, crawler)
     return Injector(crawler, registry=registry)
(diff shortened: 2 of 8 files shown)


scrapy_poet/page_input_providers.py
```diff
@@ -106,6 +106,8 @@ def __call__(self, to_provide, response: Response):
     provided_classes: Union[Set[Callable], Callable[[Callable], bool]]
     name: ClassVar[str] = ""  # It must be a unique name. Used by the cache mechanism
 
+    allow_prev_instances: bool = False
```
BurnzZ, line 109: TODO: Redo this approach since it's quite messy.

tests/test_web_poet_rules.py
```diff
@@ -1489,6 +1493,59 @@ def test_page_object_with_item_dependency_deadlock_2_d(caplog) -> None:
...
+# provided by:
+# - some provider
+# - some page objects in their ApplyRule
+
+# TODO: Same as the previous tests but multiple ApplyRules are available and top one is selected
+
+# TODO: Same as the 1st test but the settings is not overridden
+
+# TODO: Having the 2 new test cases above, update ``test_created_apply_rules``
+# below.
```
BurnzZ, line 1546: TODO: address these remaining test scenarios.

scrapy_poet/downloadermiddlewares.py
```diff
@@ -34,7 +34,7 @@
     PageParamsProvider: 700,
     RequestUrlProvider: 800,
     ResponseUrlProvider: 900,
-    ItemProvider: 1000,
+    ItemProvider: 2000,
```
BurnzZ, line 37: Increasing this value allows `ZyteApiProvider` _(as well as any other item provider in the future)_ to run before scrapy-poet's `ItemProvider`. This ensures any item dependency in page objects are available.

scrapy_poet/page_input_providers.py
```diff
@@ -306,32 +310,50 @@ async def __call__(
...
+
+                    po_instances = await deferred_or_future
+                except RecursionError:
+                    raise ProviderDependencyDeadlockError(
+                        self.template_deadlock_msg.format(plan=plan)
+                    )
+
+            # See if the recursive calls produced an item and put it in the cache.
+            # Otherwise, ``to_item()`` will be called multiple times which can
+            # lead to bad values in some fields.
```
BurnzZ, line 350: TODO: There's probably a better way to detect and prevent items from recursive calls from bubbling up.

Refactor this when the other test cases are already in place.

scrapy_poet/injection.py
```diff
@@ -169,10 +178,14 @@ def build_instances(self, request: Request, response: Response, plan: andi.Plan)
...
-        self, request: Request, response: Response, plan: andi.Plan
+        self,
+        request: Request,
+        response: Response,
+        plan: andi.Plan,
+        prev_instances: Optional[Dict] = None,
     ):
         """Build dependencies handled by registered providers"""
-        instances: Dict[Callable, Any] = {}
+        instances: Dict[Callable, Any] = prev_instances or {}
```
BurnzZ, line 188: Passing the prev instances allows the PO to gain access to the item produced by a provider.

```python
@attrs.define 
class ProductPage(WebPage[Product]):
    product: Product
```

In the above example, before the `ProductPage` instance can be made, the `product: Product` instance should first be created. This is where this passing of new `prev_instances` parameter is useful.

scrapy_poet/injection.py
```diff
@@ -414,6 +441,7 @@ class MySpider(Spider):
     spider = MySpider()
     spider.settings = settings
     crawler.spider = spider
+    crawler.stats = load_object(crawler.settings["STATS_CLASS"])(crawler)
```
BurnzZ, line 444: Scrapy 2.11 changes broke this since `stats` was missing.

tests/test_web_poet_rules.py
BurnzZ: Done in https://github.com/scrapinghub/scrapy-poet/pull/164/commits/d148db2ddc0ad311b795dea810f4230e901134d3

scrapy_poet/page_input_providers.py
BurnzZ: Done in https://github.com/scrapinghub/scrapy-poet/pull/164/commits/88cdc5e47db35bdaf21a45b5d23eae17d8efa6cd

scrapy_poet/page_input_providers.py
BurnzZ: Looks like it's a good approach.

tests/test_web_poet_rules.py
```diff
@@ -1467,28 +1471,148 @@ def test_page_object_with_item_dependency_deadlock_2_a(caplog) -> None:
...
+
+    The PO with the higher priority should be the one returning the item.
+    """
+
+    settings = {
+        "SCRAPY_POET_PROVIDERS": {**DEFAULT_PROVIDERS, **{KangarooProvider: 1000}}
+    }
+    # item from 'to_return'
+    item, deps = yield crawl_item_and_deps(Kangaroo, override_settings=settings)
+    assert item == Kangaroo(name="(modified by Joey) data from KangarooProvider")
```
BurnzZ, line 1603: Notice that only `JoeyPage` had touched the data and `KangarooPage` didn't because the earlier PO had higher priority.

scrapy_poet/injection.py
wRAR: Ouch, I've already fixed it in master before looking here.

wRAR: But it's good to see we did it in (kinda) the same way!

wRAR (approved):

Gallaecio (approved):

Gallaecio:
Conflicts resolved, but https://github.com/john-kurkowski/tldextract/issues/305 breaks tests through url-matcher’s get_domain, will fix that now.

Gallaecio:
I wonder if https://github.com/zytedata/zyte-crawlers/pull/12#issuecomment-1729658504 could be caused by this change.

Gallaecio:
I need to investigate a potential issue with `DummyResponse` not working on code that relies on this branch.

Gallaecio (approved):

kmike:
Thanks @BurnzZ and @Gallaecio!

PR #167, opened 2023-09-26
Add release notes for 0.16.0.

wRAR:


CHANGELOG.rst
```diff
@@ -2,6 +2,15 @@
 Changelog
 =========
 
+0.16.0 (YYYY-MM-DD)
+-------------------
+
+* Now requires ``time_machine >= 2.2.0``.
+
+* Add support for having the same item type as a dependency and a result.
```
wRAR, line 10: Feel free to suggest some expanded wording.

Gallaecio: This is the best I could come up with, but still feels kind of confusing :sweat: 

```suggestion
* ``ItemProvider`` now supports page objects that declare a dependency on the 
  same type of item that they return, as long as there is an earlier page 
  object input provider that can provide such dependency.
```

Gallaecio (approved):

kmike (approved):
+1 to merge after changelog rewording is applied.

PR #169, opened 2023-10-11
Initial support for typing.Annotated.

wRAR:
Requires https://github.com/scrapinghub/andi/pull/25


tests/test_injection.py
```diff
@@ -37,7 +38,13 @@ def __init__(self, crawler):
             self.crawler = crawler
 
         def __call__(self, to_provide):
-            return [cls(content) if content else cls() for cls in classes]
+            result = []
+            for cls in to_provide:
+                obj = cls(content) if content else cls()
+                if metadata := getattr(cls, "__metadata__", None):
+                    obj.__metadata__ = metadata
```
wRAR, line 45: We need the same change in all real providers if we want to support annotating their provided classes.

scrapy_poet/injection.py
```diff
@@ -261,7 +264,16 @@ def build_instances_from_providers(
...
-            objs_by_type: Dict[Callable, Any] = {type(obj): obj for obj in objs}
+            objs_by_type: Dict[Callable, Any] = {}
+            for obj in objs:
+                cls = type(obj)
+                if sys.version_info >= (3, 9) and (
+                    metadata := getattr(obj, "__metadata__", None)
+                ):
+                    from typing import Annotated
+
+                    cls = Annotated[cls, *metadata]
```
wRAR, line 275: Ouch, looks like this is invalid syntax in <3.11.

tests/test_injection.py
wRAR: Another problem with this is when `obj` has `__slots__` and this breaks.

wRAR: As we only return instances from the provider's `__call__`, and we need to specify which of them had which annotations, an alternative to this code would be changing the `__call__` interface to e.g. return a tuple.

tests/test_injection.py
```diff
@@ -37,7 +39,13 @@ def __init__(self, crawler):
             self.crawler = crawler
 
         def __call__(self, to_provide):
-            return [cls(content) if content else cls() for cls in classes]
+            result = []
+            for cls in to_provide:
+                obj = cls(content) if content else cls()
+                if metadata := getattr(cls, "__metadata__", None):
+                    obj = AnnotatedResult(obj, metadata)
```
wRAR, line 46: We need the same change in all real providers if we want to support annotating their provided classes. Also, most of those providers currently don't check `to_provide`, assuming it's the same as the declared provided class(es) so they will need logic improvement. This needs to include support for returning several instances when they are requested.

scrapy_poet/api.py
```diff
@@ -133,3 +134,14 @@ def parse(*args, item: page_or_item_cls, **kwargs):  # type:ignore
 
     setattr(parse, _CALLBACK_FOR_MARKER, True)
     return parse
+
+
+@dataclass
+class AnnotatedResult:
```
kmike, line 140: If we're exposing it in `__init__`, we should probably add a docstring to this class.

kmike (approved):

Gallaecio (approved):

PR #170, opened 2023-10-31
Set NO_CALLBACK for HttpClientProvider requests.

wRAR:
This should fix https://github.com/scrapy-plugins/scrapy-zyte-api/issues/135, not sure if any additional tests in any components are needed.

### DIFF
--- a/scrapy_poet/utils/__init__.py
+++ b/scrapy_poet/utils/__init__.py
@@ -17,6 +17,11 @@ from web_poet import (
     default_registry,
 )
 
+try:
+    from scrapy.http.request import NO_CALLBACK  # available on Scrapy >= 2.8
+except ImportError:
+    NO_CALLBACK = lambda: None  # noqa: E731
+
 
 def get_scrapy_data_path(createdir: bool = True, default_dir: str = ".scrapy") -> str:
     """Return a path to a folder where Scrapy is storing data.
@@ -37,6 +42,7 @@ def http_request_to_scrapy_request(request: HttpRequest, **kwargs) -> Request:
         method=request.method,
         headers=request.headers,
         body=request.body,
+        callback=NO_CALLBACK,
         **kwargs,
     )
 
--- a/scrapy_poet/utils/__init__.py
+++ b/scrapy_poet/utils/__init__.py
@@ -20,7 +20,8 @@ from web_poet import (
 try:
     from scrapy.http.request import NO_CALLBACK  # available on Scrapy >= 2.8
 except ImportError:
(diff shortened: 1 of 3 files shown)


scrapy_poet/utils/__init__.py
```diff
@@ -37,6 +42,7 @@ def http_request_to_scrapy_request(request: HttpRequest, **kwargs) -> Request:
         method=request.method,
         headers=request.headers,
         body=request.body,
+        callback=NO_CALLBACK,
```
BurnzZ, line 45: I'm not sure about this approach since there's some parts of the code that checks for the value of `request.callback` if it's `None`.

What about updating this part instead which reduces the scope to the part where we build the dependencies for a given callback? https://github.com/scrapinghub/scrapy-poet/blob/master/scrapy_poet/injection.py#L344-L345

wRAR: What would you change in that function? Can it assume that `request.callback is None` means no callback?

BurnzZ: Yes, my thoughts exactly

```python
if request.callback is None:
    return NO_CALLBACK
```

wRAR: Is it safe to change this?

BurnzZ: Probably safer in the context of this fix since the `Injector` is the only one using it at the moment.

BurnzZ: This one in particular https://github.com/scrapinghub/scrapy-poet/blob/master/scrapy_poet/injection.py#L140-L151 causes https://github.com/scrapy-plugins/scrapy-zyte-api/issues/135 because of `parse`.

wRAR: I just wonder if it's possible to get a request there that has `callback=None` but actually wants `self.parse`. Like one from `start_urls`?

wRAR: Yeah, it's indeed executed for requests from `start_urls` and so deps for those aren't created.

BurnzZ: Ahhh that's a good point! It seems your original approach would be better here. 👍

BurnzZ (commented):
> not sure if any additional tests in any components are needed.

Regarding additional tests, what do you think about testing `HttpClientProvider` to see if tries to use incorrect request meta values?

BurnzZ (approved):

Gallaecio (approved):

PR #172, opened 2023-11-06
Implement a request fingerprinter that accounts for dependencies

Gallaecio:
Pending decisions:
- [x] [`DummyResponse`](https://github.com/scrapinghub/scrapy-poet/pull/172#discussion_r1391433720) (ignore)
- [x] [Ignoring specific page inputs](https://github.com/scrapinghub/scrapy-poet/pull/172#discussion_r1391450915) (yes, ignore them)
- [x] [Dependency resolution](https://github.com/scrapinghub/scrapy-poet/pull/172#issuecomment-1808641236) (apply now that `ItemProvider` is gone)
- [x] [Page params](https://github.com/scrapinghub/scrapy-poet/pull/172#issuecomment-1808726190) (expect JSON-serializable page params)


scrapy_poet/_request_fingerprinter.py
```diff
@@ -0,0 +1,91 @@
...
+            callback = get_callback(request, self._crawler.spider)
+            if callback in self._callback_cache:
+                return self._callback_cache[callback]
+
+            deps = self._get_deps(request)
+            if deps is None:
+                return None
+
+            deps_key = json.dumps(deps, sort_keys=True).encode()
+            self._callback_cache[callback] = hashlib.sha1(deps_key).digest()
```
Gallaecio, line 79: We could technically save this call to sha1, and instead let callers use the generated bytes to generate their own hash, which they are expected to do anyway.

But this feels cleaner for some reason, and with callback caching, it should be rather cheap.

tests/test_request_fingerprinter.py
```diff
@@ -0,0 +1,243 @@
...
+    crawler = get_crawler(spider_cls=TestSpider)
+    fingerprinter = crawler.request_fingerprinter
+    request1 = Request("https://toscrape.com")
+    fingerprint1 = fingerprinter.fingerprint(request1)
+    request2 = Request("https://toscrape.com", callback=crawler.spider.parse_page)
+    fingerprint2 = fingerprinter.fingerprint(request2)
+    assert fingerprint1 != fingerprint2
+
+
+def test_same_deps():
```
BurnzZ, line 54: The spider is the same here with the one above. We can probably remove this and instead add a 3rd test above:

```python
    request3 = Request("https://toscrape.com", callback=crawler.spider.parse_page)
    fingerprint3 = fingerprinter.fingerprint(request3)
    assert fingerprint2 == fingerprint2
```

Also adding a comment to emphasize better that its testing for same deps.

tests/test_request_fingerprinter.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def get_crawler(spider_cls=None, settings=None):
+    settings = SETTINGS if settings is None else settings
+    kwargs = {}
+    if spider_cls is not None:
+        kwargs["spider_cls"] = spider_cls
+    return _get_crawler(settings=settings, **kwargs)
+
+
```
BurnzZ, line 37: Let's also add a test with different URLs to get some smoke when refactoring in the future, just in case.

BurnzZ: Let's also add one with `DummyResponse` as well.

scrapy_poet/_request_fingerprinter.py
```diff
@@ -0,0 +1,91 @@
...
+                WeakKeyDictionary()
+            )
+            self._crawler: Crawler = crawler
+
+        @cached_property
+        def _injector(self):
+            middlewares = self._crawler.engine.downloader.middleware.middlewares
+            for middleware in middlewares:
+                if isinstance(middleware, InjectionMiddleware):
+                    return middleware.injector
```
BurnzZ, line 51: Unrelated note: I find myself needed to access this as well when debugging. Would it be worth it having this feature in Scrapy itself? Perhaps some getter method in `Crawler` to easily grab some middleware.

BurnzZ (commented):
I was thinking earlier today if there are other approaches aside from this but it seems the Request Fingerprinter route is the best way to go 👍

tests/__init__.py
```diff
@@ -1,8 +1,44 @@
...
     from scrapy.utils.reactor import install_reactor
 
     install_reactor("twisted.internet.asyncioreactor.AsyncioSelectorReactor")
+
+
+def get_download_handler(crawler, schema):
+    return crawler.engine.downloader.handlers._get_handler(schema)
+
+
+def setup_crawler_engine(crawler: Crawler):
```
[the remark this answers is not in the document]
Gallaecio: I actually copied it from scrapy-zyte-api, so there is definitely room for reuse.

tests/__init__.py
```diff
@@ -1,8 +1,44 @@
...
+class DummySpider(Spider):
+    name = "dummy"
+
+
+def get_crawler(settings=None, spider_cls=DummySpider, setup_engine=True):
+    settings = settings or {}
+    crawler = _get_crawler(settings_dict=settings, spidercls=spider_cls)
+    if setup_engine:
+        setup_crawler_engine(crawler)
+    return crawler
```
kmike, line 44: What do you think about moving this code to e.g. tests.utils, to keep the `__init__` minimal?

scrapy_poet/_request_fingerprinter.py
```diff
@@ -0,0 +1,91 @@
...
+
+        def _get_deps(self, request: Request) -> Optional[List[str]]:
+            """Return a JSON-serializable structure that uniquely identifies the
+            dependencies requested by the request, or None if dependency injection
+            is not required."""
+            plan = self._injector.build_plan(request)
+            root_deps = plan[-1][1]
+            if not root_deps:
+                return None
+            return sorted([repr(cls) for cls in root_deps.values()])
```
kmike, line 65: I wonder if using fq path can be better than repr. `repr(cls)` includes the id, right? So, does it mean it's not going to be stable across runs? I think we need fingerprints to be stable, e.g. for cache to work.

tests/test_request_fingerprinter.py
Gallaecio: Hmm… Regarding `DummyResponse`, I 100% agree on having the test, but I wonder what the outcome should be. Right now the typing of the response is ignored, and I think that can be fine, but it could also be argued that `DummyResponse` should generate a different fingerprint (since in some cases that could result in different underlying requests).

Gallaecio:
I’m adding a test, `test_dep_resolution`, that shows how fingerprinting does *not* resolve dependencies, so callbacks that effectively request the same dependencies could end up with a different fingerprint.

I assume this is OK given how messy dependency resolution is due to ItemProvider, and how unlikely such cases are (having 2 callbacks on the same spider that request the same deps indirectly). However, if we ever [get rid of ItemProvider](https://github.com/scrapinghub/scrapy-poet/pull/151#discussion_r1386931164), it might be worth considering.

tests/test_request_fingerprinter.py
Gallaecio: On a related note, I wonder if we should ignore request inputs (e.g. RequestUrl, RequestBody, RequestHeaders) for fingerprinting, given they cannot alter the result.

Gallaecio:
One more thing: page params.

I have just realized that we should probably alter the fingerprint based on their presence in `Request.meta`, but [web-poet says it is an arbitrary dict](https://web-poet.readthedocs.io/en/stable/api-reference.html#web_poet.page_inputs.page_params.PageParams), so it might not be serializable and thus fingerprintable.

- Should we try to take page params into account any way for request fingerprinting?
- Should we log a warning if it is not serializable?
- How should we fingerprint requests with unserializable parameters?
    - Same fingerprint as requests without parameters
    - Their own fingerprint, but common to all requests with unserializable parameters (fixed salt). i.e. identical requests with different unserializable parameters are considered the same nonetheless.
    - Their own fingerprint, unique for each unserializable parameter set (random salt). i.e. identical requests with the same unserializable parameters are considered different nonetheless.

scrapy_poet/_request_fingerprinter.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def _serialize_dep(cls):
+        try:
+            from typing import Annotated
+        except ImportError:
+            pass
+        else:
+            if get_origin(cls) is Annotated:
+                annotated, *annotations = get_args(cls)
+                return f"{_serialize_dep(annotated)}{repr(annotations)}"
```
Gallaecio, line 31: @kmike Is this `repr` OK?

tests/test_request_fingerprinter.py
BurnzZ: >  but I wonder what the outcome should be. Right now the typing of the response is ignored, and I think that can be fine

+1 on this. 

> but it could also be argued that DummyResponse should generate a different fingerprint (since in some cases that could result in different underlying requests).

The resulting different fp would pretty much be based on the other injected parameters anyways, so it's okay to be ignored.

>  I wonder if we should ignore request inputs (e.g. RequestUrl, RequestBody, RequestHeaders) for fingerprinting, given they cannot alter the result.

Hmm not sure about this, as there might be changes in the future that affects them. For example, `typing.Annotated` might be introduced to them.

tests/test_request_fingerprinter.py
```diff
@@ -0,0 +1,341 @@
...
+    fingerprint1 = fingerprinter.fingerprint(request1)
+    request2 = Request("https://toscrape.com", callback=crawler.spider.parse_page)
+    fingerprint2 = fingerprinter.fingerprint(request2)
+    request3 = Request("https://toscrape.com", callback=crawler.spider.parse_page)
+    fingerprint3 = fingerprinter.fingerprint(request3)
+    request4 = Request("https://books.toscrape.com", callback=crawler.spider.parse_page)
+    fingerprint4 = fingerprinter.fingerprint(request4)
+    assert fingerprint1 != fingerprint2  # same url, no deps vs deps
+    assert fingerprint2 == fingerprint3  # same url, same callback
+    assert fingerprint2 != fingerprint4  # different url, same callback
```
BurnzZ, line 58: I think let's also add a test case for `Request.meta` where it doesn't affect the fps as well.

Although I'm not too sure about this since if the `"page_params"` key is present, it would create `PageParams` which _might_ affect how a Page Object behaves. For example, a boolean flag is passed to run additional requests.

This is a complicated example and might be too specific. But having the fp equality tests for now would be useful later on when we decide to do something about it.

tests/test_request_fingerprinter.py
Gallaecio: > The resulting different fp would pretty much be based on the other injected parameters anyways, so it's okay to be ignored.

I‘m thinking of:

```
def parse(self, response, request_url: RequestUrl)
def parse(self, response: DummyResponse, request_url: RequestUrl)
```

It is an unrealistic scenario, though, given that the use of `DummyResponse` is meant for cases where you get a response differently (e.g. BrowserHtml), scenarios where you don’t get a response at all seem impractical, so yes, I think you are right.

Gallaecio: > Hmm not sure about this, as there might be changes in the future that affects them. For example, typing.Annotated might be introduced to them.

What about [ignoring unannotated request and utility page inputs](https://github.com/scrapinghub/scrapy-poet/pull/172/commits/a611fd2714f60135af8299d184697dda1f9c3ac0) for now, and deciding what to do with their annotated versions later on (e.g. after merging https://github.com/scrapinghub/scrapy-poet/pull/169)?

tests/test_request_fingerprinter.py
```diff
@@ -0,0 +1,468 @@
...
+
+def test_dep_resolution():
+    """Currently we do not resolve dependencies, so it is possible to get a
+    different fingerprint for callbacks that resolve to identical
+    dependencies.
+
+    The reason for not resolving dependencies is that it could be hard where
+    items are involved. This might be addressed in the future, in which case
+    we should consider to fingerprint based on leaf dependencies (those with a
+    provider) and not on the root dependencies present in the callback.
```
Gallaecio, line 315: I think we can agree that resolving dependencies now would be too hard given ItemProvider. But I think the question then becomes: do we keep this PR on hold until we can easily resolve dependencies, or do we merge without dependency resolving and then later on change all fingerprints (invalidating caches)?

scrapy_poet/_request_fingerprinter.py
```diff
@@ -0,0 +1,171 @@
...
+                        f"you have requests that are identical except for "
+                        f"their page params, because unserializable page "
+                        f"params are treated the same as missing or empty "
+                        f"page params for purposes of request fingerprinting "
+                        f"(see "
+                        f"https://docs.scrapy.org/en/latest/topics/request-response.html#request-fingerprints). "
+                        f"This will be the only warning about this issue, "
+                        f"other requests might be also affected."
+                    )
+                return None
```
Gallaecio, line 154: Given [we expect JSON-serializable page params for caching](https://github.com/scrapinghub/web-poet/blob/d45aa1e74aab15466100cbf71b1e655d6c6572f0/web_poet/serialization/functions.py#L175), I went with expecting so, and logging a warning and fingerprinting as if there were no params when that’s not the case.

I thought about keeping a fixed size cache for page params, so save json.dumps calls on scenarios where the same page params are often used, but I was unsure about it.

I also thought about using a random value instead of treating them the same as missing or empty params, to force a different fingerprint, but that could cause a different fingerprint for requests with the same (unserializable) params, which I think is worse (can destroy more data through duplicate request dropping).

scrapy_poet/_request_fingerprinter.py
```diff
@@ -0,0 +1,171 @@
...
+    def _serialize_dep(cls):
+        try:
+            from typing import Annotated
+        except ImportError:
+            pass
+        else:
+            if get_origin(cls) is Annotated:
+                annotated, *annotations = get_args(cls)
+                return f"{_serialize_dep(annotated)}{repr(annotations)}"
+        return f"{cls.__module__}.{cls.__qualname__}"
```
kmike, line 44: Should we use https://github.com/scrapinghub/web-poet/blob/acef0a6a6d6dc8598176fc401d5785d1c9c4bf6a/web_poet/utils.py#L25?

scrapy_poet/_request_fingerprinter.py
kmike: It should work fine with the current features, but there are caveats to keep in mind - the repr of annotated values shouldn't contain ids. Python docs mention addresses commonly being a part of repr, see https://docs.python.org/3/library/functions.html#repr. For example, if we define an annotation like this (likely we won't be adding geolocation support this way though), it'd break:

```py
class Geolocation:
    def __init__(self, geo: str):
        self.geo = geo

@attrs.define
class MyPage(ItemPage):
    resp: Annotated[HttpResponse, [Geolocation("UK")]]
```

To fix it, developer should override `__repr__` of the Geolocation class; it looks non-obvious than `__repr__` changes the fingerprinting behavior. 

I'm not sure what's the best fix here. One option is to require annotations to be json-serializable, but that's invasive, as even Enum values are not serializable. 

We can also solve it with a piece of documentation, to mention that repr is used for fingerprints. 

It may also be an option to add support for some other special method to be used to serialize annotations, which the classes can define, and fallback to repr if it's not available. It may also be a singledispatch hook, not a method.

I'd probably go with docs approach for now, because the issue doesn't seem to affect any of the current or planned features, but I'm not 100% sure.

scrapy_poet/_request_fingerprinter.py
```diff
@@ -0,0 +1,171 @@
...
+            """Return a JSON array as bytes that uniquely identifies the
+            dependencies requested through scrapy-poet injection that could
+            impact the request, or None if there are no such dependencies."""
+            callback = get_callback(request, self._crawler.spider)
+            if callback in self._callback_cache:
+                return self._callback_cache[callback]
+
+            deps = self._get_deps(request)
+            if not deps:
+                return None
```
kmike, line 123: not a big deal, but the cache is not used in this case

scrapy_poet/_request_fingerprinter.py
```diff
@@ -0,0 +1,171 @@
...
+
+            deps_key = json.dumps(deps, sort_keys=True).encode()
+            self._callback_cache[callback] = deps_key
+            return self._callback_cache[callback]
+
+        def serialize_page_params(self, request: Request) -> Optional[bytes]:
+            """Returns a JSON object as bytes that represents the page params,
```
kmike, line 130:
```suggestion
            """Return a JSON object as bytes that represents the page params,
```

scrapy_poet/_request_fingerprinter.py
Gallaecio: https://github.com/scrapy/scrapy/pull/6181

tests/test_request_fingerprinter.py
Gallaecio: https://github.com/scrapinghub/scrapy-poet/pull/169 is merged, so later became now :sweat_smile:

scrapy_poet/_request_fingerprinter.py
Gallaecio: [Docs for now it is](https://github.com/scrapinghub/scrapy-poet/pull/172/commits/596431b1c271c13144b1fd45546261b53bba5cb3).

tests/test_request_fingerprinter.py
Gallaecio: Oh, I had already added a test for annotated dependencies, never mind.

BurnzZ (approved):

wRAR (approved):

kmike (approved):

PR #173, opened 2023-11-16
Add HttpRequestProvider

Gallaecio:
To mirror response counterparts, I only implemented a provider for `HttpRequest`, and not for `HttpRequestBody` or `HttpRequestHeaders`. Is that OK, or should we provide more granular providers? It would be trivial to do, but if we do I wonder if I should do the same with response page input classes as well.

To-do:
- [x] Get https://github.com/scrapinghub/web-poet/pull/191 merged, release a new version of web-poet, and require that new version in `setup.py`.

### DIFF
--- a/scrapy_poet/page_input_providers.py
+++ b/scrapy_poet/page_input_providers.py
@@ -22,6 +22,8 @@ from scrapy.http import Response
 from scrapy.utils.defer import maybe_deferred_to_future
 from web_poet import (
     HttpClient,
+    HttpRequest,
+    HttpRequestHeaders,
     HttpResponse,
     HttpResponseHeaders,
     PageParams,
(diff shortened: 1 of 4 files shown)


wRAR (approved):

scrapy_poet/page_input_providers.py
```diff
@@ -144,6 +146,29 @@ def __init__(self, injector):
...
+    <web_poet.page_inputs.http.HttpRequest>` instances.
+    """
+
+    provided_classes = {HttpRequest}
+    name = "request_data"
+
+    def __call__(self, to_provide: Set[Callable], request: Request):
+        """Builds a :class:`web_poet.HttpRequest
+        <web_poet.page_inputs.http.HttpRequest>` instance using a
+        :class:`scrapy.http.Response` instance.
```
BurnzZ, line 160:
```suggestion
        :class:`scrapy.http.Request` instance.
```

BurnzZ (approved):

PR #184, opened 2024-01-18
create a weak_cache in Injector

BurnzZ:
Currently, `SCRAPY_POET_CACHE` can be set to `True` to allow scrapy-poet to save instances created by providers locally. By default, this is turned off.

I'm proposing to introduce an `Injector.weak_cache` that would have its instances stored in memory and be removed when the corresponding `Request` instance is no longer available. By default, this is turned on. The motivation for this is to allow sharing of instances between providers by default without the risk of accumulating large number of instances locally which can result in running out of disk space.

This also means that the weakref cache mechanism in **ZyteApiProvider** ([code ref](https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/scrapy_zyte_api/providers.py#L48)) can be removed as it can use this new `.weak_cache` directly from the Injector.

Other related PRs:
- https://github.com/scrapinghub/web-poet/pull/195
- https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161
- https://github.com/zytedata/zyte-spider-templates/pull/28

### DIFF
--- a/scrapy_poet/injection.py
+++ b/scrapy_poet/injection.py
@@ -5,6 +5,7 @@ import os
 import pprint
 import warnings
 from typing import Any, Callable, Dict, List, Mapping, Optional, Set, Type, cast
+from weakref import WeakKeyDictionary
 
 import andi
 from andi.typeutils import issubclass_safe
@@ -94,6 +95,11 @@ class Injector:
             logger.info(
                 f"Cache enabled. Folder: {cache_path!r}. Caching errors: {self.caching_errors}"
             )
+
+        # This is different from the cache above as it only stores instances as long
+        # as the request exists. This is useful for latter providers to re-use the
+        # already built instances by earlier providers.
+        self.weak_cache: WeakKeyDictionary[Request, Dict] = WeakKeyDictionary()
 
     def available_dependencies_for_providers(
         self, request: Request, response: Response
@@ -293,6 +299,11 @@ class Injector:
                     f"provider: {provided_classes}"
                 )
             instances.update(objs_by_type)
+
+            if self.weak_cache.get(request):
+                self.weak_cache[request].update(objs_by_type)
+            else:
+                self.weak_cache[request] = objs_by_type
 
             if self.cache and not cache_hit:
                 # Save the results in the cache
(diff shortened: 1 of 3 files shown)


scrapy_poet/injection.py
```diff
@@ -294,6 +300,11 @@ def build_instances_from_providers(
                 )
             instances.update(objs_by_type)
 
+            if self.weak_cache.get(request):
+                self.weak_cache[request].update(objs_by_type)
+            else:
+                self.weak_cache[request] = objs_by_type
```
Gallaecio, line 306: So this fills the cache. Where is it used?

Gallaecio:
I don’t fully understand the motivation behind this. When do providers need this?

scrapy_poet/injection.py
BurnzZ: Sorry, just updated the PR's description to reflect this. It's primarily used now in `ZyteApiProvider` and replaces the weakref cache in it as well: https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161/commits/3cb22901c762aa2f0503707312a2ebabcefc42bc

BurnzZ:
> I don’t fully understand the motivation behind this. When do providers need this?

It's quite similar to having the serialized cache turned on via `SCRAPY_POET_CACHE=True`, but this is turned on by default and only needs the instances as long as the request exists. This is crucial to avoiding filling up the disk space. For example, in a job running in Scrapy Cloud with 1 unit, the spider only has 2.5GB of disk space. Assuming all the dependencies for a request (e.g. HttpResponse, Product, etc) amounts to 0.1 MB, we may run out of space in just 25,000 requests.


Previously, the example below, results in `422` error, since `HttpResponse` has been created by `HttpResponseProvider` earlier but then it's not available to `ZyteApiProvider` later on. This causes the Zyte API request to not have any page type request since `HttpResponse` isn't declared in `HttpResponseProvider.provided_classes`:

wRAR (approved):

docs/.providers.rst.swp
wRAR, line 1: Extra file

BurnzZ: Thanks! For some reason, these files aren't being ignored in my new machine.

Gallaecio (approved):

PyExplorer (approved):

PR #188, opened 2024-02-05
Fix DummyResponse detection when using strings as annotations

Gallaecio:
Fixes https://github.com/scrapy-plugins/scrapy-zyte-api/issues/170.

TIL about https://peps.python.org/pep-0649/.

### DIFF
--- a/scrapy_poet/injection.py
+++ b/scrapy_poet/injection.py
@@ -405,9 +405,6 @@ def is_callback_requiring_scrapy_response(
     callback_type_hints = get_type_hints(callback)
     first_parameter_type_hint = callback_type_hints.get(first_parameter_key, _UNDEFINED)
     if first_parameter_type_hint is _UNDEFINED:
-        return True
-
-    if first_parameter_type_hint is first_parameter.empty:
         # There's no type annotation, so we're probably using response here.
         return True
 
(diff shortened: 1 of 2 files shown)


wRAR (approved):

PyExplorer (approved):
It works now, thanks!

scrapy_poet/injection.py
```diff
@@ -387,11 +402,16 @@ def is_callback_requiring_scrapy_response(
...
         return True
 
-    if first_parameter.annotation is first_parameter.empty:
+    callback_type_hints = get_type_hints(callback)
+    first_parameter_type_hint = callback_type_hints.get(first_parameter_key, _UNDEFINED)
+    if first_parameter_type_hint is _UNDEFINED:
+        return True
+
+    if first_parameter_type_hint is first_parameter.empty:
         # There's no type annotation, so we're probably using response here.
```
kmike, line 411: This comment belongs to the code path above as well, right? Would it make sense to extract the logic to a function or a method?

Gallaecio: Actually, I think we can remove this second if statement altogether.

BurnzZ (approved):

PR #192, opened 2024-03-06
Make the savefixture command work without a Scrapy project.

wRAR:


scrapy_poet/commands.py
```diff
@@ -109,6 +110,10 @@ def run(self, args, opts):
         type_name = args[0]
         url = args[1]
 
+        if "" not in sys.path:
+            # when running without a Scrapy project the current dir may not be in sys.path,
+            # but the user may expect modules in the current dir to be available
```
kmike, line 115: Would it make sense to do it only if there is no Scrapy project detected, to keep it safe?

Gallaecio (approved):

PR #201, opened 2024-06-13
Initial DynamicDeps support.

wRAR:


Gallaecio (approved):
Not sure if you want to handle the docs now or in a later PR.

scrapy_poet/injection.py
```diff
@@ -170,20 +181,48 @@ def build_plan(self, request: Request) -> andi.Plan:
...
+                create_args = ns.keys()
+                create_args_str = ", ".join(create_args)
+                txt = (
+                    f"def __create_fn__({create_args_str}):\n"
+                    f" def dynamic_deps_factory({args_str}) -> DynamicDeps:\n"
+                    f"  return DynamicDeps({{{result_args_str}}})\n"
+                    f" return dynamic_deps_factory"
+                )
+                exec(txt, globals(), ns)
+                return ns["__create_fn__"](*dynamic_types)
```
kmike, line 223: Do you think this code (or a part of it) can be moved to a utility function, and unit tested separately? It looks quite tricky :)

wRAR: Done

scrapy_poet/injection.py
```diff
@@ -197,6 +216,37 @@ async def item_factory(page: page_object_cls) -> item_cls:  # type: ignore[valid
...
+    ) -> Callable[..., DynamicDeps]:
+        """Return a function that creates a :class:`.DynamicDeps` instance from its args.
+
+        It takes instances of types from ``dynamic_types`` as args and returns
+        a :class:`.DynamicDeps` instance where keys are types and values are
+        corresponding args. It has correct type hints so that it can be used as
+        an ``andi`` custom builder.
+        """
+
+        # inspired by dataclasses._create_fn()
```
kmike, line 231: 1. It'd be good to add a link to the source code (with a commit hash?), because in latest Python there is no more dataclasses._create_fn. 
2. What do you think about testing the string generation part separately? To make it clear what kind of code we're generating for the execution.

Sorry, I just find this code tricky, and it'd be good to try to make it a bit more approachable, for the future :)

scrapy_poet/injection.py
```diff
@@ -197,6 +216,37 @@ async def item_factory(page: page_object_cls) -> item_cls:  # type: ignore[valid
...
+        ]
+        result_args_str = ", ".join(result_args)
+        ns = {type_.__name__: type_ for type_ in dynamic_types}
+        create_args = ns.keys()
+        create_args_str = ", ".join(create_args)
+        txt = (
+            f"def __create_fn__({create_args_str}):\n"
+            f" def dynamic_deps_factory({args_str}) -> DynamicDeps:\n"
+            f"  return DynamicDeps({{{result_args_str}}})\n"
+            f" return dynamic_deps_factory"
```
kmike, line 245: What kind of dependency injection is supported for DynamicDeps? E.g. can you have a page object, or an item, or a non-leaf dependency in DynamicDeps? I think we need tests for that.

scrapy_poet/injection.py
wRAR: It's really good then that I've read the 3.11 implementation, the 3.12 one is more complicated :)

scrapy_poet/injection.py
wRAR: It should support anything that can go to a callback, I'll add some tests.

docs/dynamic-deps.rst
```diff
@@ -0,0 +1,54 @@
...
+
+        def start_requests(self):
+            yield scrapy.Request("http://books.toscrape.com/", self.parse_book)
+
+
+        def parse_book(self, response, book_page: BookPage, other_dep: OtherDep):
+            ...
+
+In some cases some or all of the dependencies need to be specified dynamically
+instead, e.g. because they need to be different for different pages using the
```
kmike, line 26:
```suggestion
instead, e.g. because they need to be different for different requests using the
```

"pages" looks confusing, as reader may understand it as "for different page objects"

docs/dynamic-deps.rst
```diff
@@ -0,0 +1,54 @@
...
+                meta={"inject": [OtherDep]},
+            )
+
+
+        def parse_book(self, response, book_page: BookPage, dynamic: DynamicDeps):
+            other_dep = dynamic[OtherDep]
+            ...
+
+The types passed this way will be used in the dependency resolution as usual,
+and the created instances will be available in the
```
kmike, line 53:
```suggestion
The types passed this way are used in the dependency resolution as usual,
with the created instances available in the
```

docs/dynamic-deps.rst
```diff
@@ -0,0 +1,54 @@
...
+        def start_requests(self):
+            yield scrapy.Request(
+                "http://books.toscrape.com/",
+                self.parse_book,
+                meta={"inject": [OtherDep]},
+            )
+
+
+        def parse_book(self, response, book_page: BookPage, dynamic: DynamicDeps):
+            other_dep = dynamic[OtherDep]
```
kmike, line 49: Is there a way to get a list of the created dependencies in the DynamicDeps (i.e. the available keys)? It seems we should document it. Currently only getting by type is documented, if I'm not mistaken.

scrapy_poet/injection.py
```diff
@@ -170,20 +182,28 @@ def build_plan(self, request: Request) -> andi.Plan:
...
     ) -> Callable[[Callable], Optional[Callable]]:
         """Return a function suitable for passing as ``custom_builder_fn`` to ``andi.plan``.
 
         The returned function can map an item to a factory for that item based
-        on the registry.
+        on the registry and also supports filling :class:`.DynamicDeps`.
         """
 
         @functools.lru_cache(maxsize=None)  # to minimize the registry queries
         def mapping_fn(item_cls: Callable) -> Optional[Callable]:
```
kmike, line 198: should we rename item_cls to e.g. dependency_cls, or requested_cls?

scrapy_poet/injection.py
```diff
@@ -54,6 +56,16 @@ class _UNDEFINED:
     pass
 
 
+class DynamicDeps(UserDict):
```
kmike, line 59: It seems the only thing UserDIct gives now is .data attribute with dict content; if it's not used, subclassing from dict should be fine. This seems to be useful when implementing new methods (which we don't implement), not when using instances: for the user, deps.data would work the same as deps.

tests/test_injection.py
```diff
@@ -535,6 +542,126 @@ def callback(
...
+            pass
+
+        provider = get_provider({Cls1, Cls2})
+        injector = get_injector_for_testing({provider: 1})
+
+        expected_instances = {
+            DynamicDeps: DynamicDeps({Cls1: Cls1(), Cls2: Cls2()}),
+            Cls1: Cls1(),
+            Cls2: Cls2(),
+        }
```
kmike, line 581: In this test DynamicDeps[Cls1] is not the same instance as Cls1 instance received in the callback.

Is it an artifact of the test, or is it the way the implementation works?

Does it cause any problems, and if so, should they be fixed or documented?

I guess there could be 2 kinds of issues. First, you may just get the same data, but in different instances (i.e. copies). Second, you may need to do some work multiple times (fetch web page twice?? Run page object extraction code twice to get the same item?). The second issue looks unlikely (right?), but the first seems to apply - should we document it, at least briefly?

kmike:
Looks great overall 👍

docs/dynamic-deps.rst
wRAR: I assumed the user already knows which deps are there because it's they who set the meta key, but in other uses cases (which?) they can look at `dynamic.keys()`, do you want to document this or are you thinking about something else?

tests/test_injection.py
wRAR: It's actually the same instance, these checks in several tests (including old ones) aren't really correct as Cls1 and Cls2 are strings and so are compared just by the value. If this test was checking instance equality the expectation would need to be rewritten.

kmike: Nice!
Does it worth a test, or is it already tested somewhere?

kmike (approved):

PR #208, opened 2024-10-07
Restore and improve the annotated dynamic deps support.

wRAR:
A quick fix for a regression in #207 that raised an exception for annotated types.

The proper annotated dynamic deps support including support for multiple annotated deps requires some refactoring and logic changes and will be done separately.

Edit: as the most basic fix doesn't work on 3.9 I added a better one that includes support for multiple annotated deps but not other things that are still needed.

Edit 2: here is hopefully the complete fix, with making the results not annotated.

### DIFF
--- a/scrapy_poet/injection.py
+++ b/scrapy_poet/injection.py
@@ -247,7 +247,7 @@ class Injector:
         """
         ns: Dict[str, type] = {}
         for type_ in dynamic_types:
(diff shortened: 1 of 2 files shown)


wRAR:
The docs fix is in #209

Gallaecio (approved):

kmike (approved):

PR #211, opened 2024-10-14
Add Python 3.13, drop Python 3.8, update tool versions

wRAR:


tests/test_injection.py
```diff
@@ -779,8 +743,9 @@ class TestInjectorStats:
     def test_stats(self, cb_args, expected, injector):
         def callback_factory():
             args = ", ".join([f"{k}: {v.__name__}" for k, v in cb_args.items()])
-            exec(f"def callback(response: DummyResponse, {args}): pass")
-            return locals().get("callback")
+            ns = {}
+            exec(f"def callback(response: DummyResponse, {args}): pass", None, ns)
```
wRAR, line 747: https://docs.python.org/3/whatsnew/3.13.html#defined-mutation-semantics-for-locals

Gallaecio (approved):

PR #216, opened 2025-01-09
Add-on

Gallaecio:


scrapy_poet/utils/testing.py
```diff
@@ -229,26 +232,46 @@ def process_response(self, request, response, spider):
...
+
+def create_scrapy_settings():
+    """Return the default scrapy-poet settings."""
+    warn(
+        "The scrapy_poet.utils.create_scrapy_settings() function is deprecated.",
+        DeprecationWarning,
+        stacklevel=2,
     )
-    return Settings(s)
+    return Settings(_get_test_settings())
```
Gallaecio, line 274: Deprecated because the returned settings include all Scrapy default settings as project-level settings, which prevents the add-on-level request fingerprinting from being taken into account.

wRAR (approved):

PR #218, opened 2025-01-21
Update test expectations for a future where web_poet.pages.is_injectable(type(None)) is False

Gallaecio:
To do:
- [x] Reproduce the issue (reported elsewhere) with a test.
- [x] ~Fix the issue.~ (the issue is in web-poet: https://github.com/scrapinghub/web-poet/pull/212)
- [x] [Support non-injectable type hints in attrs](https://github.com/scrapinghub/scrapy-poet/pull/218#discussion_r1925836198).

### DIFF
--- a/tests/test_middleware.py
+++ b/tests/test_middleware.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from textwrap import dedent
 from typing import Optional, Type, Union
 
+import andi
 import attr
 import pytest
 import scrapy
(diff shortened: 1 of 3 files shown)


tests/test_middleware.py
```diff
@@ -124,7 +125,39 @@ def test_deprecation_setting_SCRAPY_POET_OVERRIDES(settings) -> None:
 
 
 @attr.s(auto_attribs=True)
-class OptionalAndUnionPage(WebPage):
+class OptionalAndUnionPageNew(WebPage):
+    breadcrumbs: BreadcrumbsExtraction
+    opt_check_1: Optional[BreadcrumbsExtraction]
```
kmike, line 130: Just in case, could you please add a test which checks that in case of `Optional[something]` things don't break if `something` is not available? There are similar tests in test_optional_and_unions_old, but not exactly that.

Gallaecio: Great catch. If it’s not *injectable*, that indeed causes an issue, I need to figure out a solution for that. It might have to do with attrs’ attribute default values not being recognized by andi, because I don’t think it happens for default values in regular callables (e.g. Scrapy callbacks).

> not available

Did you mean not injectable, or should we also support scenarios where a bad type hint is provided? (e.g. `var1: "UnexistingType` or `var2: "Optional[UnexistingType]`)

Gallaecio: Addressed by https://github.com/scrapinghub/andi/pull/33.

kmike (approved):

PR #246, opened 2026-04-23
Add __version__ attribute

elacuesta:
I recently had issues in an agent session because it was trying to identify the version with `python -c "import scrapy_poet; print(scrapy_poet.__version__)"`

### DIFF
--- a/scrapy_poet/__init__.py
+++ b/scrapy_poet/__init__.py
@@ -1,3 +1,8 @@
+from importlib.metadata import version
+
+__version__ = version("scrapy-poet")
+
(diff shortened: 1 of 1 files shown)


wRAR:
https://packaging.python.org/en/latest/discussions/versioning/#accessing-version-information-at-runtime for the record (I felt like this was marked as deprecated somewhere, but at least here it's not).

AdrianAtZyte:
Shall we close this and rely on the modern (slower) approach, or do you guys want to go ahead with this?

PR #249, opened 2026-05-15
Add InjectionMiddleware.get()

AdrianAtZyte:

PR #252, opened 2026-06-10
Provide an API for inline requests

AdrianAtZyte:
Based on web-poet’s built-in framework API.


wRAR:
Supersedes #249? :)

AdrianAtZyte:
:facepalm: 

This time I did remember that I had done the work, but I was convinced I had not created a PR for it. I think this is a much cleaner API, so yes.

